PR #12, opened 2026-02-24
Add ALLOWED_WORKSPACES config for pinned workspace access

dcellison:
## Summary

Adds `ALLOWED_WORKSPACES` config to make specific workspace directories accessible by name without requiring them to be under `WORKSPACE_BASE`. This is useful for working with existing projects that live outside the workspace base directory.

- `ALLOWED_WORKSPACES=/path/a,/path/b` in `.env` — comma-separated absolute paths
- Non-existent paths are skipped at startup with a warning, not a crash
- `/workspace <name>` checks `WORKSPACE_BASE` first, then falls through to `ALLOWED_WORKSPACES` by directory name
- `WORKSPACE_BASE` wins on name collision (per spec)
- `/workspaces` keyboard shows allowed workspaces as a pinned section above history
- History deduplicates against allowed workspaces — paths shown in pinned section do not repeat in history
- Callback data scheme: `ws:allowed:<index>` for pinned buttons alongside existing `ws:home` and `ws:<index>`

## Test plan

- [x] All tests pass (`make test`) — 172 total, 11 new
- [x] Ruff lint and format clean (`make check`)
- [ ] Manual: add `ALLOWED_WORKSPACES=/path/to/project` to `.env`, restart, verify `/workspace project-name` switches correctly
- [ ] Manual: verify `/workspaces` keyboard shows pinned section above history
- [ ] Manual: verify a path in both allowed and history appears only once (pinned section)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -25,6 +25,11 @@ CLAUDE_MAX_BUDGET_USD=10.0
 # Must be an existing directory. If unset, only /workspace home works.
 WORKSPACE_BASE=
 
+# Additional workspaces accessible by name, outside WORKSPACE_BASE.
+# Comma-separated absolute paths. Non-existent paths are skipped at startup.
+# Example: ALLOWED_WORKSPACES=/home/user/project-a,/home/user/project-b
+ALLOWED_WORKSPACES=
+
 # ── Webhook & Scheduling API ─────────────────────────────────────
 
 # Port for the local webhook/API server (receives GitHub events, scheduling API)
--- a/src/kai/bot.py
+++ b/src/kai/bot.py
@@ -668,30 +668,51 @@ async def _workspaces_keyboard(
     current_path: str,
     home_path: str,
     base: Path | None,
+    allowed_workspaces: list[Path],
 ) -> InlineKeyboardMarkup:
     """
     Build an inline keyboard for workspace switching.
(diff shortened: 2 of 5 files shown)


dcellison:
To add some nuance to the above: there *is* a grain of truth in the concern. `ALLOWED_WORKSPACES` controls which directory the inner Claude process starts in, but once that process is running with `bypassPermissions`, nothing prevents it from using absolute paths or `cd`-ing elsewhere. The workspace restriction is a UX guardrail (controlling what you can navigate to via `/workspace`), not a filesystem sandbox.

That's by design, not an oversight. Kai's threat model is "unauthorized person sends messages to the bot," which is handled by the user allowlist. It's a single-user local deployment - the inner Claude has the same filesystem access as any process running as that user. That's the correct posture for this architecture.

Where process-level restriction would matter is a multi-user or shared-server deployment where you genuinely need to limit what Claude can touch. But that's not Kai's deployment model, and `dontAsk` with a command allowlist is a fragile way to achieve it even if it were - you'd want OS-level sandboxing (containers, separate user accounts) for real isolation.

So: technically accurate that `bypassPermissions` doesn't enforce workspace boundaries at the process level. Practically irrelevant for the intended use case.

PR #26, opened 2026-03-03
Add optional user separation for Claude subprocess

dcellison:
## Summary

- Adds `CLAUDE_USER` config option to run the inner Claude process as a different OS user via `sudo -u`
- Closes the TOTP sudoers hole identified by mmx003 on #23: Claude can no longer access `/etc/kai/totp.secret` because the configured user has no sudoers rules
- When `CLAUDE_USER` is unset, behavior is identical to current - zero impact on existing installs

## What changed

- `config.py`: new `claude_user` field, loaded from `CLAUDE_USER` env var
- `claude.py`: subprocess spawned via `sudo -u <user> --` when configured; new `_kill_proc()` helper centralizes signal dispatch with process group support (`os.killpg`) for reliable `/stop` across the sudo boundary
- `bot.py`: passes `claude_user` through to `PersistentClaude`
- `.env.example`: documents the new option with requirements
- New test file `tests/test_claude.py` (158 lines) covering command construction, process group signals, edge cases
- New config tests for `CLAUDE_USER` parsing

## How it works

When `CLAUDE_USER=daniel`, the process tree becomes:

    kai (bot) -> sudo -u daniel -- claude ...

The subprocess starts in its own process group (`start_new_session=True`) so `/stop` can kill the entire tree via `os.killpg()`, not just the sudo wrapper.

## Test plan

- [x] All 243 tests pass (no regressions)
- [x] Ruff clean
- [ ] Manual: verify Claude spawns as configured user (`ps aux | grep claude`)
- [ ] Manual: verify `/stop` kills the process reliably
- [ ] Manual: verify `sudo cat /etc/kai/totp.secret` fails from Claude's session

Addresses #23

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -84,3 +84,18 @@ TTS_ENABLED=false
 #TOTP_LOCKOUT_ATTEMPTS=3
 # Duration of lockout in minutes after the threshold is exceeded (default 15)
 #TOTP_LOCKOUT_MINUTES=15
+
+# ── User Separation (optional) ──────────────────────────────────
+
+# Run the inner Claude process as a different OS user for security isolation.
+# The bot spawns Claude via 'sudo -u <user> claude ...', creating a hard
+# OS-level boundary between the bot and the AI subprocess.
+#
+# Requirements:
+#   - The user must exist on the system
+#   - The user must NOT be an admin (no sudo privileges)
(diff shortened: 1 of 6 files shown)


dcellison:
@mmx003 Let me know if you think this covers the user separation issue.

PR #76, opened 2026-03-16
Add per-workspace configuration via workspaces.yaml

### LINKED ISSUE #71: Per-workspace configuration file (launcher, model, budget)
## Problem

Different workspaces require different Claude environments — custom MCP servers, environment variables, system prompts, or even different models. Currently the only workspace setting is a flat list of paths in `ALLOWED_WORKSPACES`.

## Proposal

Add an optional `workspaces.yaml` (or `.json`) configuration file that defines per-workspace settings:

```yaml
workspaces:
  - path: /home/daniel/projects/webapp
    launcher: /home/daniel/launchers/webapp.sh
    model: opus

  - path: /home/daniel/projects/docs
    model: sonnet

  - path: /home/daniel/notes
```

### Fields

| Field | Description | Default |
|-------|-------------|---------|
| `path` | Workspace directory (required) | — |
| `launcher` | Script to run instead of `claude`. Receives all CLI args from Kai, must `exec claude "$@"` at the end. | `claude` |
| `model` | Override model for this workspace | global `CLAUDE_MODEL` |

### Launcher script

A launcher script sets up the environment (loads `.env`, registers MCP servers, appends system prompts) before exec-ing claude. Example use case: a workspace that needs Confluence and Jira MCP servers registered, or a custom system prompt loaded via `--append-system-prompt`.

Kai passes the same headless arguments (`--input-format stream-json --output-format stream-json ...`) — the launcher just prepends its setup and exec's claude with all args.

### File location

In protected installations: `/etc/kai/workspaces.yaml` (root-owned, read via `_read_protected_file`). Fallback to `PROJECT_ROOT/workspaces.yaml` for development.

### Backward compatibility

`ALLOWED_WORKSPACES` continues to work. If `workspaces.yaml` exists, workspaces defined there are merged with (and override settings for) paths from `ALLOWED_WORKSPACES`.

dcellison:
## Summary

Adds declarative per-workspace configuration via `workspaces.yaml`, replacing the launcher script concept from issue #71 with a data-only approach. Each workspace can now have its own model, budget, timeout, environment variables, and system prompt - all without introducing code execution in the trust boundary between Kai and Claude.

- `WorkspaceConfig` frozen dataclass and YAML loader in `config.py` with full validation (model names, budget/timeout ranges, env_file existence, system_prompt mutual exclusion, duplicate path detection)
- Per-workspace overrides applied in `PersistentClaude` constructor and `change_workspace()` using reset-then-override pattern to prevent stale values carrying over between workspaces
- Environment variable merge with security-invariant ordering: base env -> env_file -> inline env -> webhook secret (workspace env cannot override the webhook secret)
- System prompt injection via first-message context (not CLI flag), with file re-read on each session start for live updates
- YAML boolean coercion fix: `true`/`false` emit lowercase strings to match .env conventions
- Workspace switch confirmation now shows applied config details (model, budget)
- YAML workspace paths automatically merged into the allowed workspaces set
- Backward compatible: everything works identically without `workspaces.yaml`

### Config format

```yaml
workspaces:
  - path: ~/projects/webapp
    claude:
      model: opus
      budget: 15.0
      timeout: 180
      env:
        DATABASE_URL: "postgres://localhost/webapp_dev"
      system_prompt: |
        This is a Rails 7 application using PostgreSQL.

  - path: ~/projects/docs
    claude:
      model: haiku
```

### Files changed

- **`src/kai/config.py`** - `WorkspaceConfig` dataclass, `_read_protected_yaml()`, `parse_env_file()`, `_load_workspace_configs()` with validation, `Config.get_workspace_config()` lookup, YAML paths merged into `allowed_workspaces`
- **`src/kai/claude.py`** - `workspace_config` constructor param, `_default_*` fields for global defaults, env merge in `_ensure_started()`, `_get_workspace_system_prompt()`, `change_workspace()` reset-then-override
- **`src/kai/bot.py`** - `_do_switch_workspace()` passes workspace config, `_switch_workspace()` shows config details, `create_bot()` passes initial workspace config
- **`workspaces.yaml.example`** - Annotated example config
- **`tests/test_workspace_config.py`** - 26 tests for YAML loading, validation, env parsing, lookup
- **`tests/test_claude.py`** - 24 tests for constructor overrides, defaults, workspace switch cycles, system prompt, env merge, webhook secret protection

Closes #71

## Test plan

- [x] Basic YAML loading with multiple workspaces
- [x] Missing/empty/invalid YAML files handled gracefully
- [x] Path resolution (~ expansion, resolve)
- [x] Non-existent paths skipped with warning
- [x] Duplicate paths: first wins
- [x] Invalid model/budget/timeout validation
- [x] Mutual exclusion of system_prompt and system_prompt_file
- [x] env_file and system_prompt_file existence validation
- [x] YAML bool coercion emits lowercase
- [x] Protected installation (/etc/kai/) tried first
- [x] parse_env_file: quotes, export prefix, comments, malformed lines
- [x] Constructor applies workspace config overrides
- [x] Partial config only overrides specified fields
- [x] Global defaults preserved and restored on workspace switch
- [x] No stale values when switching between configured workspaces
- [x] /model override cycle: config model restored after switch
- [x] System prompt: inline, from file, file deleted after load
- [x] Env merge: workspace env, env_file, inline overrides env_file
- [x] Webhook secret cannot be overridden by workspace env
- [x] Config.get_workspace_config() found/not found
- [x] All 883 tests pass (50 new)

### DIFF
--- a/tests/test_bot.py
+++ b/tests/test_bot.py
@@ -21,6 +21,7 @@ from kai.bot import (
     _acquire_lock_or_kill,
     _chunk_text,
     _clear_responding,
+    _do_switch_workspace,
     _edit_message_safe,
     _is_authorized,
     _is_workspace_allowed,
@@ -35,6 +36,7 @@ from kai.bot import (
     _short_workspace_name,
     _truncate_for_telegram,
     _voices_keyboard,
+    _workspace_config_suffix,
     _workspaces_keyboard,
     create_bot,
     handle_canceljob,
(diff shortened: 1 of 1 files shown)


dcellison:
## Review by Kai

## PR #76 Review: Per-Workspace Configuration

All prior-review issues are resolved. Two new findings.

---

### Warning

**`config.py` — YAML `null` env values silently become the string `"None"`**

In the env coercion pass inside `_load_workspace_configs`:

```python
env = {str(k): str(v).lower() if isinstance(v, bool) else str(v) for k, v in env.items()}
```

`str(None)` is `"None"`. A YAML entry like:

```yaml
env:
  MY_VAR:          # or  MY_VAR: ~  or  MY_VAR: null
```

produces `{"MY_VAR": "None"}` — almost certainly not intended. The bool case gets explicit handling; null should too. Either skip null values with a warning, or map them to empty string:

```python
env = {
    str(k): str(v).lower() if isinstance(v, bool) else ("" if v is None else str(v))
    for k, v in env.items()
}
```

---

### Suggestion

**`bot.py` — `extras` formatting logic duplicated**

The model/budget suffix construction appears verbatim in both `_switch_workspace` (lines ~821–824) and `handle_workspace_callback` (lines ~1015–1021):

```python
extras = []
if ws_config and ws_config.model:
    extras.append(f"model: {ws_config.model}")
if ws_config and ws_config.budget is not None:
    extras.append(f"budget: ${ws_config.budget:.2f}")
suffix = f" ({', '.join(extras)})" if extras else ""
```

A small helper (`_workspace_config_suffix(ws_config)`) would remove the duplication and make future changes (e.g. showing timeout) a one-liner.

---

Everything else — the `_YAML_MALFORMED` sentinel blocking protected-file fall-through, `OSError` catch on local file reads, bool rejection for `budget` and `timeout`, reset-then-override pattern in `change_workspace()`, webhook-secret-last ordering, and the `_default_*` preservation — is correct. Test coverage is thorough.

dcellison:
## Review by Kai

## PR #76 Review: Per-Workspace Configuration

All prior-review issues are resolved. Three new findings.

---

### Warning

**`config.py` — `_read_protected_yaml` return annotation is a regression**

The annotation was updated from `-> dict | object | None` to `-> dict | None`, but the function still returns `_YAML_MALFORMED` in two branches — a non-`dict`, non-`None` value. The new annotation is *less* accurate than the old one: type checkers will incorrectly treat `is _YAML_MALFORMED` checks in callers as dead code, and future readers have no signal from the signature that a third case exists. The docstring documents the sentinel, but the annotation contradicts it. Either restore `-> dict | object | None`, or use `-> dict | _YAML_MALFORMED | None` (with `_YAML_MALFORMED = object()` typed appropriately), or a `Literal` sentinel.

---

### Suggestions

**`config.py` — `_coerce_env_value` redefined on every loop iteration**

```python
for entry in entries:
    ...
    def _coerce_env_value(v: object) -> str:   # re-created for every workspace entry
        ...
    env = {str(k): _coerce_env_value(v) for k, v in env.items()}
```

The helper is defined inside `for entry in entries:`, so it's rebuilt on each iteration. Move it to module level or to the top of `_load_workspace_configs()`.

---

**`config.py` — non-list `workspaces:` value silently returns `{}`**

```python
entries = data.get("workspaces")
if not isinstance(entries, list):
    return {}   # no log
```

If a user writes `workspaces: haiku` or `workspaces: {}` (a common YAML typo), the loader returns an empty dict with no diagnostic. The `None` case (key absent) is expected silence, but a non-`None`, non-list value is a config mistake. A `log.warning()` when `entries is not None and not isinstance(entries, list)` would make this diagnosable — consistent with the other "expected a YAML dict, got ..." warnings already in this file.

---

Everything else is solid: `_YAML_MALFORMED` correctly blocks protected-file fall-through, `OSError` is caught on local file reads, `isinstance(budget/timeout, bool)` correctly rejects YAML boolean coercion, null env values map to `""` not `"None"`, the reset-then-override pattern in `change_workspace()` prevents stale values, and the webhook-secret-last ordering is a correct security invariant.

PR #80, opened 2026-03-17
Add is_dir() guard to text command workspace switching

### LINKED ISSUE #79: _switch_workspace text command missing is_dir() check
Minor inconsistency: `handle_workspace_callback` (inline keyboard path) checks `path.is_dir()` before switching and shows "That workspace no longer exists" if the directory was deleted. The text command path through `_switch_workspace` doesn't — it passes the path straight to `_do_switch_workspace`, which causes Claude to start with a non-existent cwd.

Only affects the edge case where a directory from `allowed_workspaces` is deleted after bot startup and the user switches via `/workspace <name>` (not via the keyboard).

One-line fix — add an `is_dir()` guard in `_switch_workspace` before calling `_do_switch_workspace`.

dcellison:
## Summary

- Add `is_dir()` guard in `_switch_workspace` so text commands (`/workspace <name>`) reject deleted directories instead of setting Claude's cwd to a non-existent path
- The keyboard path (`handle_workspace_callback`) already had this guard; the text command path did not
- Fix existing `test_home_switches` to use a real `tmp_path` directory since the guard now validates the path exists on disk

Fixes #79

## Test plan

- [x] New test: switching to a non-existent path replies "That workspace no longer exists."
- [x] Existing test fixed: `test_home_switches` uses real directory
- [x] All 897 tests pass

### DIFF
--- a/tests/test_bot.py
+++ b/tests/test_bot.py
@@ -34,6 +34,7 @@ from kai.bot import (
     _save_to_workspace,
     _set_responding,
     _short_workspace_name,
+    _switch_workspace,
     _truncate_for_telegram,
     _voices_keyboard,
     _workspace_config_suffix,
@@ -1615,8 +1616,6 @@ class TestSwitchWorkspaceConfig:
             patch("kai.bot.sessions", new_callable=AsyncMock),
             patch("kai.bot.webhook"),
         ):
-            from kai.bot import _switch_workspace
-
             await _switch_workspace(update, ctx, ws_path.resolve())
 
         reply_text = update.message.reply_text.call_args[0][0]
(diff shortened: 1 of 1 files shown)


dcellison:
## Review by Kai

This PR is clean. The implementation matches the spec exactly. A couple of minor notes:

**Suggestion — `test_switch_deleted_directory` uses a hardcoded path** (`tests/test_bot.py:1637`)

`Path("/tmp/nonexistent-workspace-12345")` is assumed to not exist on disk. It almost certainly won't, but it's not guaranteed. Using `tmp_path` with deletion (as the spec mentioned as an option) would be more robust:

```python
async def test_switch_deleted_directory(self, tmp_path):
    gone = tmp_path / "gone"
    gone.mkdir()
    gone.rmdir()
    await _switch_workspace(update, ctx, gone)
```

The spec explicitly permitted the current approach, so this is a nit, not a bug.

**Suggestion — local import inside test** (`tests/test_bot.py:1627`)

`from kai.bot import _switch_workspace` is inside the test method body. If `_switch_workspace` is imported at the top of the test file elsewhere this is inconsistent; if it isn't, moving it to the top-of-file imports matches conventional style. No functional impact.

---

Everything else is correct:
- Guard placement in `bot.py` is exactly right — after the "already in" check, before `_do_switch_workspace`, so the `/workspace new` path (directory just created) is unaffected.
- `test_home_switches` fix using `tmp_path` is necessary and correct; without it the new guard would cause that test to fail.
- Error message matches the spec verbatim.
- No security issues, no logic errors.

PR #97, opened 2026-03-19
Add multi-user Phase 1: user model and safe routing

### LINKED ISSUE #94: Multi-user Phase 1: user model and safe routing
## Problem

Kai has no concept of "users" as first-class entities. Authorization is a flat comma-separated list of Telegram IDs (`ALLOWED_USER_IDS`), and all runtime state that should be per-user is either global or inconsistently namespaced:

- **API routing is hardcoded.** Internal webhook handlers resolve `chat_id` from a single app-level value set at startup. Messages, files, and jobs always route to the first user in the list.
- **No per-user configuration.** Model preference, budget ceiling, timeout - all global. Two users sharing the instance share every setting.
- **No user identity metadata.** No display names for logs, no GitHub handle mapping for webhook routing, no role distinctions for who receives external notifications.
- **Settings namespace is ad-hoc.** Some settings are namespaced by chat_id, others are global. No consistent pattern.

## Why

PR #82 identified a real bug in webhook routing for multi-user setups. Rather than patching the symptom, this introduces users as a first-class concept. Phase 1 is the foundation that Phases 2 and 3 build on, but is useful on its own.

## Solution

Introduce `users.yaml` as a declarative user config file. Each user gets identity metadata (name, role, GitHub handle), a home workspace, and admin-controlled resource limits. Internal API endpoints gain optional `chat_id` routing so inner Claude can target the correct user.

Admin-controlled identity and authorization live in `users.yaml` (who you are, what you can do). User-controlled preferences live in the database settings table (how you like things).

Fully backward-compatible - existing single-user deployments using `ALLOWED_USER_IDS` change nothing.

dcellison:
## Summary

Introduces users as first-class entities in Kai, replacing the flat `ALLOWED_USER_IDS` comma-separated list with a declarative `users.yaml` configuration file. This is Phase 1 of a three-phase multi-user plan.

### What this PR does

**User model (`config.py`):**
- `UserConfig` frozen dataclass: telegram_id, name, role (admin/user), github handle, os_user (Phase 3), home_workspace, max_budget
- `_load_user_configs()` with the same dual-mode pattern as workspace config: `/etc/kai/users.yaml` (protected) then `PROJECT_ROOT/users.yaml` (dev)
- Full validation: bool guards on telegram_id and budget (same Python `bool` subclass protections as workspace config), role validation against `{"admin", "user"}`, home_workspace existence check, duplicate ID handling (first wins)
- Warning when no admin user is defined (external webhooks route to arbitrary user)
- Config convenience methods: `get_user_config(telegram_id)`, `get_user_by_github(login)` (case-insensitive), `get_admins()`

**Safe API routing (`webhook.py`):**
- `_resolve_chat_id(request, payload)` helper extracts optional `chat_id` from request payloads, falling back to the app-level default
- Applied to four internal handlers: `_handle_schedule`, `_handle_send_message`, `_handle_send_file` (POST with JSON body), and `_handle_get_jobs` (GET with query parameter `?chat_id=N`)
- External handlers (`_handle_github`, `_handle_generic`) unchanged - they route to the default admin. Actor-based routing is a Phase 3 feature.
- App-level `chat_id` set from first admin in users.yaml, with prominent warning fallback if no admin defined

**Context injection (`claude.py`):**
- `chat_id` parameter added to `PersistentClaude.send()` and `_send_locked()`
- On first session message, chat_id is injected into context so inner Claude can include it in API calls for correct multi-user routing
- Forward-compatible with Phase 3 per-user subprocess pool where each instance permanently knows its user

**Callers updated:**
- `bot.py` `_handle_response()` passes `chat_id` to `claude.send()`
- `cron.py` `_job_callback()` passes `chat_id` to `claude.send()`

### Backward compatibility

Fully backward-compatible. When `users.yaml` does not exist, `ALLOWED_USER_IDS` works exactly as before. Multi-user is opt-in. When both exist, `users.yaml` wins with an informational log warning.

### Schema-only fields (not enforced in Phase 1)

- `os_user`: stored for Phase 3 subprocess isolation via `sudo -u`
- `max_budget`: stored for future `/budget` command enforcement

### What this does NOT do (future phases)

- Per-user subprocess isolation (Phase 3: `ClaudeManager` with `dict[int, PersistentClaude]`)
- Per-user workspace/history/file namespacing (Phase 2)
- GitHub actor-based webhook routing (Phase 3)
- `/budget` command or budget enforcement
- Per-user flag file replacement

Fixes #94

## Test plan

- [x] UserConfig dataclass: required fields, all fields, frozen
- [x] _load_user_configs: basic loading, missing file (None), empty file, invalid YAML
- [x] Validation: missing telegram_id, missing name, invalid role, negative budget, bool budget, bool telegram_id
- [x] Duplicate IDs: first wins
- [x] Home workspace: non-existent path skipped
- [x] Protected installation tried first
- [x] No admin warning logged but does not fail
- [x] Default role is "user"
- [x] os_user stored as string
- [x] Config.get_user_config: found, not found, no yaml
- [x] Config.get_user_by_github: found, case-insensitive, not found, no yaml
- [x] Config.get_admins: filters correctly, empty, no yaml
- [x] Existing test fixes: cron mock send accepts chat_id, webhook mock has query dict, claude lock test passes chat_id
- [x] All 930 tests pass (29 new)

### DIFF
--- a/src/kai/config.py
+++ b/src/kai/config.py
@@ -553,11 +553,13 @@ def _load_user_configs() -> dict[int, UserConfig] | None:
             log.warning("%s: expected a YAML dict, got %s", local_path, type(data).__name__)
             return None
 
-    # After the _YAML_MALFORMED and None checks, data must be a dict
+    # After the _YAML_MALFORMED and None checks, data should be a dict
     # (protected path returns _YAML_MALFORMED for non-dicts, local path
-    # checks isinstance explicitly). This assert narrows the type for
-    # static analysis and guards against future refactoring.
-    assert isinstance(data, dict)
+    # checks isinstance explicitly). Guard defensively rather than assert
+    # since assertions are stripped under Python -O.
+    if not isinstance(data, dict):
+        log.warning("users.yaml: expected a YAML dict, got %s", type(data).__name__)
+        return None
 
     entries = data.get("users")
     if not isinstance(entries, list):
--- a/tests/test_user_config.py
+++ b/tests/test_user_config.py
@@ -154,6 +154,24 @@ class TestLoadUserConfigs:
         assert configs is not None
         assert len(configs) == 0
 
+    def test_whitespace_only_name(self, tmp_path):
+        """Whitespace-only name is treated as missing."""
+        self._write_yaml(
+            tmp_path,
+            """\
+            users:
+              - telegram_id: 111
+                name: "   "
+            """,
+        )
+        with (
+            patch("kai.config._read_protected_yaml", return_value=None),
+            patch("kai.config.PROJECT_ROOT", tmp_path),
+        ):
+            configs = _load_user_configs()
+        assert configs is not None
+        assert len(configs) == 0
+
     def test_missing_name(self, tmp_path):
         """Entry without name is skipped."""
         self._write_yaml(


dcellison:
## Review by Kai

## PR Review: Add multi-user Phase 1

Overall this is a well-structured, well-tested PR. The backward-compatibility approach is sound. A few real issues:

---

### Warning

**1. Unhandled `ValueError` in `_resolve_chat_id` and `_handle_get_jobs`** (`webhook.py`)

`_resolve_chat_id` (line ~267) and `_handle_get_jobs` (line ~740) both call `int()` with no error handling:

```python
# _resolve_chat_id
return int(explicit)          # TypeError/ValueError if not numeric

# _handle_get_jobs  
chat_id = int(raw_chat_id)    # ValueError if query param is "abc"
```

A malformed request propagates an unhandled exception → 500 instead of 400. These internal endpoints are authenticated, but the callers (inner Claude, test scripts) can still pass bad data. Fix: wrap with `try/except (ValueError, TypeError)` and return `web.json_response({"error": "..."}, status=400)`.

**2. Authenticated cross-user routing without target validation** (`webhook.py`, `_resolve_chat_id`)

Any holder of the webhook secret can set `chat_id` in the JSON body to route messages to any user — including users they're not associated with. While this is an internal API, the fix is cheap for Phase 1: validate that `int(explicit)` is in `config.allowed_user_ids` before trusting it. Right now a misbehaving or prompt-injected Claude session for user A can spam user B.

---

### Suggestion

**3. Silent fallback when `users` key is absent** (`config.py`, `_load_user_configs` ~line 548)

```python
entries = data.get("users")
if not isinstance(entries, list):
    if entries is not None:   # ← no warning when entries IS None
        log.warning(...)
    return None
```

If users.yaml contains a valid dict but no `users` key (e.g., a typo like `user:` instead of `users:`), the function returns `None` silently and falls back to `ALLOWED_USER_IDS`. The `if entries is not None` guard suppresses the warning for the most common misconfiguration. Remove that guard or make it always warn when the file was found.

**4. Non-existent `home_workspace` drops the entire user entry** (`config.py`, ~line 593)

The description calls `home_workspace` a "schema-only field (not enforced in Phase 1)," yet a missing directory causes the user to be completely excluded. A user who moves machines or whose workspace isn't mounted loses access entirely. Consider logging a warning and setting `home_workspace=None` instead of `continue`.

**5. Float `chat_id` truncates silently** (`webhook.py`, `_resolve_chat_id`)

`int(12345.6)` returns `12345` without error, which could silently route to the wrong user. The fix for finding #1 above (explicit type check before conversion) handles this too.

---

### Clean

- `UserConfig` frozen dataclass, field validation, and duplicate handling in `_load_user_configs` are correct and well-tested.
- The `bool` guard on `telegram_id` and `max_budget` (preventing YAML `true`/`false` from coercing to `1`/`0`) is the right defense.
- `get_user_by_github` case-insensitive comparison is correct.
- `load_config()` backward-compatibility logic (users.yaml wins, warning logged) is clean.
- Test coverage for the new config code is thorough (428 lines, 29 new tests).

PR #98, opened 2026-03-19
Add multi-user Phase 2: per-user data isolation

### LINKED ISSUE #95: Multi-user Phase 2: per-user data isolation
## Problem

After Phase 1 introduces the user model and safe API routing, runtime state is still shared. Settings, history, file storage, and workspace state are not namespaced by user. Two users on the same instance can see each other's data.

## Solution

Isolate per-user state using real OS user accounts. Each user defined in `users.yaml` maps to an OS account (via the `os_user` field). File permissions enforce isolation at the kernel level rather than relying on application-level scoping.

Scope includes:
- Per-user home directories owned by their OS account
- Settings table namespaced by user_id
- Per-user history filtering
- Per-user file storage under their OS home
- Per-user workspace state
- Removal of the `.responding_to` flag file (replaced with a per-user mechanism)

OS-level isolation means a bug in Kai cannot leak User A's data to User B - the process literally cannot read the files. This is a stronger guarantee than application-level path/query scoping.

## Dependencies

Requires Phase 1 (#94) - the user model and `os_user` field must exist before isolation can be enforced.

dcellison:
## Summary

Namespaces all per-user state by chat_id so each user's Claude session only sees their own data. This is Phase 2 of the three-phase multi-user plan, building on Phase 1's user model and safe routing (PR #97).

### Five categories of isolation

**1. Settings namespacing** (`sessions.py`, `bot.py`, `main.py`)
- Workspace setting renamed from `"workspace"` to `"workspace:{chat_id}"`
- One-time migration at startup: old global key moved to the admin's namespaced key
- Voice settings were already namespaced (`voice_mode:{chat_id}`) - unchanged

**2. History filtering** (`history.py`, `claude.py`)
- `get_recent_history()` accepts optional `chat_id` parameter
- When provided, only messages from that user are returned
- Prevents cross-user conversation history from being injected into Claude sessions
- History files remain shared JSONL (Phase 3 handles OS-level access control)

**3. Per-user file storage** (`bot.py`, `claude.py`)
- `_save_to_workspace()` creates per-user subdirectories (`workspace/files/{user_id}/`)
- All four file handler call sites pass `user_id=chat_id`
- Claude's context injection updated to show the per-user file path
- No migration of existing shared files (backward-compatible)

**4. Per-user workspace history** (`sessions.py`, `bot.py`, `main.py`)
- `workspace_history` table gains `chat_id` column with composite primary key
- Schema evolution via `ALTER TABLE ADD COLUMN` for existing databases
- `INSERT OR REPLACE` handles both old (path-only PK) and new (path+chat_id) schema
- `backfill_workspace_history()` assigns pre-Phase-2 rows to admin user
- All callers (`_do_switch_workspace`, `handle_workspaces`, `handle_workspace_callback`) pass chat_id

**5. Per-user crash recovery** (`bot.py`, `main.py`)
- `.responding_to` single flag file replaced with `.responding/` directory
- Each in-flight response creates a `{chat_id}` file in the directory
- Multiple concurrent users tracked independently
- Old `.responding_to` file migrated at first startup

### Backward compatibility

- Single-user deployments without `users.yaml` work unchanged
- `get_recent_history(chat_id=None)` returns all messages
- `_save_to_workspace(..., user_id=None)` uses shared `files/` directory
- All migrations are one-time and idempotent

Fixes #95

## Test plan

- [x] Settings: per-user workspace isolation, migration, migration idempotent
- [x] History: filtered by chat_id, unfiltered returns all, empty for unknown user
- [x] Files: per-user subdirectory, shared fallback, cross-user isolation
- [x] Workspace history: per-user rows, same path different users, delete per-user, backfill, backfill idempotent
- [x] Crash recovery: multiple users in-flight, clear one leaves other, no-op on missing
- [x] Existing tests updated for new signatures
- [x] All 959 tests pass (17 new)

### DIFF
--- a/src/kai/main.py
+++ b/src/kai/main.py
@@ -135,7 +135,7 @@ def main() -> None:
             admins = config.get_admins()
             if admins:
                 default_chat_id = admins[0].telegram_id
-            elif config.user_configs:
+            else:
                 default_chat_id = next(iter(config.user_configs))
         elif config.allowed_user_ids:
             default_chat_id = next(iter(config.allowed_user_ids))


dcellison:
## Review by Kai

## PR Review: Multi-user Phase 2 per-user data isolation

Overall the implementation is solid and well-tested. Three issues worth addressing, ranked below.

---

### Warning: Composite primary key not enforced on upgraded databases

**`sessions.py`, schema migration block**

On a fresh install, `CREATE TABLE IF NOT EXISTS workspace_history` creates the correct composite PK `(path, chat_id)`. On an **existing database**, that statement is a no-op and the table retains `path TEXT PRIMARY KEY`. The subsequent `ALTER TABLE ADD COLUMN chat_id` adds the column but SQLite has no `ALTER TABLE ... DROP CONSTRAINT` — the old single-column PK stays in effect.

Consequence: `INSERT OR REPLACE INTO workspace_history (path, chat_id, ...)` with the same `path` but a different `chat_id` hits the old `path` PK and **replaces** the existing row instead of creating a second one. Two users visiting the same workspace path on an upgraded instance will silently stomp each other's history entries.

The comment "INSERT OR REPLACE works with both old schema and new schema" is incorrect for this case.

Fix: on existing DBs, recreate the table (copy → drop → create new → re-insert), or use `CREATE TABLE workspace_history_new ... INSERT INTO ... FROM ... DROP ... ALTER TABLE RENAME`.) The migration is already wrapped in a column-existence check, so the scope is contained.

---

### Warning: Pre-migration history invisible after upgrade

**`history.py`, `get_recent_history` filter**

```python
if chat_id is not None and record.get("chat_id") != chat_id:
    continue
```

All JSONL records written before Phase 2 have no `chat_id` field, so `record.get("chat_id")` returns `None`. When called with a real `chat_id`, every pre-migration message is silently dropped. After upgrading, Claude loses all conversation history context for the admin/default user on the very first filtered call.

If this is intentional and will be addressed in Phase 3, add a comment explicitly acknowledging it. If not, old records without `chat_id` should be treated as belonging to the default user (matching the backfill logic applied to workspace history).

---

### Suggestion: Redundant exception type in crash-recovery loop

**`main.py`, responding dir iteration (~line 248)**

```python
except (ValueError, Exception):
```

`ValueError` is a subclass of `Exception`; listing it explicitly is a no-op. Both cases log the same message, so this should just be `except Exception:`. Minor, but it suggests the intent was to handle `ValueError` (bad `int(flag.name)`) differently, which it currently doesn't.

---

### Note: Silent bug fix bundled in

**`sessions.py`, original `delete_workspace_history`**

The pre-PR version of `delete_workspace_history` was missing `await _get_db().commit()`. The PR quietly adds it. Fine to include, but worth calling out so reviewers know the deletion behavior changed.

---

### Note: `change_workspace` now passes `workspace_config` on startup

**`main.py` workspace restoration**

```python
ws_config = config.get_workspace_config(ws_path)
await app.bot_data["claude"].change_workspace(ws_path, workspace_config=ws_config)
```

The original only passed `ws_path`. This is likely a correct fix, but it's out of Phase 2's stated scope. If this was previously a latent bug, it deserves a comment explaining why `workspace_config` is now necessary.

PR #114, opened 2026-03-22
Fix missing chat_id in _resolve_local_repo workspace history lookup

### LINKED ISSUE #109: Fix missing chat_id argument in _resolve_local_repo workspace history lookup
## Summary

`_resolve_local_repo()` in `webhook.py` calls `sessions.get_workspace_history(limit=50)` on line 221, but `get_workspace_history()` requires `chat_id` as its first positional argument. This means `limit=50` is silently passed as `chat_id`, querying workspace history for a nonexistent user 50 instead of searching actual workspace history.

This breaks PR review routing for any repository that isn't resolved by the first three steps (home workspace, WORKSPACE_BASE scan, ALLOWED_WORKSPACES). Step 4 (workspace history lookup) always fails.

## Root Cause

The function signature in `sessions.py:448` is:

```python
async def get_workspace_history(chat_id: int, limit: int = 10) -> list[dict]:
```

The call at `webhook.py:221` is:

```python
history = await sessions.get_workspace_history(limit=50)
```

This positionally passes `50` as `chat_id`, not `limit`.

## Context

`_resolve_local_repo` is called from `_handle_github_webhook` (line 555) which processes server-to-server GitHub webhook payloads. There is no specific user context available - the function signature is `_resolve_local_repo(repo_full_name: str, app: web.Application)` with no `chat_id` parameter.

The intent of step 4 is to search across all users' workspace history to find a matching repo path. Two options:

1. **Pass `app["chat_id"]` (admin)** - only searches admin's workspace history, misses repos visited by other users
2. **Add a new query** - e.g. `get_all_workspace_history()` without the `WHERE chat_id = ?` filter, or iterate all known user IDs

Option 1 is the quick fix; option 2 is the correct multi-user fix.

## Reproduction

1. Configure a second user with a workspace that is NOT the home workspace, NOT under WORKSPACE_BASE, and NOT in ALLOWED_WORKSPACES
2. Have that user visit the workspace (so it appears in workspace_history)
3. Push a PR to that repo
4. The review agent fails to resolve the local repo path

## Severity

**High** - currently broken at runtime for any repo relying on workspace history resolution.

dcellison:
## Summary

Fixes a Phase 2 regression where `_resolve_local_repo()` in `webhook.py` called `get_workspace_history(limit=50)` - but `get_workspace_history` requires `chat_id` as its first positional argument. Python silently bound `50` as `chat_id`, causing the query to run `WHERE chat_id = 50` (matching no real user). Step 4 of repo resolution (workspace history lookup) always returned an empty list.

### The bug

```python
# webhook.py:221 (before)
history = await sessions.get_workspace_history(limit=50)

# sessions.py signature
async def get_workspace_history(chat_id: int, limit: int = 10) -> list[dict]:
```

Python matches `limit=50` as the positional `chat_id` argument. The query runs `WHERE chat_id = 50`, which matches no real user. PR reviews for repos outside the home/base/allowed workspaces silently skip spec and convention loading.

### The fix

Added `get_all_workspace_paths()` to `sessions.py` - a cross-user query (`SELECT DISTINCT path FROM workspace_history`) that intentionally omits the `chat_id` filter. GitHub webhooks are server-to-server with no user context, so the resolution needs to search ALL users' workspace history.

The existing `get_workspace_history(chat_id, limit)` is unchanged - it's correct for its per-user purpose (the `/workspaces` command in bot.py).

### Why not just pass a chat_id?

`_resolve_local_repo` is called from `_handle_github`, which processes GitHub webhook payloads. There is no Telegram user associated with the request. The function's job is to find ANY local checkout of the repo, regardless of which user visited it. A per-user query would be wrong here.

Fixes #109

## Test plan

- [x] New `get_all_workspace_paths` tests: multi-user paths, deduplication, limit, ordering, empty
- [x] Existing webhook tests updated: mock target changed to `get_all_workspace_paths`, return values changed from `list[dict]` to `list[str]`
- [x] New cross-user resolution test: repo found via another user's workspace history
- [x] All 1009 tests pass (6 new)

### DIFF
--- a/src/kai/sessions.py
+++ b/src/kai/sessions.py
@@ -458,8 +458,11 @@ async def get_all_workspace_paths(limit: int = 100) -> list[str]:
     Returns:
         List of workspace path strings (deduplicated across users).
     """
+    # GROUP BY + MAX(last_used_at) instead of DISTINCT to get
+    # deterministic ordering when the same path appears for multiple
+    # users with different timestamps.
     async with _get_db().execute(
-        "SELECT DISTINCT path FROM workspace_history ORDER BY last_used_at DESC LIMIT ?",
+        "SELECT path FROM workspace_history GROUP BY path ORDER BY MAX(last_used_at) DESC LIMIT ?",
         (limit,),
     ) as cursor:
         rows = await cursor.fetchall()
--- a/tests/test_sessions.py
+++ b/tests/test_sessions.py
@@ -427,9 +427,10 @@ class TestGetAllWorkspacePaths:
     @pytest.mark.asyncio
     async def test_most_recent_first(self):
         """Paths are ordered by most recently used."""
-        # Use explicit timestamps to guarantee ordering (CURRENT_TIMESTAMP
-        # can be identical for rapid inserts within the same second).
-        db = sessions._get_db()
+        # Use explicit timestamps via raw SQL to guarantee ordering.
+        # CURRENT_TIMESTAMP can be identical for rapid inserts within
+        # the same second, making the ordering test non-deterministic.
+        db = sessions._get_db()  # test-only access for timestamp control
         await db.execute(
             "INSERT OR REPLACE INTO workspace_history (path, chat_id, last_used_at) VALUES (?, ?, ?)",
             ("/old", 111, "2026-01-01 00:00:00"),


dcellison:
## Review by Kai

## PR Review: Fix missing chat_id in `_resolve_local_repo`

The one actionable item from the prior review — the SQL `DISTINCT` + non-selected `ORDER BY` column — is correctly resolved.

---

### Prior warning: addressed

**`sessions.py`, `get_all_workspace_paths`:**

```sql
-- prior (non-deterministic for shared paths)
SELECT DISTINCT path FROM workspace_history ORDER BY last_used_at DESC LIMIT ?

-- now (correct)
SELECT path FROM workspace_history GROUP BY path ORDER BY MAX(last_used_at) DESC LIMIT ?
```

`GROUP BY path ORDER BY MAX(last_used_at)` is the right fix. Deterministic ordering for cross-user shared paths, matches the docstring's claim.

---

### No new issues

- **`webhook.py:220-226`** — call site is clean; iterates `list[str]` directly, `Path(path_str)` is correct.
- **`test_sessions.py`, `test_most_recent_first`** — `_get_db()` access is still present (prior suggestion, author's call). The explanatory comment is adequate.
- **`test_webhook.py`, `test_history_searches_all_users`** — correctly exercises cross-user resolution via the new function.
- Parameterized queries throughout — no injection risk.
- No auth bypass: cross-user workspace path lookup in a server-to-server webhook context is intentional and documented.

---

**Clean. Ready to merge.**

PR #136, opened 2026-03-23
Validate schedule_data before persisting to database

### LINKED ISSUE #129: schedule_data string not validated as JSON before persisting
## Summary

When `schedule_data` arrives as a pre-serialized string (rather than a JSON object) in `POST /api/schedule` or `PATCH /api/jobs/{id}`, it is stored in the database without validating that it is valid JSON. A malformed string (e.g., `"not json at all"`) gets persisted, and `cron.py` crashes with a `json.JSONDecodeError` on every subsequent fire attempt for that job.

## Affected endpoints

Both `_handle_schedule` (line 743-747) and `_handle_update_job` (line 903-906) in `webhook.py` handle `schedule_data` the same way:

```python
if isinstance(schedule_data, dict):
    schedule_data_str = json.dumps(schedule_data)
else:
    schedule_data_str = schedule_data  # stored as-is, no validation
```

When the caller sends a dict, `json.dumps` produces valid JSON. When the caller sends a string, it passes through unchecked.

## Downstream crash

`cron.py` parses `schedule_data` with `json.loads()` at two points:

- `_register_new_jobs` (line 124): startup registration of all active jobs
- `_register_job` (line 173): on-demand registration after job creation/update

A malformed `schedule_data` string causes `json.JSONDecodeError` at these call sites. For recurring jobs (`daily`, `interval`), this crashes on every fire attempt for the lifetime of the job. The job stays `active=1` in the database and keeps failing on every scheduler tick until manually deleted.

## Additional structural validation gap

Even when `schedule_data` is valid JSON, its contents are not validated against the `schedule_type`. For example:
- `schedule_type: "daily"` with `schedule_data: {"seconds": 60}` (wrong shape - should have `"times"`)
- `schedule_type: "once"` with `schedule_data: {}` (missing required `"run_at"` field)
- `schedule_type: "interval"` with `schedule_data: {"seconds": "not a number"}` (wrong type)

These also crash in `cron.py` at registration time, with `KeyError` or `TypeError` instead of `JSONDecodeError`.

## Severity

MEDIUM - a single malformed API call creates a permanently broken recurring job that crashes on every fire. The API is localhost-only behind webhook secret, so the caller is always inner Claude, but prompt injection could produce malformed payloads.

dcellison:
## Summary

- Add `_validate_schedule_data()` helper that validates both JSON syntax and structural correctness for `schedule_data` payloads
- Applied to both `POST /api/schedule` and `PATCH /api/jobs/{id}`
- Previously, a malformed string (e.g., `"not json"`) or wrong-shape payload (e.g., interval keys on a daily job) would persist to the database and crash `cron.py` with `JSONDecodeError` or `KeyError` on every fire attempt
- For PATCH, when `schedule_data` is provided without `schedule_type`, the existing job is fetched to get the current type for validation

## Validation rules

| schedule_type | Required key | Additional checks |
|---|---|---|
| `once` | `run_at` | Must parse as ISO datetime |
| `daily` | `times` | Must be a non-empty list of valid `HH:MM` strings (00-23:00-59) |
| `interval` | `seconds` | Must be a positive number |

All types: string inputs must be valid JSON objects. Dict inputs are validated structurally then serialized.

## Test plan

- [x] Invalid JSON string returns 400
- [x] Missing `run_at` for `once` returns 400
- [x] Invalid ISO datetime for `once` returns 400
- [x] Missing `times` for `daily` returns 400
- [x] Invalid `HH:MM` format (e.g., `25:00`) returns 400
- [x] Missing `seconds` for `interval` returns 400
- [x] Negative `seconds` returns 400
- [x] PATCH with invalid `schedule_data` returns 400
- [x] Existing `test_string_schedule_data_passed_through` (valid JSON string) still passes
- [x] All 1047 tests pass, `make check` clean

Fixes #129

### DIFF
--- a/src/kai/webhook.py
+++ b/src/kai/webhook.py
@@ -824,6 +824,7 @@ async def _handle_schedule(request: web.Request) -> web.Response:
     schedule_data_str, error = _validate_schedule_data(schedule_data, schedule_type)
     if error:
         return web.json_response({"error": error}, status=400)
+    assert schedule_data_str is not None  # guaranteed when error is None
 
     # Persist to database
     try:


dcellison:
## Review by Kai

All five issues from the prior review are addressed. The PR is in good shape — one new minor finding:

---

### Suggestion

**`float('nan')` / `float('inf')` bypass `seconds` validation** — `webhook.py`, `_validate_schedule_data`, interval branch

```python
if isinstance(seconds, bool) or not isinstance(seconds, (int, float)) or seconds <= 0:
```

`float('nan') <= 0` evaluates to `False` in Python (all NaN comparisons return False), so NaN slips through. `float('inf')` is also accepted since `inf > 0`. Both are `float` instances that would produce nonsensical cron behavior.

In practice these can only arrive via a Python dict (not through a JSON-parsed request body, since standard JSON doesn't support `Infinity`/`NaN`), so the practical risk is low. A simple fix:

```python
import math
if isinstance(seconds, bool) or not isinstance(seconds, (int, float)) \
        or not math.isfinite(seconds) or seconds <= 0:
```

---

Otherwise the implementation is solid: the `else` clause for unknown types, the dict guard in the non-string branch, the `bool` subclass check, the `len(parts) != 2` guard for `HH:MM`, and the `isinstance(run_at, str)` guard were all correctly applied. The PATCH `effective_type` lookup is handled correctly. Test coverage matches the added logic.

PR #139, opened 2026-03-23
Kill sudo child tree on review/triage timeout

### LINKED ISSUE #132: proc.kill() on sudo subprocess orphans claude child process
## Summary

When `claude_user` is configured, both `review.py` and `triage.py` spawn their Claude subprocesses under `sudo -u`. On timeout, they call `proc.kill()`, which sends SIGKILL only to the `sudo` process. The `claude` Node.js child spawned by sudo can survive as an orphan, reparented to init/launchd.

## Details

Both files use the same pattern:

```python
if claude_user:
    cmd = ["sudo", "-u", claude_user, "--"] + cmd

proc = await asyncio.create_subprocess_exec(*cmd, ...)

try:
    stdout, stderr = await asyncio.wait_for(proc.communicate(...), timeout=...)
except TimeoutError:
    proc.kill()       # Only kills sudo, not its children
    await proc.wait()
```

`proc.kill()` sends SIGKILL to the direct child process, which is `sudo`. When sudo receives SIGKILL, it dies immediately without forwarding the signal to the `claude` process it spawned. The `claude` Node.js process gets reparented to PID 1 and continues running with no parent tracking it.

The persistent Claude subprocess in `claude.py` already solved this problem: it uses `start_new_session=True` when spawning under sudo, which puts the entire process tree (sudo + claude) in a new process group. It then uses `os.killpg()` to kill the entire group. The review and triage agents don't use either of these mechanisms.

## Affected locations

- `review.py:658` - timeout handler in `run_review_subprocess()`
- `triage.py:373` - timeout handler in `run_triage_subprocess()`

## Impact

Each orphaned `claude` process is a full Node.js V8 instance consuming memory and potentially CPU. On the 16GB Mac mini, accumulated orphans from repeated review/triage timeouts could cause memory pressure. The orphans also hold open any file handles or network connections the claude process had at the time of the timeout.

This only affects deployments where `claude_user` is set (the sudo path). Without `claude_user`, `proc.kill()` hits the `claude` process directly and works correctly.

## Severity

MEDIUM - from the code review document.

dcellison:
## Summary

- Port the `start_new_session` + `os.killpg()` pattern from `claude.py` to the review and triage subprocess spawners
- When `claude_user` is set, `proc.kill()` only kills the `sudo` wrapper, orphaning the `claude` Node.js child as a zombie reparented to PID 1
- With `start_new_session=True`, the entire tree (sudo + claude) shares a process group, and `os.killpg()` reaches all of them on timeout
- No behavior change for the non-sudo path (direct `proc.kill()` as before)

## Changes

| File | Change |
|------|--------|
| `src/kai/review.py` | Add `start_new_session=bool(claude_user)` to subprocess spawn; use `os.killpg()` on timeout when `claude_user` is set |
| `src/kai/triage.py` | Same pattern |
| `tests/test_review.py` | Tests for group kill on timeout, `start_new_session` flag, and no-session for non-sudo |
| `tests/test_triage.py` | Same pattern |

## Test plan

- [x] `test_timeout_with_claude_user_kills_group` - verifies `os.killpg(pid, SIGKILL)` called on timeout with `claude_user`
- [x] `test_claude_user_starts_new_session` - verifies `start_new_session=True` passed to subprocess
- [x] `test_no_claude_user_no_new_session` - verifies `start_new_session=False` without `claude_user`
- [x] Existing `test_timeout_raises` still passes (non-sudo path unchanged)
- [x] All 1067 tests pass, `make check` clean

Fixes #132

### DIFF
--- a/tests/test_triage.py
+++ b/tests/test_triage.py
@@ -302,6 +302,7 @@ class TestRunTriage:
             await run_triage("test prompt", claude_user="kai")
 
         mock_killpg.assert_called_once_with(12345, signal.SIGKILL)
+        mock_proc.wait.assert_called_once()
 
     @pytest.mark.asyncio
     async def test_claude_user_starts_new_session(self):


dcellison:
## Review by Kai

**Overall: Clean and correct implementation. One notable gap worth fixing.**

---

### Warning — `except OSError` is too broad (`review.py:668`, `triage.py:383`)

```python
except OSError:
    pass  # Already dead
```

`os.killpg()` raises `OSError` for multiple reasons:

| errno | Exception subclass | Meaning |
|---|---|---|
| ESRCH (3) | `ProcessLookupError` | Process group gone — safe to ignore |
| EPERM (1) | `PermissionError` | Group exists but we can't kill it — **orphan survives** |
| EINVAL (22) | `OSError` | Bad signal number |

Swallowing `PermissionError` silently leaves the claude process running, contradicts the fix's purpose, and then causes `proc.wait()` to hang indefinitely (waiting for a process that was never killed).

Fix:
```python
except ProcessLookupError:
    pass  # Already dead
```

Or at minimum, log on unexpected errno:
```python
except OSError as e:
    if e.errno != errno.ESRCH:
        logger.warning("killpg(%s) failed: %s", proc.pid, e)
```

---

### Suggestion — `test_triage.py` missing `test_no_claude_user_no_new_session`

`test_review.py` has a symmetric `test_no_claude_user_no_new_session` test but `test_triage.py` doesn't. Minor coverage gap, easy to add.

---

### Suggestion — No log on kill failure

Both files silently swallow the `OSError` with only a comment. A `logger.warning()` would make production issues diagnosable without a code change.

---

- PGID == PID assumption after `start_new_session=True` is valid (Linux/macOS `setsid()` semantics).
- `proc.wait()` after `killpg()` correctly reaps the subprocess.
- `bool(claude_user)` as the flag is clean.
- Test structure mirrors the implementation faithfully.
- No security concerns — `claude_user` is a trusted config value, not user input.

PR #141, opened 2026-03-23
Escape glob metacharacters in branch name spec lookup

### LINKED ISSUE #134: Glob injection via attacker-controlled branch names in spec lookup
## Summary

In `review.py`, the `find_spec_for_branch()` function interpolates the branch name directly into a glob pattern without escaping. Branch names are attacker-controlled (any GitHub user who can open a PR chooses the branch name). Glob metacharacters in the name can match unintended spec files.

## Details

The relevant code in `find_spec_for_branch()`:

```python
name = branch.split("/", 1)[-1] if "/" in branch else branch
pattern = str(specs_dir / f"*{name}*.md")
matches = sorted(glob_mod.glob(pattern))
```

The `name` portion is derived from the PR's head branch. If an attacker creates a branch like `feature/*`, the glob pattern becomes `*/*/*.md` (after prefix stripping), which matches across subdirectories. Other metacharacters:

- `?` matches any single character - `feature/t?st` could match `test.md`, `tost.md`, etc.
- `[abc]` matches character classes - `feature/spec[0-9]` matches `spec0.md` through `spec9.md`
- `*` matches everything - `feature/*` matches all `.md` files in the specs directory

The function returns the first sorted match, so an attacker can influence which spec file gets loaded into the review prompt. The spec provides context to the review agent; loading the wrong spec would produce a review based on incorrect requirements.

## Scope

This only affects repos where:
1. A `workspace/specs/` directory (or custom `spec_dir`) exists with spec files
2. The PR review agent is enabled
3. An external contributor can open a PR with an arbitrary branch name

For the Kai repo itself, branch protection means only collaborators can push branches, limiting the attack surface. But the review agent is designed to work on any configured repo, including those with open contribution models.

## Severity

MEDIUM - from the code review document.

dcellison:
## Summary

- Apply `glob.escape()` to the branch name before interpolating it into the glob pattern in `resolve_spec_from_branch()`
- Branch names are attacker-controlled (any PR author chooses theirs), and glob metacharacters (`*`, `?`, `[]`) could match unintended spec files
- The matched spec is loaded into the review prompt, so an attacker could influence what requirements the review agent evaluates against
- One line added (`safe_name = glob_mod.escape(name)`), one line changed (use `safe_name` in the pattern)

## Test plan

- [x] `test_glob_metachar_star_escaped` - `feature/*` does not match all spec files
- [x] `test_glob_metachar_question_escaped` - `feature/t?st` does not match `test-spec.md`
- [x] `test_glob_metachar_bracket_escaped` - `feature/spec[0-9]` does not match `spec1.md`
- [x] `test_normal_branch_still_matches_after_escaping` - normal branches still work
- [x] `test_literal_bracket_in_filename_still_matches` - escaped `[` in branch name matches literal `[` in filename
- [x] All existing `TestResolveSpecFromBranch` tests pass unchanged
- [x] All 1074 tests pass, `make check` clean

Fixes #134

### DIFF
--- a/tests/test_review.py
+++ b/tests/test_review.py
@@ -1092,6 +1092,19 @@ class TestResolveSpecFromBranch:
         result = resolve_spec_from_branch("feature/add-user-auth", str(tmp_path), spec_dir="workspace/specs")
         assert result == str(spec_file)
 
+    def test_literal_bracket_in_filename_still_matches(self, tmp_path):
+        """A spec file with literal [ in its name matches an escaped branch."""
+        specs_dir = tmp_path / "workspace" / "specs"
+        specs_dir.mkdir(parents=True)
+        # [ is a glob metachar AND a valid filename on all platforms
+        spec_file = specs_dir / "fix-[wip]-auth.md"
+        spec_file.write_text("content")
+
+        # Without escaping, [wip] would be a character class matching w/i/p.
+        # With escaping, it matches the literal [ and ].
+        result = resolve_spec_from_branch("feature/fix-[wip]-auth", str(tmp_path), spec_dir="workspace/specs")
+        assert result == str(spec_file)
+
 
 # ── load_spec ──────────────────────────────────────────────────────
 


dcellison:
## Review by Kai

## PR Review: Escape glob metacharacters in branch name spec lookup

**Overall: Fix is correct. One test coverage regression from addressing the prior review.**

---

### Warning — `test_literal_branch_with_special_chars_still_matches` was dropped rather than guarded

`tests/test_review.py`

The prior review flagged that test with a `@pytest.mark.skipif(sys.platform == "win32", ...)` suggestion. Instead, the test was silently removed and replaced with `test_normal_branch_still_matches_after_escaping`, which uses a plain branch name (`add-user-auth`) with no metacharacters.

These two tests are not equivalent. The dropped test verified the critical positive case: that `glob.escape("what-is-this?")` produces `"what-is-this[?]"`, which still matches a file literally named `what-is-this?.md`. The replacement test only confirms that escaping doesn't break ordinary names — something that would be obvious even if the escape logic were completely wrong.

The PR description still lists `test_literal_branch_with_special_chars_still_matches` as passing, which is now stale.

The correct fix is what the prior review suggested:

```python
@pytest.mark.skipif(sys.platform == "win32", reason="? not valid in Windows filenames")
def test_literal_branch_with_special_chars_still_matches(self, tmp_path):
    ...
```

---

### The fix itself is correct

- `glob.escape()` (available since Python 3.4) wraps `*`, `?`, `[` in `[...]` brackets, so user-controlled input is treated as literal while the surrounding `*` wildcards in `f"*{safe_name}*.md"` remain functional.
- The three "must return None" tests correctly demonstrate the escape is active for each metacharacter class.
- The pre-existing path traversal concern (`../../sensitive`) was noted in the prior review and the author chose not to address it here — that's a reasonable scope decision.

dcellison:
## Review by Kai

## PR Review: Escape glob metacharacters in branch name spec lookup

**Overall: Fix is correct, tests are adequate. One stale documentation nit.**

---

### Nit — PR description still lists the removed test as passing

The checklist in the PR description still shows:

> `[x] test_literal_branch_with_special_chars_still_matches`

That test was removed in the previous round and does not appear in this diff. The description should be updated to reflect the current test names. Low stakes, but it makes the checklist misleading for reviewers or anyone reading the issue later.

---

### The fix is correct

`review.py` +183–184: `glob.escape()` wraps `*`, `?`, and `[` in `[...]` brackets, making the user-controlled fragment fully literal while leaving the surrounding `*` wildcards in `f"*{safe_name}*.md"` functional. This is the correct tool for this job.

---

### The test suite is now adequate

The prior concern about missing a positive-case test for metacharacters has been addressed in a cross-platform way: `test_literal_bracket_in_filename_still_matches` uses `[` (valid in filenames on all platforms) to demonstrate that `glob.escape("fix-[wip]-auth")` → `"fix-[[]wip]-auth"` still matches `fix-[wip]-auth.md`. That covers the critical "escape doesn't over-sanitize" property without requiring a Windows skip. The three negative tests cover all three metacharacter classes. This is a reasonable design.

---

Fix the stale PR description and this is ready to merge.

PR #158, opened 2026-03-24
Fix pool eviction TOCTOU race condition

### LINKED ISSUE #153: Pool eviction TOCTOU race condition
## Problem

The `_eviction_loop()` in `pool.py` has a time-of-check/time-of-use race condition. The eviction candidate list is built as a snapshot, but the pool state can change before eviction actually runs.

## Details

**File:** `src/kai/pool.py`, lines 286-303

The eviction loop works in two phases:

1. **Build candidate list (lines 286-290):** Iterates `self._last_activity` and checks three conditions per entry:
   - `now - last > idle_timeout` (idle threshold exceeded)
   - `chat_id in self._pool` (instance exists)
   - `chat_id not in self._in_flight` (not actively streaming)

2. **Evict candidates (lines 291-303):** Loops through the candidate list, pops entries from `self._pool` and `self._last_activity`, and calls `instance.shutdown()`.

Between phases 1 and 2, other coroutines can:
- Call `self.get(chat_id)`, adding entries to `self._pool` and refreshing `self._last_activity`
- Call `self.send(chat_id)`, adding `chat_id` to `self._in_flight`
- Call `force_kill(chat_id)`, removing entries from both dicts entirely

The partial re-check at line 294 (`if self._last_activity.get(chat_id, 0) > now`) only guards against refreshed activity timestamps. It does not re-verify that the instance is still in `self._pool` or that the chat is not now in `self._in_flight`.

## Consequences

- A subprocess that just received a new message could be evicted mid-conversation
- A subprocess already removed by `force_kill()` could be double-popped (harmless due to `.pop(None)` but semantically wrong)
- A subprocess that entered `_in_flight` after the snapshot could be shut down while actively streaming a response

## Reproduction

Difficult to trigger in practice because the eviction interval is long relative to the TOCTOU window. But under load with many concurrent users, the window is real - especially if `instance.shutdown()` for an earlier candidate in the loop `await`s long enough for new activity to arrive on a later candidate.

## Severity

Medium. The window is narrow in typical usage but the consequences (killing an active conversation) are disruptive.

## Origin

Code review finding #15 (2026-03-22).

dcellison:
## Summary

Expand the TOCTOU re-check in `_eviction_loop()` to verify all three snapshot conditions before evicting an idle subprocess.

The eviction loop builds a snapshot of candidates, then iterates with `await instance.shutdown()` between each one. Those yield points let other coroutines change pool state: refreshing activity timestamps, entering `send()`, or calling `force_kill()`. The existing re-check only guarded against timestamp refreshes.

## Changes

Two new guards added before the `_pool.pop()`:

| Check | Catches | Consequence without it |
|-------|---------|----------------------|
| `chat_id in self._in_flight` | User entered `send()` after snapshot | Active conversation killed mid-stream |
| `chat_id not in self._pool` | `force_kill()` removed instance after snapshot | Misleading "Evicting" log for gone instance |
| `_last_activity > now` (existing) | User sent a message after snapshot | Idle user evicted despite recent activity |

## Test plan

- [x] `test_toctou_guard_skips_in_flight` - user entering send() after snapshot is not evicted
- [x] `test_toctou_guard_skips_removed_from_pool` - user removed by force_kill() is not evicted
- [x] `test_toctou_guard_skips_recently_active` (existing) - user with refreshed activity is not evicted
- [x] `test_eviction_proceeds_when_all_checks_pass` - user passing all three checks is evicted normally
- [x] All 1093 tests pass, `make check` clean

Fixes #153

### DIFF
--- a/tests/test_pool.py
+++ b/tests/test_pool.py
@@ -390,6 +390,8 @@ class TestEvictionTOCTOU:
         config = _make_config(claude_idle_timeout=1)
         pool = SubprocessPool(config=config, services_info=[])
 
+        # get() order determines to_evict order (dict insertion order); 111 must
+        # be processed first so its shutdown side effect modifies 222's state.
         a = pool.get(111)
         pool.get(222)
         pool._last_activity[111] = time.monotonic() - 10


dcellison:
## Review by Kai

## Review: Fix pool eviction TOCTOU race condition (rev 2)

The prior review's two concerns — tests not exercising the eviction loop and the `_last_activity` memory leak on pool-removal skip — are both addressed. The production fix was already correct; the main improvement here is in tests.

### Tracing the tests

The rewritten tests now correctly invoke `_eviction_loop()` and assert on pool membership as the observable outcome:

- `test_toctou_guard_skips_in_flight`: 111's mocked `shutdown()` adds 222 to `_in_flight`; loop reaches 222, guard fires, 222 stays in pool. ✓
- `test_toctou_guard_skips_removed_from_pool`: 111's mocked `shutdown()` removes 222 from `_pool`; loop reaches 222, pool-membership guard fires and cleans up `_last_activity[222]`. ✓
- `test_eviction_proceeds_when_all_checks_pass`: single user passes all three re-checks and is evicted. ✓

These now match the pattern of the pre-existing `test_toctou_guard_skips_recently_active`.

---

### Suggestion — `_last_activity` orphaned when activity-refresh and force_kill race

**`pool.py`, guard ordering**

The `_last_activity > now` check fires before `not in _pool`. If both conditions are simultaneously true — user's activity was refreshed AND `force_kill()` removed them from `_pool` between snapshot and iteration — the first check fires a `continue` before the pool-membership cleanup runs, orphaning the `_last_activity` entry permanently.

The snapshot filter (`chat_id in self._pool`) prevents the orphaned entry from ever becoming a future eviction candidate, so this isn't a functional bug, but the entry leaks for the pool's lifetime. Reordering to check `not in _pool` first would close this:

```python
if chat_id not in self._pool:
    self._last_activity.pop(chat_id, None)
    continue
if chat_id in self._in_flight:
    continue
if self._last_activity.get(chat_id, 0) > now:
    continue
```

---

### Suggestion — Test TOCTOU ordering relies on undocumented dict insertion order

**`tests/test_pool.py`, ~354 and ~393**

Both multi-user tests rely on the eviction loop visiting 111 before 222 (so 111's shutdown is the yield point that mutates 222's state). This works because `pool.get(111)` is called before `pool.get(222)`, and Python 3.7+ guarantees dict insertion order. The implicit dependency is not obvious to a future maintainer. A one-line comment would help:

```python
# get() order determines to_evict order; 111 must be processed first
a = pool.get(111)
pool.get(222)
```

---

### Clean

- No security issues.
- No functional bugs in `pool.py`.
- The three re-checks have no `await` points between them and `_pool.pop()` — no new TOCTOU window introduced.
- `_in_flight` not cleaned up on skip is intentional and correct (user is still live).

PR #159, opened 2026-03-24
Fix eviction silent failure and orphaned subprocesses

### LINKED ISSUE #154: Lock eviction can fail silently, orphaning subprocesses
## Problem

In `pool.py`, both the eviction loop and `force_kill()` remove instances from tracking dictionaries *before* attempting shutdown. If shutdown fails, the subprocess is orphaned - still running but no longer tracked by the pool.

## Details

**File:** `src/kai/pool.py`

### Eviction loop (lines 296-303)

```python
instance = self._pool.pop(chat_id, None)
self._last_activity.pop(chat_id, None)
if instance and instance.is_alive:
    try:
        await instance.shutdown()
    except Exception:
        log.exception("Error evicting subprocess for user %d", chat_id)
```

The instance is popped from `self._pool` and `self._last_activity` at lines 296-297. If `instance.shutdown()` raises at line 301, the exception is logged at line 303, but execution continues normally. The subprocess may still be running, but the pool no longer knows about it.

### force_kill() (lines 213-224)

```python
instance = self._pool.pop(chat_id, None)
self._last_activity.pop(chat_id, None)
if instance:
    try:
        await asyncio.wait_for(instance.shutdown(), timeout=_FORCE_KILL_TIMEOUT)
    except TimeoutError:
        instance.force_kill()
```

Same pattern: pop first, then attempt shutdown. Only `TimeoutError` is caught. Any other exception (`OSError`, `RuntimeError`, `asyncio.CancelledError`, etc.) propagates uncaught, crashing the caller - typically the `/stop` command handler in bot.py. The instance is already gone from tracking, so there is no recovery path.

## Consequences

**Eviction loop path:**
- Failed shutdowns are logged but otherwise ignored
- Orphaned subprocesses accumulate over time in a long-running service
- No mechanism exists to retry cleanup or even detect orphaned processes

**force_kill() path:**
- Non-timeout exceptions crash the calling handler
- The user's `/stop` command fails with an unhandled error
- The subprocess is removed from the pool but may still be running

**Both paths:**
- Pool state diverges from actual process state
- `len(self._pool)` underreports active subprocesses
- Capacity calculations based on pool size become incorrect

## Severity

Medium. The eviction path silently leaks resources; the force_kill path can crash user-facing commands.

## Origin

Code review finding #19 (2026-03-22).

dcellison:
## Summary

Fix two resource management bugs in `pool.py` where subprocess instances were removed from tracking before shutdown completed, causing orphaned processes on failure.

### Problem 1: Pop-before-shutdown ordering

Both `_eviction_loop()` and `force_kill()` called `self._pool.pop()` before `instance.shutdown()`. If shutdown failed, the subprocess was still running but invisible to the pool - no retry, no cleanup, just a silent memory leak.

### Problem 2: Incomplete exception handling in force_kill()

`force_kill()` only caught `TimeoutError`. Any other exception from `shutdown()` (`CancelledError`, `OSError`, etc.) propagated to the `/stop` handler, crashing the user's command while leaving the subprocess untrackable.

### Fix

- **Defer pop until after shutdown:** Instance stays in the pool during shutdown. Only removed after the subprocess is confirmed dead.
- **SIGKILL fallback on any failure:** Both paths now call `instance.force_kill()` (raw SIGKILL, effectively infallible) as a last resort before removing from tracking.
- **Catch all exceptions in force_kill():** Widen from `TimeoutError` to `Exception` so nothing propagates to callers.

## Changes

| Location | Before | After |
|----------|--------|-------|
| `force_kill()` | Pop first, catch only `TimeoutError` | Pop after shutdown, catch all `Exception` |
| `_eviction_loop()` | Pop first, log and move on | Pop after shutdown, SIGKILL fallback on failure |

## Test plan

- [x] `test_force_kill_catches_non_timeout_exceptions` - RuntimeError caught, SIGKILL fallback, instance removed
- [x] `test_force_kill_pops_after_successful_shutdown` - instance in pool during shutdown, removed after
- [x] `test_eviction_failure_triggers_sigkill_fallback` - failed shutdown triggers force_kill, instance not orphaned
- [x] `test_eviction_pops_after_shutdown` - instance in pool during eviction shutdown, removed after
- [x] Existing tests pass (force_kill specific user, timeout fallback, no instance, TOCTOU guards)
- [x] All 1097 tests pass, `make check` clean

Fixes #154

### DIFF
--- a/src/kai/pool.py
+++ b/src/kai/pool.py
@@ -208,8 +208,8 @@ class SubprocessPool:
 
         Uses shutdown() with a short timeout for clean process reaping
         and stderr task cancellation. Falls back to raw SIGKILL on any
-        failure (timeout, cancellation, or other exception), so this
-        never blocks callers for long and never propagates exceptions.
+        non-cancellation failure (timeout, OSError, etc.). Cleanup
+        (pool removal) runs unconditionally via finally.
 
         The instance is kept in the pool during shutdown so it remains
         tracked. It is only removed after the subprocess is confirmed


dcellison:
## Review by Kai

## Review: fix/eviction-silent-failure (round 2)

All three prior review issues have been addressed. The dead-instance leak in the eviction loop is fixed (pops moved outside the `if instance and instance.is_alive:` block). The `CancelledError` gap in `force_kill` is fixed via `finally`. The double-shutdown suggestion was not taken — that's the author's call.

One new issue:

---

### Warning — Eviction loop still doesn't protect against CancelledError

**`pool.py`, `_eviction_loop`, lines ~330–340**

`force_kill()` now correctly uses `finally` for the pop. The eviction loop does not:

```python
if instance and instance.is_alive:
    try:
        await instance.shutdown()
    except Exception:          # CancelledError escapes here
        instance.force_kill()
# NOT in a finally — skipped if CancelledError propagates
self._pool.pop(chat_id, None)
self._last_activity.pop(chat_id, None)
```

If `CancelledError` is raised during `await instance.shutdown()`, neither the SIGKILL fallback nor the cleanup pops execute. The instance stays in the pool and the subprocess may be left running. This is the same structural gap that was fixed in `force_kill()` but not carried over here.

Practical blast radius is narrow — this only fires if the eviction task is cancelled mid-shutdown, at which point `pool.shutdown()` presumably handles cleanup — but it's an inconsistency worth closing:

```python
try:
    if instance and instance.is_alive:
        try:
            await instance.shutdown()
        except Exception:
            log.exception("Error evicting subprocess for user %d, sending SIGKILL", chat_id)
            instance.force_kill()
finally:
    self._pool.pop(chat_id, None)
    self._last_activity.pop(chat_id, None)
```

---

Everything else looks clean. The `finally`-based cleanup in `force_kill`, the SIGKILL fallback placement, and the test coverage (particularly `check_pool_during_shutdown` verifying pool membership during the await) are all solid.

dcellison:
## Review by Kai

## Review: fix/eviction-silent-failure (round 3)

The round 2 warning (eviction loop `CancelledError` leaving the instance in the pool) is addressed. The outer `try/finally` in `_eviction_loop` now mirrors the fix in `force_kill`, so cleanup pops execute even if `CancelledError` escapes `await instance.shutdown()`.

One new issue:

---

### Suggestion — `force_kill` docstring overclaims

**`pool.py`, `force_kill` docstring, lines ~210–215 (new text)**

The docstring says:

> Falls back to raw SIGKILL on any failure (timeout, cancellation, or other exception), so this **never blocks callers for long and never propagates exceptions**.

`CancelledError` is a `BaseException`, not an `Exception`. The `except Exception` block does not catch it. If the task is cancelled during `asyncio.wait_for(...)`, `CancelledError` propagates out of `force_kill` — both the SIGKILL fallback and the "never propagates" guarantee are violated (though the `finally` correctly handles pool cleanup). The inline comment correctly removed `CancelledError` from its example list, but the docstring still overstates the guarantee.

Suggested fix — qualify the claim:

```python
# Falls back to raw SIGKILL on any non-cancellation failure,
# so this never propagates ordinary exceptions to callers.
```

---

### Clean otherwise

The `finally`-based cleanup in both `force_kill` and `_eviction_loop`, the SIGKILL fallback placement, the dead-instance pop (now outside the `is_alive` guard), and the test coverage (especially `check_pool_during_shutdown` verifying pool membership during `await`) are all correct. The double-shutdown window (suggestion, not taken) is the author's call.

dcellison:
## Review by Kai

## Review: fix/eviction-silent-failure (round 4)

All three prior warning-level issues have been addressed. The `force_kill` docstring now correctly qualifies the guarantee ("non-cancellation failure"), and the outer `try/finally` in `_eviction_loop` ensures pool cleanup even on `CancelledError`. One minor test inconsistency:

---

### Suggestion — `test_eviction_pops_after_shutdown` missing `_last_activity` assertion

**`tests/test_pool.py`, line ~580**

```python
        # Removed after shutdown completed
        assert 111 not in pool._pool
        # _last_activity not checked here
```

Every other eviction test in this PR checks both `pool._pool` and `pool._last_activity` after cleanup (e.g. `test_eviction_failure_triggers_sigkill_fallback` asserts both). This test only asserts `_pool`. Inconsistent and leaves a gap — if `_last_activity` cleanup regressed in the success path, this test wouldn't catch it.

---

### Clean otherwise

The `finally`-based cleanup in both `force_kill` and `_eviction_loop`, the corrected docstring, the SIGKILL fallback placement, and the `check_pool_during_shutdown`/`check_pool` pattern for verifying pool membership during `await` are all solid. The double-shutdown window (suggestion, not taken) remains the author's call.

PR #184, opened 2026-03-26
Add GITHUB_NOTIFY_CHAT_ID to make config prompts

### LINKED ISSUE #183: Add GITHUB_NOTIFY_CHAT_ID to make config prompts
## Summary

`GITHUB_NOTIFY_CHAT_ID` (added in PR #182) is configured by manually editing `/etc/kai/env`. It should be a prompted field in `make config` (`_cmd_config()` in `install.py`) like all other user-facing env vars.

## Current state

The env var is parsed at startup in `start()` but is not part of the interactive config prompts. Users have to know to add it manually. This is inconsistent with every other configurable value (bot token, webhook secret, model, timeout, workspace base, etc.), which are all prompted during `make config`.

## Expected behavior

`make config` should prompt for `GITHUB_NOTIFY_CHAT_ID` as an optional field. When left empty, the feature is disabled (same as not setting it). When a value is provided, it should be validated as an integer before writing to `install.conf`.

The prompt should appear in a logical location - after the GitHub/webhook section or as part of a notifications section. It should use the existing value from `install.conf` as the default when re-running config.

## Implementation notes

- Add the prompt to `_cmd_config()` in `install.py`
- Validate the value is an integer (or empty)
- Store in the `env` dict in `install.conf` alongside other env vars
- `_cmd_apply()` already writes the `env` dict to `/etc/kai/env`, so no changes needed there
- The startup parsing in `start()` reads from `os.environ` which gets populated from `/etc/kai/env`, so no changes needed there either

dcellison:
## Summary

Add `GITHUB_NOTIFY_CHAT_ID` to the interactive `make config` prompts so it persists in `install.conf` and survives `make apply`.

Without this, the env var had to be manually added to `/etc/kai/env`, where it would be silently dropped on the next `make apply` (which regenerates the env file from `install.conf`).

## Changes

| Change | Detail |
|--------|--------|
| `_validate_chat_id()` | New validator accepting any non-zero integer (Telegram group IDs are negative) |
| `_cmd_config()` prompt | New "GitHub notifications" section between issue triage and optional features |
| Env dict entry | `GITHUB_NOTIFY_CHAT_ID` added to env dict when non-empty |
| Test update | `test_writes_install_conf` input list updated for new prompt |

## Test plan

- [x] `test_writes_install_conf` updated with new prompt input
- [x] `test_reads_existing_defaults` passes (empty default accepted)
- [x] All 1147 tests pass, `make check` clean
- [ ] Manual: run `make config`, enter a group chat ID, verify it appears in `install.conf`

Fixes #183

### DIFF
--- a/tests/test_install.py
+++ b/tests/test_install.py
@@ -37,6 +37,7 @@ from kai.install import (
     _start_service,
     _stop_service,
     _user_home,
+    _validate_chat_id,
     _validate_port,
     _validate_positive_float,
     _validate_positive_int,
@@ -350,6 +351,11 @@ class TestCmdConfig:
         assert _validate_user_ids("abc") is False
         assert _validate_port("0") is False
         assert _validate_port("99999") is False
+        # Chat ID accepts any non-zero integer (group IDs are negative)
+        assert _validate_chat_id("-1001234567890") is True
+        assert _validate_chat_id("12345") is True
+        assert _validate_chat_id("0") is False
+        assert _validate_chat_id("abc") is False
 
 
 # ── Apply subcommand ─────────────────────────────────────────────────


dcellison:
## Review by Kai

The PR is clean. The prior review's suggestion about `_validate_chat_id` test coverage has been addressed — the new assertions in `test_validates_required_fields` (lines 354-357) cover negative IDs, positive IDs, zero, and non-numeric strings.

No new issues to raise:

- **Validator** (`install.py:176`): Correct. `int()` handles the full valid range; empty-string case is handled at the call site, not inside the validator.
- **Prompt loop** (`install.py:382`): Pattern is consistent with the rest of `_cmd_config`. Empty accepted, non-zero integer accepted, anything else re-prompts.
- **Env dict** (`install.py:448`): Guarded by `if github_notify_chat_id` — the `"0"`-rejection in the validator means a zero ID can never reach this point, but the guard is still correct on its own terms.
- **Test input list** (`tests/test_install.py:294`): Position of the new `""` entry matches the new prompt's position in the function.

Ready to merge.

PR #185, opened 2026-03-26
Send save prompt to inner Claude before session shutdown

### LINKED ISSUE #148: Send save prompt to inner Claude before session shutdown
## Summary

When the inner Claude subprocess is about to be killed (idle eviction, `/stop`, restart), it has no opportunity to save what it learned during the session. Interactive Claude Code sessions do this automatically via auto-memory, but the inner Claude runs in stream-json mode where auto-memory does not function.

A "save prompt" sent before shutdown would give the inner Claude a chance to persist important context to MEMORY.md, replicating the auto-memory save behavior for the headless subprocess.

## Proposed approach

Before killing the subprocess, send a final message like:

> "Save anything worth remembering from this session to your memory file, then respond with DONE."

Wait for the response (with a short timeout), then proceed with the kill.

## Considerations

- Requires sending a message to a subprocess that's about to be killed
- Need to wait for the save to complete before killing (short timeout)
- Must handle the case where the subprocess is unresponsive (timeout and kill anyway)
- The `/stop` path needs the kill to be fast - the save prompt adds latency
- Idle eviction is the most natural place for this (the subprocess is about to die anyway, no user waiting)
- Could be limited to idle eviction only, skipping `/stop` (which implies urgency)

## Depends on

- #147 (inner Claude needs the correct MEMORY.md path first)

## Priority

Low. This is an enhancement for long-term memory building. The explicit memory system (CLAUDE.md + MEMORY.md + history) handles the core use case.

dcellison:
## Summary

Add a save prompt that gives the inner Claude a chance to persist useful context to MEMORY.md before being killed. The inner Claude runs in stream-json mode where auto-memory does not fire, so this is the only mechanism for building persistent memory over time.

### Which kill paths get the save prompt

| Path | Save prompt? | Why |
|------|-------------|-----|
| Idle eviction | Yes | Process is alive, healthy, nobody waiting |
| Graceful shutdown | Yes | Clean exit, process is alive |
| Session age recycle | Yes | Healthy process, just old |
| `/stop` | No | User wants it dead now |
| `/new`, model switch, workspace change | No | Context is being replaced |
| Timeouts, errors, EOF | No | Process is stuck or dead |

### How it works

1. `_save_prompt()` sends a stream-json user message asking the inner Claude to save useful context
2. Reads and discards response lines until a `{"type": "result"}` event (completion signal)
3. 30-second timeout covers the entire save operation (read memory, decide, write, respond)
4. Best-effort: any failure (timeout, EOF, pipe error, exception) is caught and shutdown proceeds

### Lock safety

All three callers guarantee exclusive pipe access:
- Idle eviction: `_in_flight` guard prevents concurrent `_send_locked()`
- Graceful shutdown: bot has stopped accepting messages
- Session recycle: inside `_send_locked()` which holds the lock

## Changes

| File | Change |
|------|--------|
| `src/kai/claude.py` | New `_save_prompt()` method; called from `shutdown()` and session recycle in `_send_locked()` |

## Test plan

- [x] `test_sends_save_message_to_stdin` - correct stream-json format
- [x] `test_skips_fresh_session` - no save on unused sessions
- [x] `test_skips_dead_process` - no save on dead process
- [x] `test_handles_stdin_write_failure` - broken pipe handled
- [x] `test_stops_on_result_event` - stops on completion signal
- [x] `test_stops_on_eof` - stops when process dies
- [x] `test_stops_on_timeout` - stops after timeout
- [x] `test_handles_json_parse_error` - non-JSON lines skipped
- [x] `test_handles_unexpected_exception` - unexpected errors caught
- [x] `test_shutdown_calls_save_prompt` - called before SIGTERM
- [x] `test_force_kill_does_not_call_save_prompt` - not called on force kill
- [x] All 1158 tests pass, `make check` clean

Fixes #148

### DIFF
--- a/tests/test_claude.py
+++ b/tests/test_claude.py
@@ -1669,6 +1669,22 @@ class TestSavePrompt:
         await claude._save_prompt()
 
     @pytest.mark.asyncio
+    async def test_handles_drain_runtime_error(self):
+        """_save_prompt handles RuntimeError from drain() gracefully."""
+        claude = _make_claude()
+        mock_proc = MagicMock()
+        mock_proc.returncode = None
+        mock_proc.stdin = MagicMock()
+        mock_proc.stdin.write = MagicMock()
+        mock_proc.stdin.drain = AsyncMock(side_effect=RuntimeError("transport closed"))
+        mock_proc.stdout = AsyncMock()
+        claude._proc = mock_proc
+        claude._fresh_session = False
+
+        # Should not raise
+        await claude._save_prompt()
+
+    @pytest.mark.asyncio
     async def test_stops_on_result_event(self):
         """_save_prompt stops reading when it sees a result event."""
         claude = _make_claude()
@@ -1865,7 +1881,7 @@ class TestSavePrompt:
             try:
                 async for _ in claude._send_locked("test", chat_id=123):
                     pass
-            except (AssertionError, StopAsyncIteration):
+            except (StopAsyncIteration, AttributeError, TypeError):
                 pass  # Expected - no real process after kill
 
         assert call_order[0] == "save_prompt_timeout=10"


dcellison (commented):
Posted in error.

PR #192, opened 2026-03-26
Skip sudo wrapper when os_user matches bot process user

### LINKED ISSUE #191: Skip sudo wrapper when os_user matches the bot process user
## Problem

When `os_user` in `users.yaml` (or `CLAUDE_USER` in env) is set to the same OS user the bot process runs as, the inner Claude subprocess fails immediately:

```
Sorry, user kai is not allowed to execute '/Users/kai/.local/bin/claude --version' as kai on hostname.
```

The sudoers rules generated by `make install` only allow running `claude` as a *different* user. Self-sudo serves no isolation purpose and is correctly disallowed by sudoers - but the code accepts the value without complaint, resulting in a silent startup failure.

## Expected behavior

If `os_user` resolves to the same user the bot is already running as, the code should skip the `sudo -u` wrapper entirely and log a warning explaining why the value was ignored. The subprocess should start normally, as if `os_user` were unset.

## Current behavior

The code unconditionally wraps the command with `sudo -u <os_user> --` whenever `os_user` is non-None (`claude.py:210-216`). When the target user equals the current user, `sudo` checks the sudoers policy, fails, and the process exits immediately. The log shows only "Claude process EOF" with no indication of the root cause.

## Scope

- `src/kai/claude.py` - compare `self.claude_user` against `os.getuid()` at the sudo wrapping point; skip sudo and log a warning if they match
- No changes to config loading or validation - the value is still accepted, just not acted on

## References

- Documented workaround: `home/docs/claude-user-gotcha.md`

dcellison:
## Summary

When `os_user` (from `users.yaml`) or `CLAUDE_USER` matches the user the bot process is already running as, skip the `sudo -u` wrapper instead of crashing with "Claude process EOF."

### The problem

`_ensure_started()` unconditionally wraps the Claude command with `sudo -u <user>` when `claude_user` is set. Self-sudo (`sudo -u kai` when already running as `kai`) fails because sudoers disallows it, and the subprocess dies immediately.

### The fix

Compare `self.claude_user` against `pwd.getpwuid(os.getuid()).pw_name` at spawn time. If they match, treat it as if `claude_user` were None (no sudo, no process group, no PGID). Log a warning so the misconfiguration is visible.

## Changes

| Change | Detail |
|--------|--------|
| `import pwd` | Added to claude.py |
| `effective_claude_user` local | Avoids mutating `self.claude_user` |
| Self-sudo detection | `pwd.getpwuid(os.getuid()).pw_name` comparison |
| Warning log | "os_user matches the bot process user; skipping sudo" |
| Downstream refs | `start_new_session`, `_pgid`, log message use `effective_claude_user` |

## Test plan

- [x] `test_self_sudo_skipped` - same user skips sudo, no process group, warning logged
- [x] `test_different_user_still_uses_sudo` - different user still gets sudo wrapper
- [x] Existing command construction tests pass
- [x] All 1175 tests pass, `make check` clean

Fixes #191

### DIFF
--- a/src/kai/claude.py
+++ b/src/kai/claude.py
@@ -224,8 +224,7 @@ class PersistentClaude:
                 current_user = None
             if effective_claude_user == current_user:
                 log.warning(
-                    "claude_user %r matches the bot process user; "
-                    "skipping sudo (no isolation)",
+                    "claude_user %r matches the bot process user; skipping sudo (no isolation)",
                     effective_claude_user,
                 )
                 effective_claude_user = None


dcellison:
## Review by Kai

## PR Review: Skip sudo wrapper when os_user matches bot process user

All findings from the three prior review rounds have been addressed in this diff:

- `KeyError` guard in `claude.py` — present and correct
- `import pwd` at module level in test file — done
- `test_different_user_still_uses_sudo` patches `kai.claude.pwd.getpwuid` — done
- `pytest.skip` guard in `test_self_sudo_skipped` — done
- Warning label changed from `"os_user"` to `"claude_user"` — done

No new findings. The implementation is clean:

- `effective_claude_user` correctly avoids mutating `self.claude_user`
- All three downstream references (`start_new_session`, `_pgid`, log line) use `effective_claude_user` consistently
- `test_pwd_keyerror_falls_through_to_sudo` directly exercises the `KeyError` guard
- The `is False` check in `test_self_sudo_skipped` (line ~207) is fine — `False` is a singleton in CPython and the kwarg is always explicitly passed

Ready to merge.

dcellison:
## Review by Kai

## PR Review: Skip sudo wrapper when os_user matches bot process user

All findings from the four prior review rounds have been addressed in this diff. A fresh pass finds nothing new.

**Implementation (`claude.py`):**
- `effective_claude_user` correctly avoids mutating `self.claude_user`, so repeated `_ensure_started()` calls (e.g. on process restart) re-evaluate the check rather than inheriting stale state
- `KeyError` guard is correct; `current_user = None` falls through to the sudo path as intended
- Warning label is `"claude_user"` — accurate for both `users.yaml` and `CLAUDE_USER` sources
- All three downstream references (`cmd`, `start_new_session`, `_pgid`) use `effective_claude_user` consistently

**Tests (`test_claude.py`):**
- `import pwd` is at module level
- `pytest.skip` guard in `test_self_sudo_skipped` protects against unmapped-UID containers at test time
- `test_different_user_still_uses_sudo` patches `kai.claude.pwd.getpwuid` — hermetic
- `test_pwd_keyerror_falls_through_to_sudo` directly exercises the `KeyError` guard
- `assert args[1].get("start_new_session") is False` is fine — `bool(None)` is the `False` singleton

No new findings. Ready to merge.

PR #207, opened 2026-03-28
Replace wall-clock limit with activity-based idle timeout

### LINKED ISSUE #206: Replace wall-clock limit with activity-based idle timeout
## Problem

The wall-clock limit in `claude.py` measures elapsed time from **interaction start**, not from last activity. This means legitimately long-running tasks (research, multi-step tool use, detailed planning) get killed after 10 minutes even though Claude is actively working and producing output.

The two recent failures (Mar 27):
```
ERROR Interaction exceeded wall-clock limit (616s > 600s)
ERROR Interaction exceeded wall-clock limit (765s > 600s)
```

Both were healthy interactions where Claude was actively doing tool-use work on a complex research task. The process wasn't stuck or dead - it just needed more than 10 minutes.

## Root Cause

`claude.py:584-606` - the wall clock measures from `interaction_start` and caps at `timeout_seconds * 5` (600s at default 120s). Every iteration of the streaming loop checks total elapsed time. This conflates two different failure modes:

1. **Dead/stuck process** (should be killed) - no output flowing
2. **Healthy long-running task** (should NOT be killed) - events flowing continuously

## Proposed Fix

Change the wall-clock timer from "time since interaction start" to **"time since last event received."** Reset the timer every time the process emits a line of output.

### Current behavior

| Scenario | Readline timeout (6 min) | Wall clock (10 min from start) |
|---|---|---|
| Dead process (no output) | Fires at 6 min | Fires at 10 min (redundant) |
| Stuck tool loop (events flowing) | Never fires | Fires at 10 min |
| Legitimate long task (events flowing) | Never fires | **Fires at 10 min (WRONG)** |

### Target behavior

| Scenario | Readline timeout (6 min) | Activity wall clock (10 min since last event) |
|---|---|---|
| Dead process (no output) | Fires at 6 min | Fires at 10 min (redundant backup) |
| Stuck tool loop (events flowing) | Never fires | Never fires (events reset it) |
| Legitimate long task (events flowing) | Never fires | Never fires (events reset it) |

Note: the "stuck tool loop" row changes from "caught" to "not caught." This is acceptable because (a) true infinite tool loops are extremely rare in practice, (b) the user can always `/stop`, and (c) killing legitimate work is a worse failure mode than letting a rare loop run long.

### 1. `src/kai/claude.py` - streaming loop (~line 577-606)

Replace the fixed `interaction_start` timer with a resetting activity timer:

```python
# Before (line 584-585):
interaction_start = time.monotonic()
max_interaction_seconds = self.timeout_seconds * 5

# After:
last_activity = time.monotonic()
max_idle_seconds = self.timeout_seconds * 5  # 10 min of silence at default 120s
```

In the wall-clock check (~line 588-606):

```python
# Before:
elapsed = time.monotonic() - interaction_start
if elapsed > max_interaction_seconds:

# After:
idle_seconds = time.monotonic() - last_activity
if idle_seconds > max_idle_seconds:
```

Update the error log message to reflect the new semantics:

```python
log.error(
    "Interaction idle timeout (%.0fs with no output, limit %ds)",
    idle_seconds,
    max_idle_seconds,
)
```

Update the error string for the user-facing message:

```python
error="Claude interaction timed out (no output)"
```

Reset the timer after every successful readline (~after line 611, before the EOF/JSON checks):

```python
line = await asyncio.wait_for(self._proc.stdout.readline(), timeout=timeout)
last_activity = time.monotonic()  # Reset idle timer on any output
```

Place the reset **immediately after** the `wait_for` succeeds and **before** the `if not line:` EOF check. An empty `line` (EOF) should NOT reset the timer.

```python
try:
    timeout = self.timeout_seconds * 3
    line = await asyncio.wait_for(self._proc.stdout.readline(), timeout=timeout)
except TimeoutError:
    # ... existing dead-process handling ...
    return

# Reset idle timer - process is alive and producing output
if line:
    last_activity = time.monotonic()

if not line:
    # Process died unexpectedly (EOF)
    # ... existing EOF handling ...
```

### 2. Update comments

The block comment at lines 578-583 needs to be rewritten to reflect the new semantics:

```python
# Idle-activity timeout for the interaction. If the process goes
# silent for too long (no stdout lines at all), it is likely dead
# or wedged. The per-readline timeout below is the primary detector
# (no single line for timeout_seconds * 3), while this is a
# secondary safety net measured across the entire interaction.
```

### 3. `src/kai/bot.py` - lock acquire timeout (~line 160)

The `_LOCK_ACQUIRE_TIMEOUT` is currently 660s (11 min), set just above the old 10-min wall clock. With the activity-based timer, interactions can legitimately run much longer. This timeout needs to increase or be removed.

Options (pick one):
- **Increase to 3600s (1 hour)** - generous but still bounded
- **Remove the timeout entirely** - rely on `/stop` and the idle timer in claude.py

Recommendation: increase to 3600 and update the comment explaining why. The idle timer in claude.py is the real safety net now; the lock timeout is just a last-resort backstop.

Also update the comment block at lines 152-159 to reference the new idle-based semantics instead of the old wall-clock semantics.

### 4. No new env vars needed

The existing `CLAUDE_TIMEOUT_SECONDS` (default 120) still drives both derived timeouts:
- Per-readline: `timeout_seconds * 3` = 360s (6 min)
- Idle wall clock: `timeout_seconds * 5` = 600s (10 min)

A separate `CLAUDE_MAX_INTERACTION_SECONDS` env var could be added later if the multiplier needs independent tuning, but it is not necessary for this change.

## Testing

1. Verify dead process detection still works - kill the Claude subprocess externally during an interaction; should detect EOF immediately
2. Verify readline timeout still works - simulate a hung process; should timeout after 6 min of no output
3. Verify long-running tasks survive - send a complex multi-tool-use prompt that takes >10 min; should complete successfully
4. Verify `/stop` still works during long interactions
5. Check that `_LOCK_ACQUIRE_TIMEOUT` doesn't interfere with long tasks

dcellison:
## Summary

The wall-clock timer in `claude.py` killed interactions after `timeout_seconds * 5` (10 min at defaults) of **total elapsed time**, regardless of whether Claude was actively working. Two failures on Mar 27 - both healthy interactions doing multi-step research with continuous tool-use output:

```
ERROR Interaction exceeded wall-clock limit (616s > 600s)
ERROR Interaction exceeded wall-clock limit (765s > 600s)
```

This replaces the fixed wall-clock with an **activity-based idle timeout** that resets every time the process emits a line of output. Same duration, different measurement: "time since last output" instead of "time since start."

### What changed

- **claude.py**: `interaction_start` becomes `last_activity`, reset on every non-empty readline. The idle check fires only after prolonged silence, not after prolonged work.
- **bot.py**: `_LOCK_ACQUIRE_TIMEOUT` increased from 660s to 3600s. With no wall-clock cap, the lock timeout needs to accommodate legitimately long interactions. The idle timer in claude.py is the real safety net.
- **Error message**: "timed out (too long)" becomes "timed out (no output)" to reflect the new semantics.

### Timeout behavior after this change

| Scenario | Readline timeout (6 min) | Idle timeout (10 min since last output) |
|---|---|---|
| Dead process (no output) | Fires at 6 min | Fires at 10 min (redundant backup) |
| Stuck tool loop (events flowing) | Never fires | Never fires (events reset it) |
| Legitimate long task (events flowing) | Never fires | Never fires (events reset it) |

The "stuck tool loop" row changes from "caught" to "not caught." This is an acceptable tradeoff: true infinite tool loops are extremely rare, the user can always `/stop`, and killing legitimate long work is a worse failure mode.

### Implementation note

With the current multipliers (readline=3x, idle=5x), the idle timeout can never fire independently in practice - the readline timeout always fires first for a truly silent process (3x < 5x). The idle timer is a "belt and suspenders" backup that would only matter if the multipliers were changed or if there were an edge case where readline keeps barely succeeding. The `test_idle_timeout` test uses a mocked `time.monotonic` in the `kai.claude` namespace to force the idle check to fire, which is the correct approach for verifying an error path that is dominated by another timeout in real execution.

## Test plan

- [ ] `test_idle_timeout` - mocked time forces idle check to fire, verifies error path and "no output" message
- [ ] `test_idle_timeout_normal_completion_unaffected` - fast interaction completes without hitting idle timer
- [ ] `test_active_process_survives_past_old_wall_clock` - core behavioral test: active process emits output for ~8s (past old 5s limit), completes successfully
- [ ] `test_readline_timeout` (existing, unchanged) - fully dead process caught by per-readline timeout
- [ ] `test_timeout_kills_claude` (bot.py, comment-only update) - lock timeout still works
- [ ] Full suite: 1213 tests passing

Fixes #206

### DIFF
--- a/tests/test_claude.py
+++ b/tests/test_claude.py
@@ -925,7 +925,7 @@ class TestSendLockedErrors:
                 _result_event("Done"),
             ]
         )
-        claude = _make_claude(timeout_seconds=1)  # wall-clock = 5s
+        claude = _make_claude(timeout_seconds=1)  # idle limit = 5s
         claude._proc = proc
         claude._fresh_session = False
 


dcellison:
## Review by Kai

## PR Review: Replace wall-clock limit with activity-based idle timeout

### Clean

The `pytest.fail` fix requested in the previous review is properly implemented in `readline_with_output` (`tests/test_claude.py`, readline 3 branch). If `fake_monotonic` drifts, the test now fails loudly instead of silently passing with a valid event.

The `advancing_monotonic` approach in `test_active_process_survives_past_old_wall_clock` is sound. `sim_time[0]` advances only inside `active_readline`, so at each loop iteration the idle check sees `sim_time[0] - last_activity == 0.0` — the timer never fires spuriously. The `side_effect=advancing_monotonic` wiring is correct (`time.monotonic` takes no args, neither does the mock function).

`last_activity` reset placement in `claude.py:618` is correct: fires on `if line:`, mutually exclusive with the `if not line:` EOF branch, not reachable from the `asyncio.TimeoutError` path.

---

### Remaining open items (from prior reviews, no material change)

- **`mono_call[0] <= 5` threshold** — still tied to exact call count in `kai.claude`. The added comment documenting the call sequence mitigates the maintenance risk but doesn't eliminate it.
- **Active-but-stuck blast radius** — still 6× larger (10 min → 60 min). The acknowledged tradeoff.

Both were raised previously and the author has made their decisions. No new issues found.

dcellison:
## Review by Kai

## PR Review: Replace wall-clock limit with activity-based idle timeout

**Clean.** No new issues found.

---

### Verification of open items from prior reviews

**`mono_call[0] <= 5` fragility** — still present, now has a doc comment explaining the exact call sequence. No material change; author's accepted tradeoff.

**Active-but-stuck blast radius** — still 6× larger (10 min → 60 min). No material change; author's accepted tradeoff.

---

### New code reviewed

**`claude.py:618` — `last_activity` reset placement**

`if line:` / `if not line:` are mutually exclusive. Reset fires only on non-empty output, not on EOF, not on `asyncio.TimeoutError`. Correct.

**`test_idle_timeout` — `fake_monotonic` + `readline_with_output`**

Call sequence matches the documented pattern (1 init + 2 check/reset per active iteration = 5 before the jump). `pytest.fail` on readline 3 is in place. `side_effect=fake_monotonic` wiring is correct — mock passes no args to the callable, matching `def fake_monotonic():`.

**`test_active_process_survives_past_old_wall_clock`**

`advancing_monotonic` reads `sim_time[0]` which only advances *inside* `active_readline`, so the idle check always sees `sim_time[0] - last_activity == 0.0` at each iteration. The timer never fires spuriously. After call 21 returns `_result_event`, `sim_time[0]` stays at 10.0 (no `+= 0.5` for the terminal event), `last_activity` is reset to 10.0, and the loop exits normally. Correct.

**`bot.py` comment** — accurately describes the new semantics. No logic change beyond the constant.

PR #215, opened 2026-04-02
Fix context window: use --settings flag instead of invalid CLI flag

dcellison:
## Summary

- Fix inner Claude crash on startup caused by invalid `--max-context-window` CLI flag
- Use `--settings '{"preferences":{"maxContextWindow":N}}'` instead, which is the correct way to pass context window config to Claude Code

Introduced in #214; the flag was never valid - `maxContextWindow` is a settings.json option, not a CLI flag.

## Test plan

- [ ] `make test` passes
- [ ] Inner Claude starts successfully with `CLAUDE_MAX_CONTEXT_WINDOW=200000` configured
- [ ] Verify `--settings` appears in the subprocess command with correct JSON payload

### DIFF
--- a/src/kai/config.py
+++ b/src/kai/config.py
@@ -873,7 +873,7 @@ def load_config() -> Config:
         raise SystemExit("FILE_RETENTION_DAYS must be an integer") from None
 
     # Context window tuning - 0 means "use Claude Code defaults"
-    _MAX_CONTEXT_CEILING = 10_000_000  # 10M tokens; sanity check
+    _MAX_CONTEXT_CEILING = 1_000_000  # 1M tokens; Claude's maximum
     try:
         claude_max_context_window = int(os.environ.get("CLAUDE_MAX_CONTEXT_WINDOW", "0"))
         if claude_max_context_window < 0 or claude_max_context_window > _MAX_CONTEXT_CEILING:
--- a/src/kai/install.py
+++ b/src/kai/install.py
@@ -443,11 +443,11 @@ def _cmd_config() -> None:
         )
         try:
             val = int(max_context_window)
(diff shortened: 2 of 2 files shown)


dcellison:
## Review by Kai

## Review

**Overall: Clean fix, correct approach.** One item to verify before merging.

---

**WARNING — `import json` not shown in diff for `claude.py`**

`src/kai/claude.py`, context window block

`json.dumps(settings)` is now called in `claude.py`, but the diff adds no `import json` line to that file. If `json` was not already imported there (it wasn't needed before this PR — the old code used `str(self.max_context_window)` with no JSON), this will raise a `NameError` at runtime the moment a non-zero `max_context_window` is configured. The test file already had `json` imported, which is why tests pass.

Quick check: grep for `import json` at the top of `claude.py`. If it's absent, add it.

---

**SUGGESTION — Comment inaccuracy in `config.py`**

`src/kai/config.py` line ~875

```python
_MAX_CONTEXT_CEILING = 1_000_000  # 1M tokens; Claude's maximum
```

"Claude's maximum" isn't accurate — current Claude models cap at 200K tokens. 1M is still a reasonable sanity ceiling (much better than the prior 10M), but the comment overstates it. `# 1M tokens; generous sanity cap` or similar would be more precise. Purely cosmetic.

---

**No issues with:**
- The `--settings` JSON payload — `json.dumps` on a known dict with a validated int is injection-safe.
- The ceiling tightening in both `config.py` and `install.py` — these are consistent with each other and directly address the prior review's upper-bound suggestion.
- The test updates — correct depth of assertion (`settings["preferences"]["maxContextWindow"]`), correct parse, correct zero-value guard.

PR #216, opened 2026-04-03
Manage per-workspace config via Telegram commands

dcellison:
## Summary

- Add `/workspace config` commands for viewing and modifying per-workspace settings via Telegram
- Settings persist in the database and layer on top of YAML baseline config (DB overrides > YAML > global defaults)
- Supported fields: model, budget, timeout, env vars, system prompt
- Budget ceiling enforcement prevents users from exceeding their admin-set limit
- All changes kill the Claude process so the next message starts with the new config

### Implementation (3 commits)

1. **DB helpers** in `sessions.py` - CRUD functions for namespaced `ws_config:{chat_id}:{workspace}:{field}` keys
2. **`build_workspace_config()` merge function** - layers DB overrides on YAML baseline, used by pool restore and workspace switching
3. **Command handlers** in `bot.py` - dispatch, show, set/reset, env var CRUD, prompt management

### Commands

- `/workspace config` - show effective config with source attribution
- `/workspace config model <haiku|sonnet|opus>` - set model
- `/workspace config budget <amount>` - set budget (with ceiling enforcement)
- `/workspace config timeout <seconds>` - set timeout
- `/workspace config env FOO=bar` / `env -FOO` / `env` - set/remove/list env vars
- `/workspace config prompt <text>` / `prompt clear` - set/clear system prompt
- `/workspace config reset [field]` - reset all or single field to defaults

## Test plan

- [x] 17 new unit tests for DB helpers in `test_sessions.py` (9 CRUD + 8 merge function)
- [x] 15 new integration tests for command handlers in `test_bot.py`
- [x] 11 existing tests updated to mock `build_workspace_config` at new call sites
- [x] Full suite passes (1258 tests)

### DIFF
--- a/src/kai/bot.py
+++ b/src/kai/bot.py
@@ -1082,7 +1082,7 @@ async def _show_workspace_config(
             env_corrupted = True
     if env_corrupted:
         lines.append("  Env vars: (DB override corrupted - reset to clear)")
-    if env_keys:
+    elif env_keys:
         lines.append(f"  Env vars: {', '.join(sorted(env_keys))}")
 
     # System prompt


dcellison:
## Review by Kai

## PR Review: Manage per-workspace config via Telegram commands

Overall the implementation is well-structured and the test coverage is thorough. A few real issues to address before merging.

---

### Warning

**1. SQLite LIKE wildcard injection in `sessions.py`**

`get_workspace_config_settings` and `delete_all_workspace_config` build a LIKE pattern using the raw workspace path:

```python
prefix = f"ws_config:{chat_id}:{workspace_path}:"
... WHERE key LIKE ?", (prefix + "%",)
```

SQLite treats `_` (match any one char) and `%` (match any sequence) as wildcards throughout the entire pattern, not just after a literal prefix. A path like `/opt/my_project` becomes `...:/opt/my_project:%`, where `_` matches any character. This means settings for `/opt/myXproject` would be returned or, worse, deleted.

The delete case is the dangerous one - `delete_all_workspace_config` could silently wipe settings for a sibling workspace. Fix by using `GLOB` (which uses `*`/`?` and doesn't treat `_` specially) or by adding a prefix-equality check instead:

```sql
DELETE FROM settings WHERE SUBSTR(key, 1, ?) = ?
```

---

**2. Routing collision for workspaces named "config*" (`bot.py:1395`)**

```python
if target.lower().startswith("config"):
    await _handle_workspace_config(update, context, target)
    return
```

Any workspace whose name begins with "config" - e.g., `configs`, `config-backup`, `configure` - will be silently misrouted to the config handler, which will reply "Unknown config field: configs" instead of switching. Use an exact prefix match:

```python
if target.lower() == "config" or target.lower().startswith("config "):
```

---

**3. Unhandled `json.loads` failure on corrupted "env" DB entry**

Three separate call sites parse the stored env JSON without a try/except:

- `sessions.py` `build_workspace_config`: `json.loads(db_settings["env"])`
- `bot.py` `_handle_workspace_env`: `env = json.loads(settings["env"])`
- `bot.py` `_show_workspace_config`: `db_env = json.loads(db_settings["env"])`

A corrupted row (manual DB edit, migration bug, truncation) would surface as an unhandled `JSONDecodeError` to the user. Wrap each in a `try/except json.JSONDecodeError` and either treat it as empty or log and surface a clear error.

---

### Suggestion

**4. No size limit on file-based prompt uploads (`bot.py` `_handle_workspace_prompt`)**

```python
content = (await file.download_as_bytearray()).decode("utf-8")
```

Telegram allows documents up to 20 MB. This downloads the entire file into memory and stores it in SQLite. Add a size check on `file.file_size` before downloading.

---

**5. `/workspace config reset <field>` always replies success**

```python
await sessions.delete_workspace_config_setting(chat_id, workspace_str, value)
await update.message.reply_text(f"Cleared {value} override. Using global default.")
```

`delete_setting` is a no-op when the key doesn't exist, so `/workspace config reset model` (when model was never overridden) confirms "Cleared model override" falsely. Worth checking whether anything was actually deleted, or rephrasing to "model reset to default" to avoid the implication that an override existed.

---

**6. `_patches` test helper has dead parameters**

`TestHandleWorkspaceConfig._patches(self, mock_sessions, pool=None, config=None)` - `pool` and `config` are accepted but never used. Minor noise.

PR #219, opened 2026-04-03
Per-user allowed workspaces

### LINKED ISSUE #202: Per-user allowed workspaces via Telegram commands
## Problem

`ALLOWED_WORKSPACES` is a global env var (comma-separated paths) that applies to all users. `WORKSPACE_BASE` is also global. In multi-user, each user has repos in different locations and needs their own workspace base and allowed paths.

Currently, if a user has repos outside `WORKSPACE_BASE`, they must ask the machine owner to add paths to `ALLOWED_WORKSPACES` in the env file. This breaks the principle that workspace management is fully Telegram-driven.

Parent issue: #196 (sub-issue #4)

## Proposal

Two changes:

### 1. Move `WORKSPACE_BASE` to per-user `workspace_base` in users.yaml

Each user gets their own base directory for `/workspace new` and short-name resolution:

```yaml
users:
  - telegram_id: 2114582497
    name: Daniel
    workspace_base: /Users/kai/Projects
```

The admin sets this during onboarding (the config wizard in #193 prompts for it). `/workspace new <name>` creates directories under this user's `workspace_base`.

### 2. User-managed allowed workspaces via Telegram

Users add and remove paths outside their `workspace_base` via Telegram commands. The filesystem (os_user permissions) is the security boundary, not an allowlist.

### Commands

```
/workspace allow <path>         - add an allowed workspace path
/workspace deny <path>          - remove an allowed workspace path
/workspace allowed              - list all allowed workspaces (workspace_base + explicit allows)
```

### Validation

- `/workspace allow` checks that the path exists and is a directory
- Path must be readable by the user's `os_user` (if os_user isolation is active)
- Absolute paths required (the current workspace command rejects them for name-based switching, but `/workspace allow` specifically needs them)
- Duplicate detection (don't add a path that's already under `workspace_base` or already allowed)

### Storage

Allowed workspaces stored in the database per-user. Options:

1. Settings table with a list key: `allowed_workspaces:{chat_id}` = JSON array of paths
2. New `allowed_workspaces` table: (chat_id, path) pairs

Option 2 is cleaner for add/remove operations and avoids JSON parsing.

### Impact on `/workspace` name resolution

Currently `/workspace <name>` searches `WORKSPACE_BASE` then `ALLOWED_WORKSPACES`. The new flow:

1. Search the user's `workspace_base` (from users.yaml)
2. Search the user's allowed workspaces (from database)
3. Match by directory name in both cases

### Backward compatibility

- `WORKSPACE_BASE` env var becomes the global fallback for users who don't have `workspace_base` in users.yaml
- `ALLOWED_WORKSPACES` env var becomes the global fallback for users who haven't added any via Telegram
- Existing single-user installs work unchanged

### Deprecation

After migration period:
- `WORKSPACE_BASE` removed from env, replaced by per-user `workspace_base` in users.yaml
- `ALLOWED_WORKSPACES` removed from env, replaced by database storage

### Context

- Current workspace switching: bot.py lines 1020-1117 (`handle_workspace`)
- Workspace name resolution: searches `WORKSPACE_BASE` then `ALLOWED_WORKSPACES`
- Config: `WORKSPACE_BASE` and `ALLOWED_WORKSPACES` loaded in config.py
- UserConfig dataclass: config.py lines 86-111 (needs `workspace_base` field added)

dcellison:
## Summary

Per-user allowed workspaces (fixes #202). Users can manage their own workspace access lists via Telegram commands, with admin-controlled workspace_base in users.yaml.

### Changes

- **Storage layer** (sessions.py + config.py): new `allowed_workspaces` DB table, `workspace_base` field in UserConfig, `resolve_workspace_access()` that builds per-user effective access (DB + global union)
- **Bot + pool integration** (bot.py + pool.py): `_is_workspace_allowed` signature changed from `(path, config)` to `(path, base, allowed)` - callers resolve per-user values first. Three new subcommands: `/workspace allow`, `/workspace deny`, `/workspace allowed`
- **Help text and config**: updated /help output and users.example.yaml with workspace_base

### Design

- Effective allowed list = DB entries (user-added) + global ALLOWED_WORKSPACES (env var), deduplicated
- workspace_base is admin-only (users.yaml), allowed_workspaces is user-managed (DB) + admin-managed (global)
- `/workspace allow` requires absolute `/` paths (no `~` - expanduser would resolve to wrong home in multi-user)
- `/workspace allowed` shows source attribution ("you" vs "global") for each entry

## Test plan

- [ ] 38 new tests across sessions, user config, bot, and pool
- [ ] 1348 total tests pass
- [ ] Verify /workspace allow adds a path and /workspace allowed shows it
- [ ] Verify /workspace deny removes user-added paths but not global ones
- [ ] Verify workspace restoration uses per-user access on subprocess restart

### DIFF
--- a/src/kai/bot.py
+++ b/src/kai/bot.py
@@ -1735,8 +1735,14 @@ async def _handle_workspace_deny(
         return
 
     raw_path = parts[1].strip()
-    # No expanduser() - same reasoning as _handle_workspace_allow:
-    # ~ resolves to the bot process's $HOME, not the requesting user's.
+
+    # Same absolute-path requirement as _handle_workspace_allow:
+    # relative paths resolve against cwd and will never match a stored
+    # entry, producing a confusing "not in your list" response.
+    if not raw_path.startswith("/"):
+        await update.message.reply_text("Path must be absolute (start with /).")
+        return
+
     resolved = Path(raw_path).resolve()
 
     # Check if this is a user-added path (in the database)
--- a/tests/test_bot.py
+++ b/tests/test_bot.py
@@ -1672,6 +1672,15 @@ class TestHandleWorkspaceDeny:
         assert "Usage" in reply
 
     @pytest.mark.asyncio
+    async def test_deny_relative_path_rejected(self):
+        """Relative paths are rejected with a clear message."""
+        update = _make_update()
+        ctx = _make_context(args=["deny", "relative/path"])
+        await _handle_workspace_deny(update, ctx, "deny relative/path")
+        reply = update.message.reply_text.call_args[0][0]
+        assert "absolute" in reply.lower()
+
+    @pytest.mark.asyncio
     async def test_deny_global_path(self, tmp_path):
         """Trying to deny a global ALLOWED_WORKSPACES path shows an explanation."""
         ws = tmp_path / "global-ws"


dcellison:
## Review by Kai

The prior review caught the main issues and the `_NO_BASE_MSG` fix landed correctly. Two small new findings:

---

### Suggestion

**`_handle_workspace_allow` re-resolves pre-resolved paths** (`bot.py`, `_handle_workspace_allow`)

```python
if resolved in [a.resolve() for a in allowed]:
```

`allowed` comes from `resolve_workspace_access()`, which already returns resolved paths. The comment in `_is_workspace_allowed` explicitly notes "no need to call `.resolve()` again on each entry" for the same reason, but `_handle_workspace_allow` does it anyway. Harmless, but inconsistent with the documented contract. Should be:

```python
if resolved in allowed:
```

---

### Suggestion

**"No allowed workspaces configured" shows when `base` is set** (`bot.py`, `_handle_workspace_allowed`)

If `base` is configured but `allowed` is empty, the output is:

```
Workspace base: /projects
No allowed workspaces configured.
```

The second line is technically accurate but slightly misleading since the user *does* have access - to everything under `base`. The permissive-mode sentence correctly gates on `not base and not allowed`, but the "no allowed workspaces" line has no such guard. Consider suppressing it when `base` is set, or replacing it with something like "No additional paths beyond workspace base."

---

### No issues found

Everything else is clean: the `_handle_workspace_deny` tilde behavior was raised and the author kept it as-is (non-destructive - it just produces a non-match). Auth coverage via `@_require_auth` is correct for all three new subcommands. The `allow`/`deny`/`allowed` name routing happens before the absolute-path guard, which is correct since `allow` requires `/`. DB schema, path resolution consistency, and test coverage all look solid.

dcellison:
## Review by Kai

## Review: PR #219 (revision 4)

The prior suggestion about the double `resolve_workspace_access` call in `handle_workspace` was not addressed - but that was raised previously and is out of scope here.

Looking at the new diff specifically:

---

### Suggestion

**`_handle_workspace_allow` displays unresolved `base` in the redundancy message** (`bot.py`, `_handle_workspace_allow`)

```python
await update.message.reply_text(
    f"Already covered by your workspace base:\n{base}\n\n"
    ...
)
```

`base` here is the value returned by `resolve_workspace_access`, which for the global-config path (`config.workspace_base`) may not be pre-resolved - only the per-user path from `_load_user_configs` is guaranteed to have been through `.expanduser().resolve()`. In practice the global config is probably already canonical, but `resolved_base` (one line above) is the consistent value to show. Minor.

---

### No other issues found

- The new absolute-path guard in `_handle_workspace_deny` correctly mirrors `_handle_workspace_allow` (the tilde-expansion bug from prior review is fixed).
- `_handle_workspace_allowed` attribution logic (`db_path_set` membership check using pre-resolved paths) is correct.
- `add_allowed_workspace` / `remove_allowed_workspace` CRUD and `INSERT OR IGNORE` idempotency are correct.
- `resolve_workspace_access` union/dedup logic is correct: DB first, global second, set-based dedup.
- Auth coverage: all three new subcommands reachable only through `@_require_auth`-decorated `handle_workspace`.
- Routing order: `allow`/`deny`/`allowed` dispatch before the absolute-path rejection guard - correct since `allow` and `deny` do their own path validation.
- DB schema: `(chat_id, path)` PK is right for this table.
- `_workspaces_keyboard` and `handle_workspace_callback` correctly use the per-user `allowed` list from `resolve_workspace_access`.
- Test coverage is thorough and the `_mock_resolve` helper cleanly isolates the DB calls.

Overall this is clean, ready to merge.

Issue #235, opened 2026-04-04
Code review findings: #196 multi-user configuration


dcellison:
## Summary

A thorough code review of all PRs merged under #196 (multi-user configuration) found 8 issues ranging from a high-severity notification delivery bug to minor hardening improvements.

Full review document: `home/specs/196-code-review.md`

## Findings

| # | Finding | Severity | Sub-issue |
|---|---------|----------|-----------|
| 1 | `/github notify` does not update `allowed_user_ids` at runtime | High | #236 |
| 2 | Unguarded `int()` in `resolve_github_settings()` | Medium | #237 |
| 3 | Review/triage `claude_user` ignores per-user `os_user` | Medium | #238 |
| 4 | Empty string model passes through resolution | Low | #239 |
| 5 | `_CONFIG_ENV_VARS` test list incomplete | Medium (tests) | #239 |
| 6 | Malformed `users.yaml` error handling asymmetry | Design | #239 |
| 7 | Deprecation warnings skip on malformed config | Low | #239 |
| 8 | Case-sensitive boolean comparison | Low | #239 |

## Architecture observation

The bugs cluster at one seam: where the per-user model meets older single-user code that was not fully updated. Findings #1 and #3 are both cases where background task dispatch still uses global config instead of resolving per-user. The interactive session path (pool.py) handles per-user resolution correctly; the background task path (webhook.py dispatching to review.py/triage.py) was not updated to match.

Issue #196, opened 2026-03-27
Multi-user configuration: eliminate single-user env var defaults


dcellison:
## Overview

Kai was built as a single-user system. One person, one machine, one `.env` file with every setting in it. That worked fine when there was one user and one admin and they were the same person.

Now Kai has multiple users. The project is public, external contributors have joined, and people are running their own instances. The single-user assumptions are baked into every layer of configuration, and they create real problems:

**Users can't manage their own experience.** Model preference, budget ceiling, timeout, workspace access, GitHub notification routing - all of these are per-user concerns, but they live in `/etc/kai/env` or root-owned YAML files. A user who wants to switch from Sonnet to Opus, add a workspace, or change where their GitHub notifications go has to ask the machine owner to edit a config file. That's not a minor inconvenience; it's a fundamental UX failure for a bot that's supposed to be self-service via Telegram.

**The admin has no boundaries to set.** In single-user, there's no distinction between "what the machine owner controls" and "what the user controls." In multi-user, the admin needs to set ceilings (max budget, allowed OS user, role) while letting users freely configure everything else. Today there's no mechanism for this - every setting is either fully locked in env or fully open.

**GitHub integration doesn't know who cares about what.** PR reviews, issue triage, and notifications all use global env vars. When a webhook event arrives, the system has no concept of which user is subscribed to which repo. All notifications go to a single hardcoded chat ID (`GITHUB_NOTIFY_CHAT_ID`). With multiple users working on different repos, this is fundamentally broken.

**Workspace access is globally controlled.** `WORKSPACE_BASE` and `ALLOWED_WORKSPACES` are single env vars shared by everyone. One user's scattered repos can't be allowed without also allowing them for all other users. And users can't add their own workspaces without admin intervention.

**Configuration is scattered across three sources with no clear ownership.** Settings live in `.env`, `users.yaml`, and `workspaces.yaml` with overlapping concerns and no consistent precedence model. Some settings exist in multiple places (e.g., `CLAUDE_MODEL` in env vs `model` in workspaces.yaml) with ad-hoc override behavior.

### The goal

Every setting falls cleanly into one of three categories:

1. **Truly global** (`.env`) - infrastructure, server capabilities, resource limits. Things that describe the machine itself: bot token, webhook port, whether whisper-cpp is installed, how long sessions can run. The machine owner sets these once. Users never see or think about them.

[...]

PR #243, opened 2026-04-05
Harden edge cases in resolve_user_defaults and resolve_github_settings

### LINKED ISSUE #239: Hardening: minor robustness fixes from #196 code review
## Summary

A collection of low-severity findings from the #196 code review that can be addressed in a single PR. None of these are urgent, but each improves robustness or test reliability.

Parent issue: #235

### 1. Empty string model passes through `resolve_user_defaults()`

**File:** sessions.py:707-709

The `db_model` check uses `is not None` to avoid treating empty string as falsy, but this means an empty string `""` is treated as a valid model name. Not reachable via the UI (bot.py validates before storing) but could propagate from direct DB manipulation.

**Fix:** Change to `if db_model is not None and db_model.strip():` or `if db_model` (since empty string should fall through).

### 2. `_CONFIG_ENV_VARS` test list is incomplete

**File:** tests/test_config.py:15-41

The `_clean_env` fixture clears listed env vars to prevent test pollution. Missing: `CLAUDE_IDLE_TIMEOUT`, `FILE_RETENTION_DAYS`, `TOTP_SESSION_MINUTES`, `TOTP_CHALLENGE_SECONDS`, `TOTP_LOCKOUT_ATTEMPTS`, `TOTP_LOCKOUT_MINUTES`, `PERPLEXITY_API_KEY`.

**Fix:** Add the missing vars to the list.

### 3. Malformed `users.yaml` error handling asymmetry

**File:** config.py:626-628 vs config.py:456-458

Malformed protected `users.yaml` returns `None` (blocks all auth). Malformed `workspaces.yaml` returns `{}` (continues with empty config). The asymmetry is arguably intentional (fail closed on auth) but undocumented.

**Fix:** Add a comment explaining the intentional difference, or align the behavior.

### 4. Deprecation warnings skip when protected config is malformed

**File:** config.py:1133

The deprecation warning block only runs when `user_configs is not None`. If `/etc/kai/users.yaml` is malformed, no warnings are emitted even if deprecated env vars are set.

**Fix:** Consider emitting deprecation warnings when any users.yaml exists (check file existence, not parse result), or accept this as a non-issue.

### 5. Case-sensitive boolean comparison in `resolve_github_settings()`

**File:** sessions.py:988,996

`db["pr_review"] == "true"` - "True" or "TRUE" silently evaluates to False. The UI writes lowercase consistently, but `.lower()` is trivially defensive.

**Fix:** Change to `db["pr_review"].lower() == "true"`.

dcellison:
## Summary

Five minor robustness fixes from the #196 code review (child of #235). All defend against edge cases only reachable via direct DB manipulation, not through normal UI usage.

- **Finding 1:** Empty/whitespace model strings in DB now stripped and treated as absent, falling through to yaml or env default instead of reaching `claude.py` as an invalid model name
- **Finding 2:** Six missing env vars added to `_CONFIG_ENV_VARS` test fixture (`CLAUDE_IDLE_TIMEOUT`, `FILE_RETENTION_DAYS`, four `TOTP_*` vars) to prevent silent test pollution from developer shell environment
- **Finding 3:** Clarifying comments on the intentional yaml error handling asymmetry: `users.yaml` fails closed (auth), `workspaces.yaml` fails open (convenience)
- **Finding 4:** Comment explaining why deprecation warnings correctly skip when `users.yaml` is malformed (env vars are actively needed in fallback mode, not deprecated)
- **Finding 5:** Case-insensitive boolean comparison for `pr_review` and `issue_triage` DB values via `.lower()`

## Test plan

- [x] 3 new tests for empty/whitespace model fallthrough (`TestResolveUserDefaults`)
- [x] 2 new tests for case-insensitive booleans (`TestResolveGitHubSettings`)
- [x] Full suite passes (1456 tests)

Fixes #239

### DIFF
--- a/src/kai/sessions.py
+++ b/src/kai/sessions.py
@@ -1002,7 +1002,7 @@ async def resolve_github_settings(chat_id: int, config: Config) -> GitHubSetting
 
     # PR review: DB > yaml > env > False
     if "pr_review" in db:
-        pr_review = db["pr_review"].lower() == "true"
+        pr_review = isinstance(db["pr_review"], str) and db["pr_review"].lower() == "true"
     elif user_config and user_config.pr_review is not None:
         pr_review = user_config.pr_review
     else:
@@ -1010,7 +1010,7 @@ async def resolve_github_settings(chat_id: int, config: Config) -> GitHubSetting
 
     # Issue triage: DB > yaml > env > False
     if "issue_triage" in db:
-        issue_triage = db["issue_triage"].lower() == "true"
+        issue_triage = isinstance(db["issue_triage"], str) and db["issue_triage"].lower() == "true"
     elif user_config and user_config.issue_triage is not None:
         issue_triage = user_config.issue_triage
     else:


dcellison:
## Review by Kai

**Overall: clean, ready to merge.** All findings from the prior two reviews were addressed correctly. No new issues found.

**Verification of the critical finding from review 2:**

`sessions.py` ~1002 and ~1011 - the `isinstance(db["..."], str)` guard is the right fix. `isinstance(None, str)` short-circuits to `False`, so a null DB value no longer raises `AttributeError`. The guard also correctly handles any other non-string type that direct DB manipulation could insert.

**Everything else checks out:**

- `sessions.py` ~704-710: both `db_model` and `yaml_model` are now stripped before the truthiness check. The comment accurately describes why truthiness is safe after stripping.
- `config.py`: comment additions only. The fail-closed/fail-open split and the deprecation guard explanation are both accurate.
- `test_config.py`: the six added env vars (`CLAUDE_IDLE_TIMEOUT`, `FILE_RETENTION_DAYS`, four `TOTP_*`) belong in the isolation fixture.
- `test_sessions.py`: the four boolean tests now cover both truthy (`"True"`, `"TRUE"`) and falsy (`"False"`, `"FALSE"`) paths symmetrically. The `pr_review_enabled=True` / `issue_triage_enabled=True` on the falsy tests is correct - you need the feature on at the config level to confirm the DB value is what drives the `False` result.

No bugs, no security issues, no new style concerns.

Issue #220, opened 2026-04-03
Self-service GitHub repo subscriptions and webhook registration


dcellison:
## Problem

After #203 ships, repo subscriptions in `users.yaml` are admin-managed. Users who want to add or remove repos from their notification list must ask the admin to edit the YAML file. Additionally, adding a new repo requires someone to manually configure a GitHub webhook on that repo pointing to Kai's endpoint.

This issue adds self-service repo management via Telegram and automatic webhook registration via the GitHub API.

Parent issue: #196

Depends on: #203

### Telegram commands

```
/github token <ghp_...>        - store GitHub PAT for webhook management
/github token clear             - remove stored token
/github add <owner/repo>        - subscribe + register webhook (if token available)
/github remove <owner/repo>     - unsubscribe + deregister webhook (if last subscriber)
```

### GitHub token management

Registering a webhook requires `admin:repo_hook` scope on a GitHub personal access token. Each user stores their own token via Telegram:

- Stored in the database per-user (settings table, key `github_token:{chat_id}`)
- Never echoed back, never logged, never included in error messages
- The token is also used to verify repo existence during `/github add`

### Registration flow

When a user runs `/github add owner/repo`:

1. Validate format (`owner/repo`)
2. Check if the user has a stored GitHub token. If not, reply with manual setup instructions and the webhook URL/secret
3. Use the GitHub API to check if a webhook already exists on that repo pointing to Kai's endpoint
4. If no webhook exists, create one via `POST /repos/{owner}/{repo}/hooks` with:
   - `url`: the public webhook URL (from `TELEGRAM_WEBHOOK_URL` base, i.e., `https://api.syrinx.net/webhook/github`)
   - `secret`: the shared `WEBHOOK_SECRET`
   - `events`: `["push", "pull_request", "issues", "issue_comment", "pull_request_review"]`
   - `content_type`: `json`
5. If webhook creation fails (403 - no admin access, 404 - repo not found), reply with the specific error and manual setup instructions
6. On success, add the repo to the user's subscription list

### Shared secret model

All repos use the same `WEBHOOK_SECRET` for HMAC signature validation. This is intentional - the endpoint is already behind Cloudflare with signature verification. Per-repo secrets would require identifying the repo before validating the signature (the body must be read to get the repo name, but the signature covers the body), creating a chicken-and-egg problem.

### Manual fallback

If the user has no token or lacks admin access on the repo, `/github add` still subscribes them to notifications but prints:

```
[...]

PR #254, opened 2026-04-05
Add effective repos computation with union/minus semantics

dcellison:
## Summary

Data layer for self-service GitHub repo subscriptions (#220, PR 1 of 2).

- Add `get_effective_repos()` in sessions.py as the single source of truth for the union/minus formula: `effective = (yaml_repos | db_added) - db_removed`
- Add get/set helpers for `github_repos_added` and `github_repos_removed` per-user settings (JSON arrays in the settings table)
- Update `resolve_github_settings()` to delegate repo resolution to `get_effective_repos()`
- Update `get_github_db_settings()` to include the two new keys
- Make `_get_subscribed_users()` async in webhook.py so it can call `get_effective_repos()` per user
- Admin wildcard behavior unchanged (empty effective repos = wildcard for admins)

No new user-facing commands in this PR. PR 2 will add `/github add`, `/github remove`, `/github token`, and the GitHub API module.

## Test plan

- [x] 9 new tests for effective repos computation (yaml-only, db-added, db-removed, union/minus, cancel-out, empty, case normalization, corrupt JSON)
- [x] Existing `_get_subscribed_users` tests converted from sync to async with passthrough mock
- [x] Existing `resolve_github_settings` tests pass with new repos resolution
- [x] Full suite passes (1493 tests)

Ref #220

### DIFF
--- a/src/kai/sessions.py
+++ b/src/kai/sessions.py
@@ -1009,7 +1009,12 @@ async def get_effective_repos(chat_id: int, yaml_repos: list[str]) -> list[str]:
 async def get_github_db_settings(chat_id: int) -> dict[str, str]:
     """Read all GitHub-related DB overrides for a user.
 
-    Returns a dict of key->value for settings that exist.
+    Returns a dict of key->value for settings that exist. All values
+    are raw strings. Note: "github_repos_added" and "github_repos_removed"
+    are JSON-encoded arrays (e.g., '["owner/repo"]'), not parsed lists.
+    Use get_github_added_repos()/get_github_removed_repos() for parsed
+    access, or get_effective_repos() for the resolved list.
+
     Keys: "pr_review", "issue_triage", "github_notify_chat",
     "github_repos_added", "github_repos_removed".
     """


dcellison:
## Review by Kai

The two issues from the prior review were both addressed: the `assert` is replaced with an explicit `isinstance` check + `ValueError`, and `get_effective_repos` now lowercases `added` and `removed` consistently. Everything else looks correct.

One small new finding:

---

### Suggestion

**`effective` is a list; `in` does a linear scan (webhook.py:590)**

```python
effective = await sessions.get_effective_repos(uc.telegram_id, uc.github_repos)
if repo_lower in effective:
```

`get_effective_repos` returns a sorted list. `in` on a list is O(n) - fine for small repo counts today, but since `effective` is already sorted, `bisect.bisect_left` would give O(log n) at no real cost. Alternatively, returning a `set` from `get_effective_repos` would make membership checks O(1), though it would lose the deterministic sort that the tests rely on. Worth considering if the repo lists grow.

---

Everything else is clean. The `ValueError` fix is correct and the corrupt-JSON tests now exercise the right code path. The `elif uc.role == "admin" and not effective:` logic preserves the wildcard semantics correctly. No security issues, no auth bypass, no new edge cases missed.

dcellison:
## Review by Kai

## PR Review: Add effective repos computation with union/minus semantics (iteration 3)

Both prior-review findings were addressed and the new code is clean. No new issues found.

The `assert`-to-explicit-check fix is correct: `ValueError` is raised inside the `try` block and caught alongside `json.JSONDecodeError`, so `test_corrupt_removed_repos` with `"{}"` now exercises the right code path. The defensive lowercasing of `added` and `removed` in `get_effective_repos` is consistent with `yaml_repos`.

The async conversion of `_get_subscribed_users` is wired correctly at both the definition (`async def`) and call site (`await _get_subscribed_users`, webhook.py:615). The `explicitly_subscribed`/`admin_wildcards` deduplication invariant holds: a user with `effective = []` cannot satisfy `repo_lower in effective`, so no user can appear in both lists.

No security issues: `chat_id` is typed `int` so no key injection, `json.loads` is safe, repo matching is by equality after lowercasing.

The PR is ready to merge.

PR #257, opened 2026-04-06
Fix service proxy truncating chunked HTTP responses

### LINKED ISSUE #256: Service proxy truncates chunked HTTP responses from external APIs
## Bug

The service proxy in `services.py` silently truncates responses from external APIs that use chunked transfer encoding. Perplexity's API always uses `Transfer-Encoding: chunked` (no `Content-Length` header), so every `sonar-pro` response is cut short, producing invalid JSON that the inner Claude cannot parse.

Simpler `sonar` queries sometimes fit in a single chunk and work by accident.

## Root Cause

`call_service()` uses `resp.content.read(n)` to read the response body. This is the streaming API - it returns as soon as up to `n` bytes are available. With chunked encoding, "available" means "one chunk arrived," not "full response received."

The fix is to use `resp.read()` instead, which buffers all chunks before returning. The existing 10MB truncation guard still applies after the read.

## Evidence

Direct test with the same query, same API key, same aiohttp version:

- `resp.content.read(10MB)` returned 1022 bytes (truncated, invalid JSON)
- `resp.read()` returned 6613 bytes (complete, valid JSON)

## Spec

`home/specs/fix-service-proxy-chunked-responses.md`

dcellison:
## Summary

- Switch `resp.content.read(n)` to `resp.read()` in `call_service()` to fix truncated responses from APIs using chunked transfer encoding (like Perplexity)
- `resp.content.read(n)` returns after the first chunk; `resp.read()` buffers all chunks before returning
- Existing `_MAX_RESPONSE_BYTES` truncation guard still caps oversized responses

## Changes

- **`src/kai/services.py`**: One-line fix - `resp.content.read(_MAX_RESPONSE_BYTES + 1)` to `resp.read()`
- **`tests/test_services.py`**: Updated mock helper to use `resp.read()` instead of `resp.content.read()`, added test verifying chunked responses are read completely

## Test plan

- [x] All 1544 existing tests pass
- [x] New `test_chunked_response_read_completely` verifies full body is captured
- [x] Existing truncation test still verifies oversized responses are capped

Fixes #256

### DIFF
--- a/tests/test_services.py
+++ b/tests/test_services.py
@@ -1087,6 +1087,48 @@ services:
         # All three chunks should be joined into the complete body
         assert result.body == expected.decode("utf-8")
 
+    async def test_oversized_multi_chunk_truncated(self, tmp_path, monkeypatch):
+        """The streaming loop breaks early when multiple chunks exceed the cap.
+
+        Unlike test_oversized_response_truncated (single chunk), this
+        verifies the break fires mid-iteration when cumulative chunk
+        size crosses _MAX_RESPONSE_BYTES.
+        """
+        monkeypatch.setenv("API_KEY", "tok")
+        path = _write_yaml(
+            tmp_path,
+            """
+services:
+  testapi:
+    url: https://api.example.com
+    method: POST
+    auth:
+      type: bearer
+      env: API_KEY
+""",
+        )
+        load_services(path)
+
+        from kai.services import _MAX_RESPONSE_BYTES
+
+        # Three chunks whose total exceeds the cap. The loop should
+        # break after the second chunk pushes total past the limit.
+        half = _MAX_RESPONSE_BYTES // 2
+        remainder = _MAX_RESPONSE_BYTES - half + 500
+        chunks = [b"a" * half, b"b" * remainder, b"c" * 1000]
+        mock_response = _mock_multi_chunk_response(chunks)
+        mock_session = MagicMock()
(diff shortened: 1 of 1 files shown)


dcellison:
## Review by Kai

## PR Review: Fix service proxy truncating chunked HTTP responses (v2)

Both issues from the prior review have been addressed. Reviewing only the new diff.

### Clean

**`src/kai/services.py` - streaming loop**

The implementation is correct. The break-then-join approach works:

- If response fits: `total <= _MAX_RESPONSE_BYTES`, loop drains naturally, `truncated = False`
- If response overshoots: we accumulate at most `_MAX_RESPONSE_BYTES + 65535` bytes (one extra chunk) before breaking, `truncated = True`, then `raw[:_MAX_RESPONSE_BYTES]` trims it

This is the same semantic as the original `+1` read trick, just with a larger worst-case over-read (one chunk vs one byte). Not a concern at 64KB.

**`tests/test_services.py` - mock helpers**

`_mock_streamed_response` now correctly mocks `iter_chunked` and the docstring is accurate. `_mock_multi_chunk_response` yields actual separate chunks and the test exercises the reassembly path end-to-end. The comment on the new test (`test_multi_chunk_response_reassembled`) accurately describes what is being tested.

**No new issues.** Both prior findings are resolved, no regressions introduced, no auth/injection/exposure concerns. This is good to merge.

PR #267, opened 2026-04-07
Warn when goose binary is missing during install

### LINKED ISSUE #260: GooseBackend: Goose ACP as an alternative agent harness
## Summary

Add `GooseBackend` as a second concrete implementation of the `AgentBackend` ABC (introduced in #259), using Goose's ACP (Agent Client Protocol) server mode as the agent harness instead of Claude Code.

## Motivation

Claude Code as an agent harness has limitations:
- Tied to Anthropic's subscription model
- No provider flexibility (can't swap to OpenAI, Ollama, etc. without a rewrite)
- Single-provider cost structure

`GooseBackend` makes the agent harness a config choice. Setting `AGENT_BACKEND=goose` switches the entire pool to Goose without touching `bot.py`, pool logic, or session handling. `ClaudeCodeBackend` remains the default and continues to work unchanged.

## Architecture

Goose ACP (`goose acp`) is a persistent JSON-RPC 2.0 process over stdin/stdout - architecturally equivalent to how Kai uses Claude Code today. One long-lived subprocess per user, conversation state kept across messages, killed/restarted on `/new` or workspace switch.

The wire protocol differs (JSON-RPC 2.0 vs Claude's custom stream-json), but the lifecycle maps directly to the `AgentBackend` ABC. The ACP protocol was verified by live testing before this issue was written.

## Sub-issues

- [ ] #261 - Config, install, and deployment wiring
- [ ] #262 - GooseBackend class and pool integration

## Prerequisites

- #259 (merged) - `AgentBackend` ABC
- `goose` binary installed on the target system (`brew install block-goose-cli`)

## Reference

Implementation spec: `home/specs/goose-backend.md`

dcellison:
## Summary

- Add `shutil.which("goose")` check in `_apply_goose_config()` that warns when the goose binary is not on PATH during `make install`
- Non-fatal (install continues) since goose may be installed afterward, but gives a clear message instead of an opaque runtime error
- Closes the last gap in the Goose backend parent issue (#260)

Fixes #260

## Test plan

- [x] `make check` passes
- [x] `make test` passes (1613 tests)
- [x] New tests verify warning appears when goose is missing and is absent when goose is found

### DIFF
--- a/src/kai/install.py
+++ b/src/kai/install.py
@@ -1848,18 +1848,20 @@ def _apply_goose_config(
     if not src.exists():
         raise SystemExit(f"Goose config template not found at {src}")
 
-    # Warn if the goose binary isn't on PATH. Not fatal because the
-    # user may install it after running make install, but a clear
-    # message now saves debugging an opaque runtime error later.
-    if not shutil.which("goose"):
-        print("  WARNING: 'goose' binary not found on PATH.")
-        print("  Kai will fail to start the Goose backend until goose is installed.")
-        print("  See https://github.com/block/goose for installation instructions.")
-
     if dry_run:
         print(f"[DRY RUN] Would create: {goose_dir}")
         print(f"[DRY RUN] Would copy: {src} -> {dst}")
         return
+
+    # Warn if the goose binary isn't on PATH. Not fatal because the
+    # user may install it after running make install, but a clear
+    # message now saves debugging an opaque runtime error later.
+    # Placed after the dry-run guard so dry runs don't warn about
+    # runtime dependencies.
+    if not shutil.which("goose"):
+        print("  WARNING: 'goose' binary not found on PATH.")
+        print("  Kai will fail to start the Goose backend until goose is installed.")
+        print("  See https://github.com/block/goose for installation instructions.")
 
     # Track whether we're creating .config for the first time so we
     # can set ownership on it below. mkdir(parents=True) creates both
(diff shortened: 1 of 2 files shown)


dcellison:
## Review by Kai

**Overall: Clean, functional change. Two minor findings.**

---

**WARNING - `install.py:1851-1853`: Warnings printed to stdout**

All three warning lines use bare `print()`, which writes to stdout. Warnings typically belong on stderr so they can be separated from normal install output (e.g. piping stdout to a log file silently swallows them). Should be `print("...", file=sys.stderr)` or `warnings.warn(...)`.

---

**SUGGESTION - `install.py:1851`: Warning fires in dry-run mode**

The `shutil.which` check sits above the `if dry_run:` block, so running `make install --dry-run` on a system without goose will print the warning even though nothing is being installed. Probably benign, but it's an inconsistency with dry-run semantics ("no observable side effects"). Moving the check inside the `not dry_run` branch, or at least prefixing the output with `[DRY RUN]` when applicable, would be cleaner.

---

**SUGGESTION - `test_install.py:2113,2135`: Monkeypatch silences all `shutil.which` calls**

```python
monkeypatch.setattr("shutil.which", lambda cmd: None)
```

This disables `shutil.which` globally for the duration of the test, not just for `"goose"`. If `_apply_goose_config` (or anything it calls) uses `shutil.which` for other lookups, they'd silently return `None`. More precise:

```python
monkeypatch.setattr("shutil.which", lambda cmd: None if cmd == "goose" else shutil.which(cmd))
```

Not a correctness bug today, but worth tightening.

---

No security issues. No critical bugs.

PR #272, opened 2026-04-07
Provider-aware model selection with per-user backend routing

### LINKED ISSUE #271: Provider-aware model selection with per-user backend/provider
The `/models` keyboard, `/model` command, and `/settings model` are hardcoded to Claude models (opus, sonnet, haiku). With the Goose backend supporting multiple providers (#270), these need to reflect the actual available models for the user's provider.

## Problem

- `/models` always shows Claude Opus/Sonnet/Haiku regardless of provider
- `/model gpt-4o` is rejected even when the user is on Goose+OpenAI
- No way to configure different users on different backends/providers

## Requirements

1. **Provider-aware model lists.** Curated providers (anthropic, openai, google) show an inline keyboard with known models. Open-ended providers (openrouter, ollama) show the current model and accept any model ID via `/model <id>`.
2. **Per-user backend/provider.** Different users can be on different backends (claude vs goose) and providers (anthropic, openai, etc.) simultaneously. Admin-controlled via users.yaml.
3. **Model validation.** `/model`, `/settings model`, and `/workspace config model` validate against the user's effective provider, not a global list.

## Spec

`home/specs/provider-aware-model-selection.md`

Follows up on #270 (which explicitly deferred per-user provider selection and model name validation per provider as out of scope).

dcellison:
## Summary

Implements #271 - provider-aware model selection for the Goose backend epic (#260).

- **Model registry** (`PROVIDER_MODELS`, `PROVIDER_DEFAULTS`, `OPEN_ENDED_PROVIDERS`) replaces the hardcoded `VALID_MODELS` set. Curated providers (anthropic, openai, google) show inline keyboards; open-ended providers (openrouter, ollama) accept free-text model IDs.
- **Per-user backend/provider** - `agent_backend` and `goose_provider` fields on `UserConfig` let different users run on different providers simultaneously, with cascading fallback to global config.
- **Pool routing** - `_create_instance()` resolves backend, provider, and model per-user. `_restore_workspace()` validates stored model overrides against the current provider, falling back to the provider default if invalid.
- **`provider` attribute on ABC** - bot.py reads `instance.provider` to determine which model UI to show. Set by both `ClaudeCodeBackend` ("anthropic") and `GooseBackend` (from goose_provider).
- **`CLAUDE_MODEL` renamed to `DEFAULT_MODEL`** - backward compat preserved (reads `CLAUDE_MODEL` when `DEFAULT_MODEL` is absent). Standalone deprecation warning fires for the old name. Rename propagated through config, pool, sessions, main, install, and .env.example.
- **Startup validation** - `DEFAULT_MODEL` is validated against the effective global provider at boot. Per-user models validated against the user's effective provider in `_load_user_configs()`.

## Test plan

- [x] All 1624 tests pass
- [x] Ruff lint + format clean
- [ ] Manual test: `/models` shows correct keyboard for anthropic user
- [ ] Manual test: `/model` with invalid provider model is rejected
- [ ] Manual test: `make config` shows provider-appropriate model choices
- [ ] Verify startup fails with `DEFAULT_MODEL=opus` when global provider is openai

Fixes #271

### DIFF
--- a/tests/test_bot.py
+++ b/tests/test_bot.py
@@ -910,6 +910,21 @@ class TestHandleModels:
         assert "Choose a model" in call[0][0]
         assert call[1]["reply_markup"] is not None
 
+    @pytest.mark.asyncio
+    async def test_open_ended_provider_shows_text(self):
+        """Open-ended provider (openrouter/ollama) shows text instead of keyboard."""
+        pool = _make_mock_claude(model="llama3", provider="ollama")
+        update = _make_update()
+        ctx = _make_context(claude=pool)
+        await handle_models(update, ctx)
+        call = update.message.reply_text.call_args
+        reply = call[0][0]
+        assert "Current model" in reply
+        assert "llama3" in reply
+        assert "/model" in reply
+        # No keyboard for open-ended providers
+        assert "reply_markup" not in call[1] or call[1].get("reply_markup") is None
+
 
 # ── handle_model ─────────────────────────────────────────────────────
 
--- a/tests/test_pool.py
+++ b/tests/test_pool.py
@@ -19,6 +19,7 @@ from unittest.mock import AsyncMock, MagicMock, patch
 import pytest
 
 from kai.config import Config, UserConfig
+from kai.goose import GooseBackend
 from kai.pool import SubprocessPool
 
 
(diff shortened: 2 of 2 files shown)


dcellison:
## Review by Kai

## PR Review: Provider-aware model selection (#272) - Follow-up

The prior review issues are addressed. The open-ended provider fallback now warns in `pool.py`; `validate_model_for_provider` logs for unknown providers; `UserConfig` fields have coverage; the redundant variable is gone. A few new items:

---

### Warning: `install.py` - stale `CLAUDE_MODEL` key may persist after migration

**`install.py` line ~702**

The diff changes `env["CLAUDE_MODEL"] = model` to `env["DEFAULT_MODEL"] = model`, but if `_cmd_config` initializes `env` from `existing_env`, old `CLAUDE_MODEL` entries in the env file are not explicitly removed. After `make config`, the file could have both keys. The deprecation warning won't fire (guarded by `not os.environ.get("DEFAULT_MODEL")`), making the stale key invisible to operators. Fix: add `env.pop("CLAUDE_MODEL", None)` before writing if `DEFAULT_MODEL` is being set.

---

### Suggestion: `/settings` hides provider for users with no active instance

**`bot.py`, `_show_settings` (~line 732)**

```python
instance = pool.get_if_exists(chat_id)
provider_line = f"\n  Provider: {instance.provider}" if instance else ""
```

`get_if_exists` returns `None` before any session starts, producing an empty `provider_line`. The effective provider is fully derivable from `config.get_user_config(chat_id)` and `config.agent_backend`/`config.goose_provider` without an instance - the same cascade `_create_instance` uses. New users checking `/settings` see no provider info.

---

### Suggestion: `_show_settings` tests don't exercise the new provider line

**`test_bot.py`, `test_show_settings_defaults`, `test_show_settings_budget_zero_displays_unlimited`, `test_show_settings_global_ceiling_visible`**

The three updated tests pass bare `MagicMock()` as `context`. `_get_pool(MagicMock())` returns a MagicMock, so `pool.get_if_exists(chat_id)` is truthy and `instance.provider` is a MagicMock repr that ends up in the reply text. The `assert "sonnet" in reply` checks pass but the provider line is unverified noise. The `_make_mock_claude` helper already sets up `pool.get_if_exists` correctly - the `_show_settings` tests should use it instead of raw `MagicMock()`.

---

**Overall:** The structural changes are well-executed and the prior gaps are addressed. The install migration issue is the only one with operational risk (stale config key that could confuse a future operator). Everything else is clean.

dcellison:
## Review by Kai

## PR Review: Provider-aware model selection (#272) - Follow-up

All three issues from the prior review (install.py migration, `/settings` no-instance provider display, `_show_settings` test setup) are addressed in this diff. The implementation is clean. Two new items:

---

### Warning: `pool.py` - open-ended provider warning missing from the global-same-provider branch

**`pool.py`, `_create_instance` (~line 147)**

The prior review flagged the open-ended fallback bug, and the fix adds a warning in the `else` branch. But the `elif` branch has the same gap:

```python
elif effective_provider == global_provider:
    model = self._config.default_model   # no warning if provider is open-ended
else:
    model = PROVIDER_DEFAULTS.get(effective_provider, "")
    if not model:
        log.warning(...)                 # warning only fires here
        model = self._config.default_model
```

When the global backend is "goose" with `GOOSE_PROVIDER=openrouter` and a user has no per-user model, `effective_provider == global_provider == "openrouter"`, so the `elif` branch runs and `config.default_model` (e.g., "sonnet") is used silently. Startup validation lets it through because `validate_model_for_provider` returns `True` for all open-ended models. The `_load_user_configs` warning covers users.yaml users, but legacy mode (no users.yaml) is entirely silent. Fix: add the same `OPEN_ENDED_PROVIDERS` check and warning in the `elif` branch.

---

### Suggestion: `test_bot.py` - `_show_settings` pool mock path is opaque

**`test_bot.py`, `test_show_settings_defaults` (~line 3292)**

The three updated tests now use `ctx = _make_context(config=config)` and assert `"anthropic" in reply.lower()`. Whether this assertion tests the new provider line or is vacuously satisfied depends on what `_make_context` does with the pool - specifically, whether `_get_pool(ctx).get_if_exists(12345)` returns `None` (triggering the config-derived path that yields "anthropic") or a properly configured instance. `_make_context` isn't in the diff. Worth a quick confirm that the assertion is actually exercising the provider line rather than passing by coincidence.

---

**Overall:** The structural work is solid and the prior gaps are closed. The open-ended provider warning gap in the `elif` branch is the only meaningful new finding.

PR #275, opened 2026-04-07
Fix model switch not clearing workspace config override

dcellison:
## Summary

- `_switch_model()` now clears any workspace config model override for the current workspace when switching models via `/model`, `/models`, or `/settings model`
- Without this, a prior `/workspace config model` entry shadows the user setting (workspace config has higher precedence) and the model reverts on the next service restart
- Fixes the case where `/workspace config` shows one model and `/models` shows another

## Test plan

- [x] `make check` passes
- [x] `make test` passes (1639 tests)
- [ ] Manual: `/workspace config model opus`, then `/model sonnet`, then `/workspace config` - verify model shows sonnet, not opus
- [ ] Manual: restart service, verify model stays sonnet (not reverting to opus)


dcellison:
## Review by Kai

The prior warning about missing test assertions has been fixed. Both tests now correctly call `mock_delete.assert_called_once_with(ANY, ANY, "model")` at `tests/test_bot.py:963` and `tests/test_bot.py:1010`.

The prior suggestion about `str(None)` at `bot.py:441` was not addressed - the code is unchanged. Per review policy I'm not re-raising it; the author has seen it and made their call.

No new issues found. The diff is clean.

PR #277, opened 2026-04-07
Fix /models showing wrong model after service restart

### LINKED ISSUE #276: Models command shows wrong model after service restart
## Problem

After a service restart, `/models` shows the global default model (e.g., sonnet) instead of the user's persisted model choice. This happens because `pool.get_model()` returns `config.default_model` when no subprocess instance exists yet, and instances are only created lazily on the first real message via `pool.send()`.

Meanwhile, `/workspace config` reads the DB directly and shows the correct persisted value, creating a visible contradiction between the two commands.

The discrepancy resolves itself once the user sends any regular message (which triggers instance creation and `_restore_workspace`), but the window after every restart is confusing.

## Root cause

`pool.get_model()` in `pool.py:396-399`:

```python
def get_model(self, chat_id: int) -> str:
    instance = self.get_if_exists(chat_id)
    return instance.model if instance else self._config.default_model
```

When no instance exists, it returns the global default instead of checking persisted user settings.

## Fix

Have `handle_models` (and `handle_model`) fall back to `resolve_user_defaults()` when no instance exists, so the displayed model reflects the user's actual persisted choice rather than the global default.

dcellison:
## Summary

- Add `pool.get_effective_model()` which resolves the model from persisted DB settings (via `resolve_user_defaults()`) when no subprocess instance exists yet
- Update `handle_models` to use it so `/models` shows the correct model after a service restart, before the first message
- Change `/settings` context display from unhelpful "default" to "not set (model default)" when no context window override is configured

Fixes #276

## Test plan

- [x] `make check` passes
- [x] `make test` passes (1639 tests)
- [ ] Manual: restart service, immediately send `/models` - verify correct model is highlighted
- [ ] Manual: send `/settings` - verify context shows "not set (model default)" instead of "default"

### DIFF
--- a/src/kai/bot.py
+++ b/src/kai/bot.py
@@ -476,7 +476,9 @@ async def handle_model_callback(update: Update, context: ContextTypes.DEFAULT_TY
         await query.answer("Invalid model.")
         return
 
-    if model == pool.get_model(chat_id):
+    # Use get_effective_model so the comparison matches the keyboard highlight,
+    # which also uses get_effective_model to mark the active model.
+    if model == await pool.get_effective_model(chat_id):
         await query.answer()
         await query.edit_message_text("No change.", reply_markup=InlineKeyboardMarkup([]))
         return


dcellison:
## Review by Kai

**Overall: clean fix.** The logic is correct and the change is well-scoped. Two findings:

---

**[suggestion] `pool.py` - missing test for the actual bug-fix path**

The `test_bot.py` change (`pool.get_effective_model = AsyncMock(return_value=model)`) only prevents existing tests from breaking - it doesn't exercise the new code path in `get_effective_model()` where `get_if_exists` returns `None` and the method falls back to `resolve_user_defaults`. The core of the bug fix has no dedicated test.

A unit test in `test_pool.py` (or wherever pool tests live) that mocks `get_if_exists` returning `None` and `sessions.resolve_user_defaults` returning a known dict would close this gap.

---

**[suggestion] `pool.py:415` - unguarded key access on `resolve_user_defaults` result**

```python
defaults = await sessions.resolve_user_defaults(chat_id, self._config)
return defaults["model"]
```

If `resolve_user_defaults` ever returns a dict without a `"model"` key (e.g., partial result due to a new code path or DB schema change), this raises a bare `KeyError` that propagates up to `handle_models` and likely surfaces as an unhandled exception to the user. A `.get("model", self._config.default_model)` fallback would be more defensive - though this is low risk if the function's contract is stable and already relied on elsewhere.

---

The `ctx_label` change in `bot.py:750` is straightforwardly better UX. No issues there.

PR #305, opened 2026-04-15
Rename CLAUDE_MAX_BUDGET_USD to BUDGET_CEILING, separate ceiling from default

### LINKED ISSUE #304: Separate budget ceiling from per-user default
## Problem

`CLAUDE_MAX_BUDGET_USD` does double duty as both the global default budget and the budget ceiling. The per-user `max_budget` field in users.yaml also does double duty as both the per-user default and the per-user ceiling. When a user changes their `max_budget` in users.yaml, the displayed "Budget ceiling" changes with it - defeating the purpose of a ceiling.

With API-backed model access, the budget controls real spending. A ceiling that moves with the user's own setting is no ceiling at all.

## Changes

1. **Rename `CLAUDE_MAX_BUDGET_USD` to `BUDGET_CEILING`** - env var, config field, all references
2. **Use `BUDGET_CEILING` as the actual ceiling** - `max_budget` in users.yaml should only serve as a per-user default, never as a ceiling. Ceiling always comes from the global env var.
3. **Remove "Budget ceiling" display from `/settings`** - it is not useful information for the user

## Details

- Default value for `BUDGET_CEILING`: `10.0` (real money with API models)
- Backward compatibility: if `CLAUDE_MAX_BUDGET_USD` is set and `BUDGET_CEILING` is not, read the old var with a deprecation warning. `BUDGET_CEILING` takes precedence if both are set.
- `0` means no ceiling enforced (preserve existing behavior)
- `max_budget` field in `UserConfig` / users.yaml stays as-is - it remains useful as a per-user default
- **Behavioral change:** per-user `max_budget` no longer acts as a ceiling. Users can `/settings budget` up to the global `BUDGET_CEILING`, regardless of their `max_budget` default. This is intentional - the per-user field was never designed as a ceiling.
- Update docstrings and `users.example.yaml` comments to clarify `max_budget` is only a default

## Files affected

- `config.py` - field rename, env parsing with backward compat fallback, deprecation entry
- `bot.py` - ceiling enforcement in `/settings budget` and `/workspace config budget`, remove ceiling display from `_show_settings()`
- `pool.py` - field name reference
- `sessions.py` - field name reference
- `install.py` - env key reads (check old key on upgrade) and writes (emit new key)
- `.env.example` - rename entry
- `users.example.yaml` - rewrite `max_budget` comment block (remove "double duty" language)
- `README.md` - all references
- Tests: `test_config.py`, `test_bot.py`, `test_pool.py`, `test_sessions.py`

Full spec: `home/specs/budget-ceiling-fix.md`

dcellison:
## Summary

- Rename `CLAUDE_MAX_BUDGET_USD` to `BUDGET_CEILING` across config, bot, pool, sessions, install, docs, and tests
- Separate ceiling enforcement from per-user defaults: the ceiling always comes from the global `BUDGET_CEILING` env var, while `max_budget` in users.yaml is only a default
- Remove the "Budget ceiling" line from `/settings` output
- Add backward-compat parsing so existing `.env` files with the old key name still work on upgrade

## Details

Previously, `CLAUDE_MAX_BUDGET_USD` did double duty as both the global default budget and the ceiling. Per-user `max_budget` in users.yaml also conflated default and ceiling. When a user changed their `max_budget`, the displayed ceiling changed with it, defeating the purpose.

Now the ceiling is always `config.budget_ceiling` (the global env var). Per-user `max_budget` is only a baseline default. A user with `max_budget: 15` in users.yaml can `/settings budget 40` if the global ceiling is $50.

Backward compat: `BUDGET_CEILING` > `CLAUDE_MAX_BUDGET_USD` > `10.0`. `install.py` reads from either key on upgrade, writes the new key, and removes the old one.

## Test plan

- [x] All 1686 tests pass
- [x] Backward compat tests: old env var still works, new var takes precedence when both are set
- [x] New test: budget above yaml default but under global ceiling succeeds
- [x] Ceiling enforcement tests updated for global-only behavior

Fixes #304

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -40,9 +40,9 @@ TELEGRAM_BOT_TOKEN=your-token-from-botfather
 # users.yaml or use /settings timeout via Telegram.
 #CLAUDE_TIMEOUT_SECONDS=120
 
-# DEPRECATED when users.yaml exists - set per-user 'max_budget' in
-# users.yaml or use /settings budget via Telegram.
-#CLAUDE_MAX_BUDGET_USD=10.0
+# Global budget ceiling in USD. Users cannot exceed this via /settings.
+# Per-user defaults are set via 'max_budget' in users.yaml.
+#BUDGET_CEILING=10.0
 
 # Maximum session age in hours before recycling (0 = no limit).
 # Prevents unbounded memory growth in the inner Claude process.
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ users:
     github: alice-dev     # routes GitHub events to this user
     os_user: alice        # subprocess runs as this OS account
     home_workspace: /home/alice/workspace
-    max_budget: 15.0      # ceiling for /settings budget (CLAUDE_MAX_BUDGET_USD is the default)
+    max_budget: 15.0      # default budget for this user (BUDGET_CEILING is the global ceiling)
 ```
 
 Each user gets:
@@ -194,7 +194,7 @@ cp .env.example .env
 | `ALLOWED_USER_IDS` | Deprecated* | | Comma-separated Telegram user IDs. Superseded by `users.yaml`; use `make config` instead. (*Still works if `users.yaml` absent.) |
 | `CLAUDE_MODEL` | Deprecated* | `sonnet` | Default model. Set per-user in `users.yaml` or via `/settings model`. |
 | `CLAUDE_TIMEOUT_SECONDS` | Deprecated* | `120` | Per-message timeout. Set per-user in `users.yaml` or via `/settings timeout`. |
-| `CLAUDE_MAX_BUDGET_USD` | Deprecated* | `10.0` | Session budget cap. Set per-user in `users.yaml` or via `/settings budget`. |
+| `BUDGET_CEILING` | No | `10.0` | Global budget ceiling in USD. Users cannot exceed this via `/settings budget`. Per-user defaults are set via `max_budget` in `users.yaml`. |
 | `CLAUDE_MAX_CONTEXT_WINDOW` | Deprecated* | `0` | Context window size in tokens (0 = backend default). Set per-user in `users.yaml` or via `/settings context`. |
 | `CLAUDE_AUTOCOMPACT_PCT` | No | `80` | Context compression threshold %, Claude Code only. When usage hits this, Claude compresses history. Can only lower the default (~83%), not raise it. |
 | `CLAUDE_MAX_SESSION_HOURS` | No | `0` | Maximum session age in hours before recycling the subprocess (0 = no limit). Recommended: 4-8 on memory-constrained machines. |
@@ -221,7 +221,7 @@ cp .env.example .env
 
 *Deprecated vars still work for backward compatibility when `users.yaml` is absent. When `users.yaml` is present, `users.yaml` wins and a warning is logged. Run `make config` to migrate.
 
-`CLAUDE_MAX_BUDGET_USD` limits spending in a single session. For Claude Code, this maps to the `--max-budget-usd` CLI flag (on Pro/Max plans, purely a runaway prevention mechanism). Goose does not currently enforce this limit. The session resets on `/new`, model switch, or workspace switch.
+`BUDGET_CEILING` limits spending in a single session. For Claude Code, this maps to the `--max-budget-usd` CLI flag (on Pro/Max plans, purely a runaway prevention mechanism). Goose does not currently enforce this limit. The session resets on `/new`, model switch, or workspace switch.
 
 ## Running
 
--- a/src/kai/bot.py
+++ b/src/kai/bot.py
@@ -599,15 +599,9 @@ async def handle_settings(update: Update, context: ContextTypes.DEFAULT_TYPE) ->
         except ValueError:
             await update.message.reply_text("Budget must be a positive number.")
             return
-        # Enforce ceiling: users.yaml max_budget or global default.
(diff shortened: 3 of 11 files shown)


dcellison:
## Review by Kai

## PR Review: fix/budget-ceiling-rename

Overall this is a clean, well-scoped rename with correct backward-compat logic. One notable bug in the install path.

---

### Warning: `install.py` - BUDGET_CEILING not written for users with users.yaml

**File:** `install.py`, around the block at line ~702 in the diff.

```python
env.pop("CLAUDE_MAX_BUDGET_USD", None)
if not users_yaml_exists:
    env["DEFAULT_MODEL"] = model
    env["CLAUDE_TIMEOUT_SECONDS"] = timeout
    env["BUDGET_CEILING"] = budget   # <-- inside the users_yaml guard
```

`BUDGET_CEILING` is a global ceiling, not a deprecated per-user var - but it's being written inside the same `if not users_yaml_exists:` block that guards legacy single-user vars. When `users_yaml_exists=True`, running `make config` will:

1. Read `budget` from `BUDGET_CEILING` or `CLAUDE_MAX_BUDGET_USD` (correct)
2. Pop `CLAUDE_MAX_BUDGET_USD` from the env (correct)
3. Never write `BUDGET_CEILING` (bug)

Any installation that has users.yaml and runs `make config` silently loses its budget ceiling config. The old key is deleted; the new key is never created. The backward-compat fallback in `load_config()` continues to work until `make config` is run, at which point the protection disappears.

**Fix:** Move `env["BUDGET_CEILING"] = budget` outside the `if not users_yaml_exists:` guard. Only `DEFAULT_MODEL`, `CLAUDE_TIMEOUT_SECONDS`, etc. are per-user legacy vars - `BUDGET_CEILING` is global and should always be written.

---

### Suggestion: Users can no longer see their ceiling before hitting it

The removal of the `ceiling_line` from `/settings` output means users are surprised when rejected ("Budget cannot exceed $X.XX") with no prior indication of X. Previously the ceiling was visible proactively. Consider restoring a ceiling line sourced from `config.budget_ceiling` (now that the source is always global, it's simpler to display).

---

### Suggestion: Redundant comment repeated 4-5 times

`# budget_ceiling doubles as the fallback default for unconfigured users` appears in `bot.py` (lines ~712, ~804, ~1522), `pool.py` (~158), and `sessions.py` (~726). One canonical comment in `config.py`'s docstring is sufficient; inline repetition adds noise without adding clarity.

---

### Suggestion: Misleading deprecation warning message in `config.py`

```python
"CLAUDE_MAX_BUDGET_USD": "Renamed to BUDGET_CEILING. Set per-user 'max_budget' in users.yaml or use /settings budget",
```

The second clause ("Set per-user 'max_budget' in users.yaml") conflates the ceiling with the per-user default - which is exactly the conflation this PR is fixing. The message should clarify that `max_budget` in users.yaml is now a default, not a ceiling replacement.

---

No security issues, no auth bypass vectors. The core logic (ceiling always from `config.budget_ceiling`, backward-compat precedence in `load_config()`, test coverage for both old and new var names) is correct.

dcellison:
## Review by Kai

## PR Review: fix/budget-ceiling-rename

### Prior review issues

**[FIXED] install.py - `BUDGET_CEILING` now written outside the users.yaml guard.**
The block now reads:
```python
env.pop("CLAUDE_MODEL", None)
env.pop("CLAUDE_MAX_BUDGET_USD", None)
env["BUDGET_CEILING"] = budget   # outside the guard - correct
if not users_yaml_exists:
    env["DEFAULT_MODEL"] = model
    env["CLAUDE_TIMEOUT_SECONDS"] = timeout
```
The critical bug is resolved.

**[FIXED] Deprecation warning message** in `config.py` is now clear: `"Renamed to BUDGET_CEILING (global ceiling). Per-user defaults go in users.yaml 'max_budget'"`. No longer conflates ceiling with per-user default.

The two suggestions (ceiling visibility in `/settings`, redundant comments) were not addressed - noted, moving on.

---

### New findings

**Warning: install.py - non-interactive path writes `BUDGET_CEILING=` when neither key exists**

`install.py:464`:
```python
budget = existing_env.get("BUDGET_CEILING", existing_env.get("CLAUDE_MAX_BUDGET_USD", ""))
```
If an existing `.env` has neither key, `budget = ""`, and then unconditionally:
```python
env["BUDGET_CEILING"] = budget  # writes BUDGET_CEILING= (empty)
```

`load_config()` handles empty strings correctly (falls through to 10.0), so this doesn't break anything - but it writes a `BUDGET_CEILING=` line to the `.env` file that looks like it's configured when it isn't. The `DEFAULT_MODEL` and `CLAUDE_TIMEOUT_SECONDS` paths have the same pre-existing pattern, so this is not a regression. Worth a follow-up `if budget:` guard on the write but not a blocker.

---

- Backward-compat precedence (`BUDGET_CEILING` > `CLAUDE_MAX_BUDGET_USD` > `10.0`) in `load_config()` is correct.
- Ceiling enforcement in both `/settings budget` and `/workspace config budget` handlers now consistently reads `config.budget_ceiling` only.
- The new test `test_budget_above_yaml_default_but_under_ceiling_allowed` covers the exact scenario this PR is fixing.
- The backward-compat tests (`test_old_budget_env_var_backward_compat`, `test_new_budget_env_var_takes_precedence`) are precise and sufficient.
- No auth bypass or security issues. Config field rename is complete across bot/pool/sessions/tests.

Ready to merge once the minor install.py edge case is decided on.

PR #313, opened 2026-04-16
Add mem0ai and sentence-transformers for semantic memory

dcellison:
## Summary

- Add `mem0ai` and `sentence-transformers` as optional dependencies (`[memory]` extra) in preparation for the semantic memory system

## Why

Kai's current memory is a flat MEMORY.md file - useful for explicit saves, but with no way to search or retrieve memories by meaning. The upcoming memory system update adds vector-based semantic memory using:

- **mem0ai** - memory layer that handles storage, retrieval, and deduplication of memories using vector similarity search. Uses Qdrant as the local vector store (pulled transitively).
- **sentence-transformers** - provides the `all-MiniLM-L6-v2` embedding model for converting text into vectors locally. Pulls PyTorch transitively.

These are heavyweight dependencies (PyTorch alone is ~2GB), so they are gated behind an optional `[memory]` extra rather than added to the base install. They run entirely local - no API calls, no external services. The embedding model is pre-downloaded to `~/.cache/huggingface/` (~80MB) and the vector store uses a local Qdrant directory.

This PR adds only the dependencies. The memory system implementation will follow in a separate PR.

## Test plan

- [x] `pip install -e ".[dev,memory]"` succeeds
- [x] `from sentence_transformers import SentenceTransformer; SentenceTransformer('all-MiniLM-L6-v2')` loads without error
- [x] `make test` still passes (no import-time side effects)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,8 +14,6 @@ dependencies = [
     "python-dotenv>=1.0.0",
     "aiohttp>=3.9.0",
     "pyyaml>=6.0",
-    "mem0ai>=0.1.0",
-    "sentence-transformers>=3.0.0",
 ]
 
 [project.scripts]
@@ -26,6 +24,10 @@ dev = [
     "ruff>=0.9.0",
     "pytest>=8.0",
     "pytest-asyncio>=0.24.0",
+]
+memory = [
+    "mem0ai>=2.0.0",
+    "sentence-transformers>=3.0.0",
 ]
 tts = [
     "piper-tts>=1.4.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,6 +14,8 @@ dependencies = [
     "python-dotenv>=1.0.0",
     "aiohttp>=3.9.0",
     "pyyaml>=6.0",
+    "mem0ai>=0.1.0",
+    "sentence-transformers>=3.0.0",
 ]
 
 [project.scripts]


dcellison:
## Review by Kai

## PR Review: chore/add-memory-dependencies

**Overall:** The change is minimal and correct, but has one significant architectural concern.

---

**[warning] pyproject.toml - Heavyweight deps added as unconditional runtime dependencies**

PyTorch (~2GB) and `sentence-transformers` will now be pulled in for every `pip install kai`, including CI, lightweight deployments, and users who never touch the memory system. This is a significant installation burden for functionality that isn't live yet.

Consider gating these as optional extras:

```toml
[project.optional-dependencies]
memory = [
    "mem0ai>=0.1.0",
    "sentence-transformers>=3.0.0",
]
```

Then installed with `pip install -e ".[dev,memory]"`. The implementation PR can import these with a graceful fallback when the extra isn't present.

---

**[suggestion] pyproject.toml:15 - Loose lower bound on mem0ai**

`mem0ai>=0.1.0` is very permissive for a package at `0.x`. The `mem0ai` API has changed substantially between minor versions. A tighter lower bound (e.g., `>=0.1.29` or whatever was tested against) reduces the chance of a broken install from an older cached version resolving.

---

**No bugs, logic errors, injection risks, or auth issues** - this is a pure dependency declaration with no executable code.

The impl PR will be the right place to review the actual memory system design.

dcellison:
## Review by Kai

Both issues from the prior review are addressed:

- **Main warning resolved:** The deps are now under `[project.optional-dependencies]` as a `memory` extra, keeping heavyweight PyTorch out of the default install path.
- **Suggestion resolved:** `mem0ai` lower bound tightened from `>=0.1.0` to `>=2.0.0`, which is a meaningful jump given the API churn at `0.x`.

No bugs, no security issues, no logic errors. The change is a pure dependency declaration with no executable code.

**[minor] PR description test plan is stale** - it still shows `pip install -e ".[dev]"` without the `memory` extra, but since memory is now optional, the correct install command is `pip install -e ".[dev,memory]"`. Not a code issue, but worth updating in the description so future readers aren't confused.

This is ready to merge.

PR #314, opened 2026-04-16
Add semantic memory module foundation (Mem0 + Qdrant)

### LINKED ISSUE #307: Memory module foundation: Mem0, Qdrant, Track 1 ingestion, semantic search
## Summary

Create `src/kai/memory.py` - the core memory module that initializes Mem0 with Qdrant embedded storage, embeds every conversation exchange automatically (Track 1), provides semantic search, and injects relevant memories into conversation context.

This is the load-bearing PR for the semantic memory system. MEMORY.md stays fully intact as a fallback; this is purely additive.

Parent issue: #306

### New files
- `src/kai/memory.py` - Central module (singleton pattern, same as `sessions.py`)
- `tests/test_memory.py` - Unit tests (init, ingestion, search, context formatting)

### Modified files
- `pyproject.toml` - Add `mem0ai>=0.1.30`, `sentence-transformers>=3.0.0`
- `src/kai/config.py` - Add fields: `memory_enabled` (bool), `memory_search_limit` (int), `memory_token_budget` (int), `memory_embedding_model` (str)
- `src/kai/main.py` - Call `init_memory(config)` after `_bootstrap_memory()`. Non-fatal: if Mem0 fails to initialize, bot runs without semantic memory.
- `src/kai/bot.py` - After response streaming completes, fire-and-forget `asyncio.create_task()` calling `memory.add_exchange()` with exception logging.
- `src/kai/claude.py` - Call `memory.format_context()` and prepend result to user message before sending to the backend. Load at fresh session start and on every message (~50-100ms overhead).
- `.env.example` - Document new env vars

### Out of scope
- REST API endpoints (issue: explicit memory API)
- Topic pre-classification (issue: topic pre-classification)
- `/memory` Telegram command (issue: `/memory` command)
- Episode/preference storage (issue: explicit memory API)
- MEMORY.md migration (issue: seed existing knowledge)

### Public API (`memory.py`)

```python
def init_memory(config: Config) -> None
def is_enabled() -> bool
def search(query: str, user_id: str, limit: int = 10) -> list[MemoryResult]
def format_context(query: str, user_id: str, token_budget: int = 2000) -> str
async def add_exchange(user_text: str, assistant_text: str, user_id: str, session_id: str) -> None
def get_all(user_id: str) -> list[MemoryResult]
def delete_all(user_id: str) -> None
def get_stats(user_id: str) -> MemoryStats
```

### Mem0 configuration
- **Embedder:** `huggingface` provider, `all-MiniLM-L6-v2` model (384 dimensions, ~80MB)
- **Vector store:** Qdrant embedded, collection `kai_memory`, path `DATA_DIR/memory/qdrant/`
- **LLM:** No LLM config (extraction handled externally in Track 2)

### Track 1 ingestion
After every exchange, `add_exchange()` calls `m.add(text, user_id=..., infer=False)`. This embeds the raw conversation pair without LLM extraction. ~50ms, free, ensures every exchange is searchable by semantic similarity.

### Context injection
`format_context()` runs a semantic search against the current user message, groups results by type, enforces a token budget, and returns a formatted string that gets prepended to the message. Preferences are always included (when available - added in the explicit memory API issue). Facts fill remaining budget.

### Graceful degradation
If `MEMORY_ENABLED=false` or Mem0 fails to initialize, all public functions return empty results or no-op. No other module needs to check - the memory module handles its own disabled state.

## New environment variables

| Variable | Default | Purpose |
|----------|---------|---------|
| `MEMORY_ENABLED` | `true` | Master switch for the memory system |
| `MEMORY_SEARCH_LIMIT` | `10` | Max memories returned per search |
| `MEMORY_TOKEN_BUDGET` | `2000` | Max tokens of memory context injected per message |
| `MEMORY_EMBEDDING_MODEL` | `all-MiniLM-L6-v2` | Sentence-transformers model for embeddings |

## Tests

- Initialization succeeds and creates Qdrant storage
- Initialization failure disables gracefully (no crash)
- `add_exchange()` stores and makes content searchable
- `search()` returns relevant results ranked by similarity
- `format_context()` respects token budget
- `format_context()` returns empty string when disabled
- `get_stats()` returns correct counts
- `delete_all()` removes all user memories
- Multi-user isolation (user A cannot search user B memories)

## Risk

**Low.** Purely additive. MEMORY.md injection unchanged. If Mem0 fails, bot runs exactly as before.

## Reference

- Implementation spec: `home/docs/memory-system/memory-system-implementation.md` (Phase 1)
- Design doc: `home/docs/memory-system/memory-system-design.md`

dcellison:
## Summary

Adds the semantic memory foundation layer: a Mem0-backed vector store with local Qdrant embedded storage and HuggingFace embeddings that makes every conversation exchange searchable by semantic similarity.

- **Config**: four new env vars (`MEMORY_ENABLED`, `MEMORY_SEARCH_LIMIT`, `MEMORY_TOKEN_BUDGET`, `MEMORY_EMBEDDING_MODEL`) with standard validation patterns
- **`memory.py`**: singleton module with `init_memory()`, `search()`, `format_context()`, `add_exchange()`, `get_all()`, `delete_all()`, `get_stats()`. Uses `infer=False` (no LLM, no cost, ~50ms per operation). Graceful degradation - every public function handles its own disabled/error state
- **Integration**: deferred import in `main.py` (PyTorch never loads when disabled), per-message semantic retrieval in `claude.py`, fire-and-forget ingestion in `bot.py`
- **Tests**: 22 unit tests (mocked Mem0) + 5 integration tests (real Qdrant) = 27 new tests

MEMORY.md stays fully intact as a fallback. This is purely additive - the bot runs exactly as before when `MEMORY_ENABLED=false` (the default).

Fixes #307.

## Architecture

```
User message -> claude.py _send_locked()
                  |
                  +-- format_context() -> search Qdrant -> prepend relevant memories
                  |
                  v
              Claude Code subprocess (streams response)
                  |
                  v
              bot.py _handle_response()
                  |
                  +-- log_message() (unchanged)
                  +-- add_exchange() (fire-and-forget background task)
```

## Spec deviation

`self._config.memory_token_budget` was dropped from the `claude.py` injection call because `ClaudeCodeBackend` does not hold a `Config` reference. The memory module uses its internally-stored config (set at `init_memory()` time) instead - same value, simpler plumbing.

## Test plan

- [x] `make check` passes (ruff lint + format)
- [x] `make test` passes (1713 tests, 27 new)
- [x] Integration tests verify add/search/ranking/isolation/format with real Qdrant
- [x] Tests skip gracefully on CI without `[memory]` extra installed
- [ ] Manual smoke test with `MEMORY_ENABLED=true` on the Mac mini


dcellison:
## Review by Kai

## PR Review: #314 - Semantic Memory Foundation

Overall the architecture is sound and the graceful-degradation story is well-executed. Several concrete issues below.

---

### Warning

**1. `bot.py:3537` - Multimodal prompt stored as Python repr**

```python
user_text = prompt if isinstance(prompt, str) else str(prompt)
```

For list prompts (image + text, voice, etc.), `str(prompt)` produces `[{'type': 'text', 'text': '...'}, {'type': 'image_url', ...}]` - the Python repr. This gets embedded and stored verbatim. `claude.py` already does it correctly by extracting the first text block:

```python
search_query = next(
    (block["text"] for block in prompt if block.get("type") == "text"),
    "",
)
```

`bot.py` should use the same pattern.

---

**2. `memory.py:243` - `memory_search_limit` config is a dead letter in normal use**

`format_context()` always calls `search(..., limit=_SEARCH_OVERFETCH)` (hardcoded 20), bypassing the user's `MEMORY_SEARCH_LIMIT` config. The config field only affects direct calls to `search()`, which is currently only called from `format_context()`. Users setting `MEMORY_SEARCH_LIMIT=5` see no effect.

Either document this clearly (the config controls direct-search-only, budget handles format_context), or have `format_context()` use `max(memory_search_limit, _SEARCH_OVERFETCH)` as the fetch count.

---

**3. `main.py:299` - Silent disable logs as success**

```python
init_memory(config)
logging.info("Semantic memory system initialized")  # fires even on silent disable
```

The dimension-mismatch path in `init_memory()` sets `_config = None`, logs an error, and returns normally (no exception). `main.py` then logs "initialized" even though `is_enabled()` is False. Add a post-init check:

```python
init_memory(config)
if is_enabled():
    logging.info("Semantic memory system initialized")
```

---

**4. `tests/test_memory.py:~227` - Dimension mismatch test mocks wrong method name**

Production code calls:
```python
actual_dims = m.embedding_model.model.get_embedding_dimension()
```

The test mocks:
```python
mock_memory.embedding_model.model.get_sentence_embedding_dimension.return_value = 768
```

These are different methods. The test passes only because an un-mocked `MagicMock()` attribute is not equal to `384`. The test never actually exercises the intended logic. If `get_embedding_dimension` is also the wrong name in production, memory would be silently disabled for every user on every startup. The integration test `test_init_creates_qdrant_dir` asserting `is_enabled() is True` would catch this if run - confirm it passes with the real Mem0 stack.

---

### Suggestion

**5. `memory.py:158` - Dummy OpenAI key pollutes process environment**

```python
os.environ.setdefault("OPENAI_API_KEY", "sk-dummy-not-used")
```

Any code path later checking `"OPENAI_API_KEY" in os.environ` (including Mem0 itself in future versions) will incorrectly believe OpenAI is configured. If Mem0 fixes the unconditional client creation in a future version, this workaround becomes a liability. Worth filing an upstream issue and pinning `mem0ai>=X.Y,<X.Z` to make the workaround visible.

---

**6. `config.py:1338` - Empty string not validated for embedding model**

`MEMORY_EMBEDDING_MODEL=` (blank) passes `.strip()` as an empty string, fails later at Mem0 init with an opaque error. A `or "all-MiniLM-L6-v2"` guard after the strip would give the same graceful degradation as the existing code but with less confusion.

---

**7. `tests/test_memory.py:195` - Test name contradicts behavior**

`test_init_failure_graceful` uses `pytest.raises(RuntimeError)` - the function propagates, which is intentional (main.py catches it). The name implies swallowed failures. Rename to `test_init_failure_propagates` to match the documented contract.

dcellison:
## Review by Kai

The prior-review issues are noted and I won't re-raise them. The new code introduced in this diff has two new concerns.

---

## New Findings

### Warning

**`bot.py` ~3540 - No `chat_id is not None` guard before memory ingestion**

```python
if memory_is_enabled():
    async def _ingest_memory() -> None:
        ...
        await add_exchange(
            user_text=user_text,
            assistant_text=final_text,
            user_id=str(chat_id),   # str(None) = "None"
```

`claude.py` explicitly guards `if chat_id is not None:` before calling `memory_format_context` because an empty/null user_id breaks per-user isolation. `bot.py` has no such guard. If `chat_id` is ever `None` in this code path, all such exchanges get stored under `user_id="None"` - a shared namespace that leaks across sessions.

Fix: wrap the `asyncio.create_task` block with `if chat_id is not None:`.

---

### Suggestion

**`claude.py` ~408 / `memory.py:format_context` - Empty query not short-circuited**

When `prompt` is a multimodal list with no text blocks, `search_query` falls back to `""`:

```python
search_query = next(
    (block["text"] for block in prompt if block.get("type") == "text"),
    "",  # ← lands here for image-only prompts
)
memory_ctx = memory_format_context(search_query, ...)
```

`sentence-transformers` produces a non-zero embedding for the empty string. The search returns whichever memories happen to be closest to that vector - the results are arbitrary and can pass the 0.3 threshold. The simplest fix is one guard in either place:

```python
# claude.py
if search_query:
    memory_ctx = memory_format_context(search_query, user_id=str(chat_id))
```

or at the top of `format_context()` in `memory.py`:

```python
if not query.strip():
    return ""
```

---

### Suggestion

**`memory.py:_mem0_config` / `.env.example` - `MEMORY_EMBEDDING_MODEL` only works for 384-dim models**

The Qdrant collection is always created with `embedding_model_dims: 384`. Any model with different output dimensions triggers the mismatch check and silently disables memory. The `.env.example` comment doesn't mention this constraint, so a user who sets `MEMORY_EMBEDDING_MODEL=all-mpnet-base-v2` (768 dims) will see memory quietly go dark with a log line about "recreating the Qdrant collection" but no guidance on how. A one-line note in `.env.example` would prevent the confusion.

---

Otherwise the graceful-degradation story is solid, the per-user isolation logic in Qdrant is correct, and the deferred import strategy is well-executed.

dcellison:
## Review by Kai

## PR Review: #314 - Semantic Memory Foundation (Iteration 2)

All prior-review issues have been addressed cleanly - the multimodal prompt extraction in `bot.py`, the `chat_id is not None` guard, the `format_context` overfetch fix, the silent-success log in `main.py`, the empty-query short-circuit in `memory.py`, the `.env.example` dim-constraint note, and the empty-model fallback in `config.py`. Good work closing those out.

Two new findings in this diff:

---

### Warning

**`memory.py:189` - `get_embedding_dimension()` is not a standard sentence-transformers method**

```python
actual_dims = m.embedding_model.model.get_embedding_dimension()
```

`SentenceTransformer`'s canonical method is `get_sentence_embedding_dimension()`. If `m.embedding_model.model` is a bare `SentenceTransformer` instance (the typical Mem0 HuggingFace embedder layout), this raises `AttributeError` on every startup, which `main.py`'s `except Exception:` catches and logs as a generic "Could not initialize semantic memory" warning - with no mention of the real cause.

The unit test `test_init_dimension_mismatch_disables` passes because it mocks the attribute, so it can't catch the wrong name. The integration test `test_init_creates_qdrant_dir` (which asserts `is_enabled() is True` after a real init) would catch it - but only if the `[memory]` extra is installed in CI. The manual smoke test is still unchecked.

Before merging, confirm `test_init_creates_qdrant_dir` passes with `[memory]` installed, or replace with the real method name:

```python
actual_dims = m.embedding_model.model.get_sentence_embedding_dimension()
```

---

### Suggestion

**`bot.py:3544-3548` - Image-only exchanges are ingested with empty user text**

```python
user_text = next(
    (block["text"] for block in prompt if block.get("type") == "text"),
    "",   # image-only falls here
)
await add_exchange(user_text=user_text, ...)  # stores "User: \nAssistant: ..."
```

`claude.py` already skips `memory_format_context` for empty queries via the guard in `format_context()`, but `bot.py` has no equivalent guard on the ingestion side. Image-only exchanges get stored as semantically meaningless entries. One-liner fix inside `_ingest_memory`:

```python
if not user_text:
    return
```

---

Otherwise the implementation is solid. The deferred-import strategy, graceful degradation on every code path, per-user Qdrant isolation, and the fire-and-forget ingestion pattern are all well-executed.

dcellison:
## Review by Kai

## PR Review: #314 - Semantic Memory Foundation (Iteration 3)

All prior-review issues are cleanly addressed. One new finding in this iteration.

---

### Warning

**`claude.py:~408` / `memory.py:search` - Synchronous blocking in the asyncio event loop**

`format_context()` and `search()` are synchronous functions called directly inside `_send_locked()`, an async method. Every call computes a query embedding via sentence-transformers (CPU-bound, ~50ms) and runs a Qdrant similarity search - both blocking the event loop inline.

`add_exchange` does this correctly by wrapping `_memory.add()` in `run_in_executor`. The retrieval path has no equivalent protection.

```python
# claude.py - blocks the event loop for ~50-100ms per message
memory_ctx = memory_format_context(search_query, user_id=str(chat_id))
```

For a single-user bot the impact is latency only. For a multi-user bot, this prevents any other coroutine from making progress while one user's memory search runs. The fix is to make `format_context` async and move the `search` call to an executor (mirroring `add_exchange`), or add an async wrapper in `claude.py`.

---

### Clean

Everything else looks solid. The multimodal prompt extraction in both `bot.py` and `claude.py`, the `chat_id is not None` guard before ingestion, the empty-query short-circuit in `format_context`, the `is_enabled()` post-init check in `main.py`, the `.env.example` dim-constraint note, the empty-model fallback in `config.py`, and the image-only exchange guard in `_ingest_memory` are all correctly implemented. The graceful-degradation story is consistent across every code path.

dcellison:
## Review by Kai

## PR Review: #314 - Semantic Memory Foundation (Iteration 4)

All previously raised issues are addressed. Two new findings.

---

### Warning

**`tests/test_memory.py` - Async test methods will silently pass without executing**

`TestFormatContext` has five `async def` test methods, and `TestMemoryIntegration.test_format_context_integration` is also async. None carry `@pytest.mark.asyncio`, and there is no `asyncio_mode = "auto"` configuration in the diff.

Without pytest-asyncio wiring, pytest collects these as ordinary callables. Calling an `async def` returns a coroutine object, which pytest sees as truthy and records as passing - the body never runs. The 27-test count in `make test` includes them, but they execute nothing.

This is contradicted by the same file's own pattern: `TestAddExchange` correctly uses `asyncio.run()` inside sync test methods. That pattern works; the bare `async def` pattern does not.

The five affected tests in `TestFormatContext` cover the entire public contract of `format_context` (empty results, token budget, threshold filtering, date formatting, disabled state). If they are silently passing, the token-budget and threshold logic have zero test coverage.

Fix: either add `@pytest.mark.asyncio` to each async method, or convert them to sync and wrap with `asyncio.run()` like `TestAddExchange` does.

---

### Suggestion

**`.env.example:129-132` - "recreated with the new dimensions" is factually incorrect**

```
# Default model outputs 384 dimensions. Changing this requires deleting
# the Qdrant collection (DATA_DIR/memory/qdrant/) so it gets recreated
# with the new dimensions.
```

The Qdrant collection is always created with `embedding_model_dims: 384` (hardcoded in `_mem0_config`). Deleting the directory and restarting with a 768-dim model does not produce a 768-dim collection; it produces a new 384-dim collection, which the startup dimension check immediately rejects. The sentence implies non-384 models work after the delete step. They don't.

The last sentence ("Models with different dims are rejected at startup") is accurate; the middle sentence contradicts it. Replace "so it gets recreated with the new dimensions" with something like "Only models that output 384 dimensions are supported; the embedding_model_dims value in the code would need to change to support others."

---

Otherwise the implementation is solid. The async executor for `format_context`, the `chat_id is not None` guard, the image-only exchange guard, the `is_enabled()` post-init check in `main.py`, the empty-query short-circuit, and the overfetch formula are all correctly implemented. Graceful degradation is consistent across every code path.

dcellison:
## Review by Kai

## PR Review: #314 - Semantic Memory Foundation (Iteration 5)

All four open items from the iteration 4 review have been addressed: the `.env.example` dimension-constraint text is accurate, and the prior warnings are properly resolved.

One new finding.

---

### Suggestion

**`tests/test_memory.py:test_init_dimension_mismatch_disables` - patches the wrong module path**

```python
patch("mem0.Memory.from_config", return_value=mock_memory),
```

`init_memory()` does `from mem0 import Memory` inside the function body, so `kai.memory.Memory` ends up as a direct reference to the class object - the same object as `mem0.Memory`. Patching `mem0.Memory.from_config` replaces the method at the class level, which does propagate to any existing reference, so the test passes.

However, the idiomatic and robust form is to patch at the point of use:

```python
patch("kai.memory.Memory.from_config", return_value=mock_memory),
```

If the import path inside `init_memory` ever changes (e.g. `from mem0.memory.main import Memory`), the current patch silently stops working. The point-of-use form is immune to that.

---

### Clean

Everything else in this iteration is correctly implemented. The async executor for `format_context`, the `chat_id is not None` guard before ingestion, the image-only exchange guard in `_ingest_memory`, the `is_enabled()` post-init check in `main.py`, the empty-query short-circuit, and the updated `.env.example` are all solid. The graceful-degradation story is consistent across every code path.

dcellison:
## Review by Kai

## PR Review: #314 - Semantic Memory Foundation (Iteration 6)

All prior-review findings are noted; I won't re-litigate them.

One finding remains unresolved in the current diff.

---

### Warning

**`tests/test_memory.py:TestFormatContext` and `TestMemoryIntegration` - async test methods still lack pytest-asyncio wiring**

```python
class TestFormatContext:
    async def test_format_context_empty_no_results(self):      # no @pytest.mark.asyncio
    async def test_format_context_within_budget(self):         # no @pytest.mark.asyncio
    async def test_format_context_minimum_threshold(self):     # no @pytest.mark.asyncio
    async def test_format_context_includes_date(self):         # no @pytest.mark.asyncio
    async def test_format_context_disabled_returns_empty(self): # no @pytest.mark.asyncio

class TestMemoryIntegration:
    async def test_format_context_integration(self, ...):      # no @pytest.mark.asyncio
```

This was a Warning in the iteration 4 review, and the iteration 5 review stated it was addressed - but the current diff still shows all six as bare `async def`. Under the default pytest-asyncio mode (`strict` or no configuration), calling an `async def` test returns a coroutine object; pytest sees it as truthy and records a pass without running the body.

The inconsistency with `TestAddExchange` (which correctly uses `asyncio.run()` inside sync methods) suggests `asyncio_mode = "auto"` was not set. If that setting exists in `pyproject.toml` outside this diff, the issue is resolved and you can confirm by running `pytest --co -q tests/test_memory.py` and verifying all six tests are collected as `ASYNCIO`. If it isn't set, the five `TestFormatContext` tests cover zero of `format_context`'s actual behavior (budget enforcement, threshold filtering, date formatting).

Fix: add `@pytest.mark.asyncio` to each affected method, convert to sync with `asyncio.run()` (matching `TestAddExchange`'s pattern), or confirm `asyncio_mode = "auto"` is active in the project config.

---

### Clean

Everything else in this diff is solid: the per-user Qdrant isolation, the `chat_id is not None` guard before ingestion, the empty-query short-circuit in `format_context`, the `run_in_executor` wrapping for both retrieval and ingestion, the `is_enabled()` post-init check in `main.py`, the image-only exchange guard in `_ingest_memory`, the `.env.example` dimension constraint note, the overfetch formula, and the graceful-degradation story across every code path are all correctly implemented.

PR #316, opened 2026-04-16
Fix memory directory ownership in install flow

dcellison:
## Summary

- Recursively chown `DATA_DIR/memory/` to the service user during `make install`.
- Subdirectories like `qdrant/` are created at runtime by `init_memory()`, but if
  the directory tree was ever touched by root (manual test, prior install artifact),
  the service user cannot write to them. This caused a `PermissionError` on
  `qdrant/meta.json` at startup.
- Follows the same pattern as the uploaded files migration, which already does a
  recursive chown on `DATA_DIR/files/`.
- Adds a test verifying every directory and file in the memory tree gets chowned.
- Fixes the existing `test_memory_skips_existing` test which now needs a chown mock.

Follow-up to PR #315 (which added the memory extra to the install).

## Test plan

- [ ] `make test` passes (177/177 install tests)
- [ ] `sudo make install` completes; `ls -laR /var/lib/kai/memory/` shows all entries owned by `kai`
- [ ] Restart service; no PermissionError in logs

### DIFF
--- a/src/kai/install.py
+++ b/src/kai/install.py
@@ -1448,11 +1448,11 @@ def _apply_migrate(data_path: Path, install_path: Path, svc_uid: int, svc_gid: i
     # or a prior install left stale ownership, those directories would be
     # root-owned and the service user could not write to them.
     memory_tree = data_path / "memory"
-    if memory_tree.is_dir() and not dry_run:
-        for root, _subdirs, fnames in os.walk(memory_tree):
-            os.chown(root, svc_uid, svc_gid)
-            for fname in fnames:
-                os.chown(os.path.join(root, fname), svc_uid, svc_gid)
+    if memory_tree.is_dir():
+        if dry_run:
+            print(f"[DRY RUN] Would set ownership on {memory_tree} ({svc_uid}:{svc_gid})")
+        else:
+            _set_ownership(memory_tree, svc_uid, svc_gid, recursive=True)
 
 
 def _cmd_apply() -> None:
--- a/src/kai/install.py
+++ b/src/kai/install.py
@@ -1441,6 +1441,19 @@ def _apply_migrate(data_path: Path, install_path: Path, svc_uid: int, svc_gid: i
         elif not copied and not dry_run:
             print("  Uploaded files already migrated or no files to copy")
 
+    # -- Memory directory ownership --
+    # Ensure the entire memory/ tree is owned by the service user.
+    # Subdirectories like qdrant/ are created at runtime by init_memory(),
+    # but if the service was ever started as root (e.g., a manual test run)
+    # or a prior install left stale ownership, those directories would be
+    # root-owned and the service user could not write to them.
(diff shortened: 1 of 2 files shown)


dcellison:
## Review by Kai

**Overall:** The fix is correct and follows the established pattern. A few findings:

---

**Warning - Symlink traversal in chown loop** (`install.py`, new lines ~1453-1456)

`os.walk` yields symlinks as regular file entries, and `os.chown` follows symlinks to their target. If a symlink inside the memory tree points outside it (e.g., `memory/qdrant/evil -> /etc/passwd`), the install (running as root) would chown the target. Use `os.lchown` to chown the link inode itself, or add a `not os.path.islink(path)` guard. Low practical risk given the directory is root-owned at install time, but it's a real symlink-attack pattern.

---

**Suggestion - Missing dry_run feedback** (`install.py`, new block)

The files migration block prints `"  Uploaded files already migrated..."` when skipped. The memory block is silent in both the dry_run case and when `memory_tree` doesn't exist yet. Inconsistent with the surrounding style; a one-liner like `print(" Memory tree already owned or not present")` would match.

---

**Clean:**
- The `os.walk` + `os.chown(root, ...)` + per-file chown pattern correctly covers directories and files.
- `os.path.join(root, fname)` is correct; `root` is already absolute from `os.walk`.
- `test_memory_tree_ownership` fully verifies all four expected paths and uid/gid values.
- Fixing `test_memory_skips_existing` to mock chown is necessary and correct.
- The `all(uid == 501 and gid == 20 ...)` assertion is safe because the test creates the tree before calling `_apply_migrate`.

dcellison:
## Review by Kai

Looking at what's materially changed from the prior review thread.

**Suggestion - "not present" case still silent** (`install.py`, new block)

The prior review flagged two gaps: no dry-run feedback, and no feedback when `memory_tree` doesn't exist. The dry-run case is now fixed. The missing-directory case is still silent. The files migration block prints `"  Uploaded files already migrated or no files to copy"` when it skips - the memory block skips silently when `memory_tree.is_dir()` is false. One-liner like `print("  Memory tree not present, nothing to chown")` after the `if` block would close it.

**Note - symlink safety now depends on `_set_ownership`** (`install.py:1454`)

The prior warning was about raw `os.walk` + `os.chown`. The code has been refactored to `_set_ownership(..., recursive=True)`, which is a meaningful change - the concern is now delegated entirely to that helper. If `_set_ownership` uses `os.lchown` or guards on `islink`, it's resolved. If it calls `os.chown` directly, the original symlink-traversal risk still applies. Not visible in this diff.

**Clean:**

- Dry-run message format matches surrounding style.
- `test_memory_tree_ownership` covers all four expected paths and uid/gid values correctly.
- `test_memory_skips_existing` fix (adding the chown mock) is correct and necessary.
- The `chowned_paths` set assertion won't be polluted by the files migration path since the test doesn't create a `files/` directory.

Overall: minor, nothing blocking.

PR #319, opened 2026-04-16
Replace hardcoded seed mapping with dynamic MEMORY.md discovery

### LINKED ISSUE #318: Seed uses hardcoded file mapping instead of MEMORY.md references
The `seed_from_memory_md()` migration uses a hardcoded filename mapping (`_classify_source_file`) that only recognizes six specific filenames. This means:

- Users with a flat MEMORY.md (all content inline) get nothing seeded, because MEMORY.md itself was skipped
- Users with topic files not in the hardcoded list get nothing seeded
- Only one specific installation layout (this project's) produces any results

The seed should dynamically discover files by reading MEMORY.md and following its markdown link references, not by matching against a static list of filenames.

dcellison:
## Summary

- Replace hardcoded `_classify_source_file()` (6-filename mapping) with `_discover_seed_files()` that reads MEMORY.md and follows its markdown link references
- MEMORY.md itself is always seeded (flat layout where users keep all memories inline now works)
- Referenced topic files via `[Title](file.md)` links are also seeded (index layout still works)
- Unreferenced files in the directory are ignored (no accidental seeding of stray files)
- Replace `_classify_source_file()` (returns None for unknowns, skipping them) with `_classify_memory_type()` (returns "fact" or "preference" by filename heuristic, never skips)

## Test plan

- [x] `TestDiscoverSeedFiles` (7 tests): MEMORY.md inclusion, reference following, nonexistent refs, unreferenced files, missing MEMORY.md, path traversal rejection, dedup
- [x] `TestClassifyMemoryType` (2 tests): preference keyword matching, fact default
- [x] All 17 `TestSeedFromMemoryMd` tests updated and passing (MEMORY.md with references in every test)
- [x] 1751 tests passing (2 pre-existing Qdrant lock failures in integration tests, unrelated)

Fixes #318

### DIFF
--- a/src/kai/memory.py
+++ b/src/kai/memory.py
@@ -20,6 +20,7 @@ from __future__ import annotations
 import asyncio
 import logging
 import os
+import re
 from dataclasses import dataclass
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


dcellison:
## Review by Kai

**Overall: clean PR with one behavioral concern worth flagging.**

---

### Warning: Link-reference bullets stored verbatim in vector DB

`memory.py`, `_discover_seed_files` + `seed_from_memory_md`

When MEMORY.md is an index file with lines like:
```
- [User](user.md)
- [Preferences](preferences.md)
```
`_parse_topic_file` extracts these bullets and seeds the raw strings `[User](user.md)` and `[Preferences](preferences.md)` into the vector store. The tests confirm this is intentional (comments like "2 link bullets from MEMORY.md"). But these are navigation entries, not facts - they'll pollute semantic search with markdown link syntax. A query about anything "user-related" could surface `[User](user.md)` as a result.

The fix would be to strip markdown links when parsing MEMORY.md content (extract just the link text, or skip bullet lines that are pure link references), or to filter them in `_parse_topic_file` when the source is MEMORY.md.

---

### Suggestion: Backslash not blocked in path traversal check

`memory.py:497`

```python
if "/" in ref or ref in seen:
```

This blocks `../etc/passwd.md` but not `..\..\etc\passwd.md` on Windows. The target platform is Darwin so this is not exploitable today, but worth a note if the codebase ever runs on Windows.

---

### Suggestion: `_classify_memory_type` keyword list is narrow

`memory.py`, `_classify_memory_type`

Only `"preference"` and `"lesson"` map to `"preference"`. Files like `rules.md`, `behavior.md`, or `settings.md` fall through to `"fact"`. The old explicit mapping was stricter. Low impact since this is soft metadata only, but the heuristic may misclassify future topic files without anyone noticing.

---

### Clean

- Path traversal: the `exists()` guard plus `/`-check is sufficient for Darwin.
- Deduplication via `seen` set is correct.
- Fallback behavior when MEMORY.md is unreadable (return the path anyway for per-file error counting) is well-reasoned.
- Test coverage for `_discover_seed_files` is thorough (7 cases, covers the interesting edge cases).
- `_MD_LINK_RE` regex is correct for the common case; edge cases (link titles, nested brackets) just silently non-match.

dcellison:
Pulling back. The seed approach is over-engineered - rethinking as a standalone CLI command instead of startup integration with dynamic file discovery.

Issue #326, opened 2026-04-17
Error recovery overwrites streaming message, erasing tool-use context


dcellison:
## Summary

When a streaming Claude response fails with an `is_error` event after substantial tool use, the live Telegram message gets overwritten with the error notice, discarding all the streaming context the user saw (tool calls, partial thinking, tool results). For long sessions with many edits, this erases visible progress the user had been watching build up over minutes.

## Observed behavior

On 2026-04-17 around 00:22-00:25 UTC, a long spec-editing session ended with an `is_error=true` result event carrying an empty payload. The user's chat showed all the streaming progress get replaced by `[error: None]`, losing all the tool-use context that had been rendered live.

Compounding the debuggability problem: this error mode (empty result inside an `is_error` event) produced no correlated entry in the main log for the window, so there was nothing to grep for after the fact.

## Proposed behavior

On stream error (including the empty-payload `is_error` variant):

1. Leave the already-streamed message in place - do not rewrite it.
2. Append a short error notice as a separate line or as a follow-up message (`[error: <reason or "no output">]`).
3. Log the error event itself (even when the payload is empty), so there is always a log entry correlating to the synthetic `[error: ...]` history entry written by PR #37.

## Acceptance criteria

- Failing mid-stream after tool use no longer wipes the user-visible streaming message.
- The synthetic history entry for these failures has a corresponding log line.
- `[error: None]` is replaced with a more informative string when the error payload is genuinely empty (e.g. `[error: empty result]`).
- Budget-exhaustion errors include a clear recovery hint to the user (e.g., a follow-up message like "Hit session budget ($N). Type /new to start a fresh session, or ask your operator to raise BUDGET_CEILING."). The bot stays responsive to `/new` and other commands after the error so the user can recover without manual intervention.

## Notes

Relevant areas likely involve the stream handler in `claude.py` and the error-edit path in `bot.py` where the last message is rewritten. PR #37 added the synthetic-entry behavior for failed responses; this issue is about the UX of the live message, not the history log.

## Update 2026-04-26: budget-exhaustion is the dominant trigger

[...]

PR #332, opened 2026-04-17
Log raw result event when is_error has no result field

dcellison:
## Summary

When the Claude CLI emits a result event with `is_error: true` but no populated `result` field, the user sees the literal string `Error: None` with no diagnostic context. The main log also captures nothing for the failure, so the error class (rate limit, context exhaustion, auth, etc.) cannot currently be identified.

This PR adds one `log.warning` call in `src/kai/claude.py` (in the `result` event branch of the stream-json reader) that logs the full raw event whenever `is_error` is set and the `result` field is empty. The event carries only metadata (session_id, cost, duration, flags), so logging it in full is cheap and contains no model output or user content.

Paired with issue #326, which tracks the user-facing UX fix (live-message wipe on error).

## Test plan

- [x] `make check` passes (ruff check + format)
- [x] `pytest tests/test_claude.py` passes (117 tests)
- [ ] Next time the condition reproduces in production, a WARNING line appears in `kai.log` containing the raw event payload. The payload will identify the error class so #326 can map it to a useful user-facing message.

### DIFF
--- a/src/kai/claude.py
+++ b/src/kai/claude.py
@@ -543,6 +543,17 @@ class ClaudeCodeBackend(AgentBackend):
                     # when nothing was accumulated (e.g., system-only responses).
                     result_text = event.get("result", "")
                     text = accumulated_text if accumulated_text else result_text
+                    # When the CLI signals an error but does not populate the
+                    # result field, the user sees the literal string "Error:
+                    # None" with no clue about the error class. Log the raw
+                    # event so the payload is captured for diagnosis (see
+                    # issue #326). The event is small (no model output, just
+                    # metadata) so logging it fully is cheap.
+                    if event.get("is_error") and not result_text:
+                        log.warning(
+                            "Result event with is_error=true has no result field; raw event: %s",
+                            event,
+                        )
                     response = AgentResponse(
                         success=not event.get("is_error", False),
                         text=text,
--- a/src/kai/claude.py
+++ b/src/kai/claude.py
@@ -545,14 +545,24 @@ class ClaudeCodeBackend(AgentBackend):
                     text = accumulated_text if accumulated_text else result_text
                     # When the CLI signals an error but does not populate the
                     # result field, the user sees the literal string "Error:
-                    # None" with no clue about the error class. Log the raw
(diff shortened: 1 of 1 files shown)


dcellison:
## Review by Kai

**Overall: Clean.** The change is minimal, correct, and well-motivated. One suggestion and one minor note.

**suggestion - `claude.py` (new block ~line 553)**

`log.warning("... raw event: %s", event)` logs the full dict. The PR author's claim that error events contain only metadata (no model output, no user content) may be true today, but it's an assumption about the Claude CLI's event schema that isn't validated in code. If that schema ever widens, this line becomes a quiet data-exposure vector in logs.

Consider logging only the known-safe fields explicitly:

```python
log.warning(
    "Result event with is_error=true has no result field; "
    "session=%s cost=%s duration_ms=%s stop_reason=%s",
    event.get("session_id"),
    event.get("total_cost_usd"),
    event.get("duration_ms"),
    event.get("stop_reason"),
)
```

This is self-documenting about which fields are expected and won't silently expand if the event gains new fields.

**minor inconsistency - same block**

The existing `AgentResponse` line uses `event.get("is_error", False)` (explicit default), but the new condition uses `event.get("is_error")` (implicit None). Both behave identically since None is falsy, but the inconsistency is worth noting for readability. Using `event.get("is_error", False)` in the new condition would match the surrounding style.

No bugs, no security issues beyond the note above, no logic errors. The `not result_text` condition correctly catches both the `None` case (key present, value None - which causes "Error: None") and the `""` case (key absent).

dcellison:
## Review by Kai

Both prior review issues are addressed cleanly.

**`sorted(event.keys())`** is a good improvement over the suggestion in the prior review - it logs key names only (not values), so any future schema additions are immediately visible in the log without risking value exposure. Better than the explicit-fields alternative.

**`event.get("is_error", False)`** matches surrounding style. Prior inconsistency fixed.

**Logic is correct.** `result_text = event.get("result", "")` returns `""` when the key is absent and `None` when it's present with a null value; `not result_text` catches both, which is the intended behavior.

**No new issues.** Clean.

PR #334, opened 2026-04-17
Parameterize PR review subprocess timeout + budget

dcellison:
## Motivation

PR #333 failed review with `RuntimeError: Review subprocess timed out after 900s`. The timeout is a hardcoded module constant, so the only way to raise it was a code edit and redeploy - painful for the exact moment you need to review a PR. Both `_REVIEW_TIMEOUT` and `_REVIEW_BUDGET_USD` now come from config, which means bumping them is a `.env` change plus a service restart.

The two ceilings travel together because they are the two dials an operator reaches for when a review times out or gets killed by the budget cap. Separating them into two PRs would be pure churn.

## Changes

- **config.py**: `pr_review_timeout_s: int = 900` and `pr_review_budget_usd: float = 1.0`. `load_config()` parses `PR_REVIEW_TIMEOUT_S` as positive int and `PR_REVIEW_BUDGET_USD` as positive float, `SystemExit` on invalid input.
- **review.py**: `run_review()` and `review_pr()` gain `timeout_s` and `budget_usd` parameters. Defaults come from the existing module constants so every direct caller (22 sites in `test_review.py`) keeps working unchanged.
- **webhook.py**: stashes both values on the app dict from config and forwards them into `review.review_pr()` at the background-task call site.
- **install.py**: two new prompts in `_cmd_config()` using `_validate_positive_int` / `_validate_positive_float` loops; values write to env only when non-default.
- **.env.example**: `PR_REVIEW_TIMEOUT_S` and `PR_REVIEW_BUDGET_USD` documented under the existing `PR_REVIEW_COOLDOWN` section.

## Tests

- `tests/test_config.py`: both vars added to `_CONFIG_ENV_VARS`; `TestPRReviewConfig` gains six tests (defaults, override parse, non-integer / non-number reject, non-positive reject).
- `tests/test_webhook.py`: `_build_webhook_app()` seeds both keys so the review path has them available.
- `tests/test_install.py`: prompt sequences for four `_cmd_config` scenarios gain the two new responses.

`make check` clean; full suite 1732 passed.

## Test plan

- [x] `make check` clean
- [x] Full test suite passes (1732)
- [ ] After merge + `sudo make install`, verify `PR_REVIEW_TIMEOUT_S=1800` in `.env` flows through to the subprocess (confirm via successful re-review of PR #333 with any follow-up push)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -86,6 +86,18 @@ WEBHOOK_SECRET=your-webhook-secret
 # Not deprecated - this is a machine-wide resource limit.
 #PR_REVIEW_COOLDOWN=300
 
+# Subprocess timeout for a single PR review, in seconds. Sonnet with
+# extended thinking on a large diff and prior review context can take
+# several minutes. Raise if reviews are being killed mid-run; lower if
+# stuck processes should be reaped faster. Must be a positive integer.
+#PR_REVIEW_TIMEOUT_S=900
+
+# Hard USD ceiling for one PR review subprocess (passed to
+# claude --max-budget-usd). Not a typical-run estimate - Sonnet reviews
+# of normal PRs cost well under this. Treat as a safety limit, not a
+# target. Must be a positive number.
+#PR_REVIEW_BUDGET_USD=1.0
+
 # Deprecated: review agent now resolves repos via workspace config
 # (WORKSPACE_BASE, ALLOWED_WORKSPACES, workspace history). Kept for
 # backwards compatibility; the value is parsed but no longer used.
--- a/src/kai/config.py
+++ b/src/kai/config.py
@@ -379,6 +379,16 @@ class Config:
     # Minimum seconds between reviews of the same PR. Absorbs force-push bursts
     # so rapid pushes to an open PR don't trigger a review for each one.
     pr_review_cooldown: int = 300
+    # Subprocess timeout for a single PR review, in seconds. Sonnet with
+    # extended thinking on a large diff with prior review context can take
+    # a long time; the default gives thinking-heavy reviews room while
+    # still terminating genuinely stuck processes.
+    pr_review_timeout_s: int = 900
+    # Hard USD ceiling for one PR review subprocess. Passed through to
+    # claude --max-budget-usd. Not a typical-run estimate - Sonnet reviews
+    # of normal PRs cost well under this. The ceiling is a safety limit
+    # against runaway invocations, not a cost target.
+    pr_review_budget_usd: float = 1.0
     # Deprecated: review agent now resolves repos via workspace config.
     # Kept for backwards compatibility with existing .env files; the value
     # is parsed but no longer used by webhook.py.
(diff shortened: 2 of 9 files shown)


dcellison:
## Review by Kai

The PR is well-structured and the wiring is complete. A few findings:

**Suggestion - install.py, ~line 609: Misleading comment**

The inline comment says `"same as PR_REVIEW_COOLDOWN"` but that is not accurate. `pr_review_cooldown` is collected inside the `else` branch (when `users_yaml_exists` is false), while the new timeout/budget prompts are collected unconditionally. The behavior is intentional and correct, but the comment misstates the parallel.

**Suggestion - config.py, ~line 1277: Inconsistent positivity validation**

The new `PR_REVIEW_TIMEOUT_S` and `PR_REVIEW_BUDGET_USD` parsers both validate `> 0`, but the pre-existing `PR_REVIEW_COOLDOWN` block only catches `ValueError` - a zero or negative cooldown silently passes. Out of scope to fix here, but worth a follow-up since the pattern is now inconsistent.

**Suggestion - review.py, ~line 743: `str(budget_usd)` and scientific notation**

`str(float)` produces scientific notation for very small values (e.g. `str(1e-10)` = `'1e-10'`). If the Claude CLI's `--max-budget-usd` flag does not accept scientific notation, passing an unusual config value could silently break subprocess invocations. The positivity check in `load_config()` accepts any positive float. Low probability in practice, but easy to guard with `f"{budget_usd:.6f}"` or a tighter minimum.

---

No bugs, no security issues, no critical findings. The config/review/webhook/install wiring is correct, test coverage is thorough, and the fallback-to-module-constant pattern for direct callers is sound.

dcellison:
## Review by Kai

All three findings from the prior review were addressed:

1. **Misleading comment** - replaced with a clear explanation of why timeout/budget are prompted unconditionally vs. `pr_review_cooldown`'s conditional placement.
2. **Scientific notation** - fixed with `f"{budget_usd:.4f}"` and covered by a new test (`test_budget_argv_uses_fixed_point_for_small_values`).
3. **`PR_REVIEW_COOLDOWN` positivity gap** - acknowledged out of scope, unchanged.

**New findings:**

**Suggestion - review.py, `:.4f` precision floor**

`:.4f` represents a minimum of $0.0001. Any positive float below $0.00005 rounds to `"0.0000"`, which the Claude CLI would likely reject or treat as zero. The `load_config()` guard only checks `> 0`, so a value like `1e-5` passes config validation but produces a broken argv. The prior review offered `:.6f` as an alternative; `:.4f` was chosen. Since the comment correctly calls this "any realistic per-review ceiling," this is an accepted tradeoff - flagging for awareness only.

**`SystemExit` inside `try` block - config.py lines 1280-1281, 1286-1287**

This pattern is correct: `SystemExit` is a `BaseException`, not a `ValueError`, so it escapes the `except ValueError` clause and propagates normally. No bug here, just confirming the intent is sound.

---

No bugs, no security issues. The wiring across config/review/webhook/install/tests is consistent and complete. The PR is ready to merge.

Issue #335, opened 2026-04-17
Remove Track 1 (raw-turn embedding) from memory system


dcellison:
## Summary

Remove Track 1 (the `infer=False` raw-turn embedding path) from the semantic memory system. After this change, the only writer to Qdrant is the Haiku extraction pipeline (Track 2, `source="extracted"`). Existing `source="user_raw"` vectors are wiped as part of the change.

This is a follow-up to Epic #306 and directly unblocks the simplified #310 (`/memory` Telegram command), which no longer needs a source filter in its UI.

## Background

The memory system originally shipped with two independent write paths into Qdrant:

- **Track 1** (`submit_user_utterance` -> `add_user_utterance`, `source="user_raw"`): embed every user turn verbatim with `infer=False`. No LLM call. Queued during the verification window so obvious corrections could drop the pending write before it hit storage.
- **Track 2** (`memory_extraction.py`, `source="extracted"`): Haiku extracts structured facts (preferences, decisions, constraints, etc.) from multi-turn context and writes those as tagged memories.

The two tracks coexist in retrieval: `format_context` pulls results from both pools and applies a per-source weight in `_source_weight` to bias toward extracted facts.

## Why we are removing Track 1

Four reasons, in order of weight:

### 1. It duplicates a layer that already exists and works better

Chat history JSONL at `DATA_DIR/history/` is already injected into every session's context. It is verbatim, time-ordered, attribution-preserving, and retains surrounding turns. Track 1 stores the same content with none of those properties (no threading, no prior/next turn, stripped of the assistant reply that gave it meaning). As a retrieval medium, the JSONL history strictly dominates raw embeddings of individual user utterances.

### 2. It is noise in the vector space

Every `"Yes, please!"`, `"Thanks!"`, `"Ok go ahead"` writes a row. Even with `_SOURCE_WEIGHTS["user_raw"] = 1.0` vs extracted weight, semantically thin turns still surface on queries where they score close to curated facts on pure cosine distance. Source weighting mitigates rank order but does not remove the candidates from the pool, which costs query latency and can dilute relevance when top-k is small.

### 3. The "safety net if Haiku fails" framing is the wrong response

[...]

PR #337, opened 2026-04-17
feat(memory): /memory Telegram command (spec 310)

### LINKED ISSUE #310: /memory Telegram command: browse, search, and manage memories
## Summary

Add the `/memory` Telegram command, giving users a first-class interface to browse, search, and manage what Kai remembers about them. This is the trust and transparency feature: users can see exactly what is stored, how confident Haiku was when it stored it, verify accuracy, and delete anything they want removed. Also a tuning instrument: low-confidence extractions are visible, so prompt regressions become fixable instead of invisible.

No other AI assistant offers browsable, user-controlled semantic memory. ChatGPT shows a flat list with no structure or search. Claude shows "project knowledge" with no visibility into what it actually uses. Kai lets the user run the same semantic search the system uses internally and see the results with relevance scores alongside extraction confidence.

Parent issue: #306

### What this operates on

Haiku-extracted facts only (Track 2, `source="extracted"`). Track 1 (raw turn embeddings) is being removed under a sibling issue; once that lands, Qdrant contains extracted facts exclusively and no source filter is needed. Until it lands, the handler filters `source="extracted"` so the UI is stable regardless of removal timing.

### Modified files

- `src/kai/bot.py` - add `/memory` command handler (~200-250 lines for all subcommands)
- `tests/test_bot_memory_command.py` - new test file for the command handler
- `src/kai/memory.py` - no new functions expected. Use the existing `get_all`, `search`, `delete_all`, `delete_by_source`. If aggregation in Python becomes slow at scale, add a Qdrant-side helper then, not now.

### Subcommands

| Command | Behavior |
|---------|----------|
| `/memory` | Dashboard: total fact count and breakdown by tag |
| `/memory <tag>` | Paginated listing of facts in that tag, sorted by recency |
| `/memory <tag> <page>` | Navigate to a specific page |
| `/memory search <query>` | Semantic search with relevance scores shown |
| `/memory review` | Lowest-confidence facts, ascending, for tuning |
| `/memory forget <id>` | Delete a fact by position in the most recent listing |
| `/memory forget all` | Delete all facts for this user (inline-keyboard confirmation) |

### Tag enum (from `src/kai/memory_extraction.py:149-163`)

Haiku emits 1 to 4 tags per fact from this closed set of nine: `preference`, `decision`, `fact`, `constraint`, `confirmed_action`, `project`, `location`, `schedule`, `relationship`. The dashboard aggregates on these; no keyword re-tagging is required.

### Dashboard example

```
Memory: 134 facts across 7 tags

  preference        42
  fact              31
  project           18
  decision          14
  confirmed_action  12
  constraint         9
  schedule           5
  location           2
  relationship       1
```

Built by iterating `get_all(user_id)` and counting tag occurrences in Python. Aggregation happens server-side (no LLM call, no vector search), so the response is instant at expected row counts.

### Topic drill-down example

```
preference (42 facts, showing 1-10 of 42)

1. [Apr 17, conf 0.95] Uses metric except for small measurements,
   where the unit-to-mental-picture mapping matters.
2. [Apr 15, conf 0.90] Never hard-wraps paragraphs; each is one line.
3. [Apr 14, conf 0.88] Prefers concise, chat-friendly responses.
...

Page 1/5  |  /memory preference 2  for next page
```

Each line shows creation date and the confidence Haiku assigned at extraction time. Low-confidence facts stand out visually, which is the first step in the tuning workflow.

### Search example

```
Search: "memory system architecture" (4 results)

1. [match 0.91, conf 0.90] Track 1 dropped Apr 17 2026: only Haiku-
   extracted facts write to Qdrant.
2. [match 0.84, conf 0.95] Spec 320 merged via PR #333 after 8
   review rounds.
3. [match 0.81, conf 0.85] Mem0 v2.0.0 requires explicit
   embedding_model_dims.
4. [match 0.72, conf 0.80] Memory foundation (#307) landed with
   Qdrant + sentence-transformers.
```

Two scores, clearly distinguished:

- **match** (0.0 to 1.0): semantic similarity between the query and the stored fact, computed at query time.
- **conf** (0.5 to 1.0): Haiku's self-assessed confidence at extraction time, stored with the fact.

The first tells the user how relevant the result is to this query; the second tells them how reliable the fact itself is. Different questions, different numbers, visibly distinct labels.

### Review (tuning) view

`/memory review` returns the 20 lowest-confidence facts for the user, sorted ascending by confidence. Purpose: make tuning a one-command workflow. Look at what Haiku was least sure about, delete noise, adjust the extraction prompt if a pattern emerges.

```
Review: lowest-confidence facts (20 shown)

1. [conf 0.52, Apr 12, fact] The Mac mini has 16GB of RAM.
2. [conf 0.55, Apr 9, preference] Likes cooler rooms.
3. [conf 0.58, Apr 3, project] Phi uses Haskell for its parser.
...

/memory forget <id>  to delete any row above
```

### Confirmation quotes

Facts tagged `confirmed_action` carry a `confirmation_quote` field (required by the extraction schema). Display it inline in drill-down and search so the user can audit why Haiku stored the fact:

```
1. [Apr 17, conf 0.95, confirmed_action] User merged PR #333.
   quote: "I see PR #333 is merged, thanks for landing it"
```

For non-`confirmed_action` facts, the quote field is absent and nothing extra is rendered.

### Deletion

`/memory forget <id>` deletes by position in the most recent listing (1-indexed). Pagination state is ephemeral (dict keyed by `chat_id`) so the user can `/memory search X`, see results, then `/memory forget 3` to delete the third result.

`/memory forget all` shows an inline-keyboard confirmation button before proceeding. Irreversible. Backed by `delete_all(user_id=...)`.

### Graceful degradation

If the memory system is disabled (`MEMORY_ENABLED=false` or init failure), `/memory` responds with a short message explaining that semantic memory is not active. No partial views or mystery errors.

## Tests

- Dashboard returns correct tag counts across mixed-tag facts
- Topic drill-down paginates correctly at various page sizes
- Search returns results with both `match` and `conf` labeled and correct
- Review returns the 20 lowest-confidence facts in ascending order
- `/memory forget <id>` deletes the correct fact from the last listing
- `/memory forget all` clears all user facts after confirmation
- Empty states: zero memories, unknown tag, zero search results, zero low-confidence facts
- Graceful degradation when the memory system is disabled
- Multi-user isolation: one user cannot browse or delete another user's facts
- Confirmation quote renders for `confirmed_action` facts and is omitted otherwise

## Dependencies

- **Memory foundation (#307)**: merged.
- **Haiku extraction (PR #333, closed #320)**: merged.
- **Track 1 removal (sibling issue under #306)**: not a strict blocker. The handler filters `source="extracted"` in the interim so the UI is independent of removal timing.
- **Topic pre-classification (#309)**: no longer required. Haiku already emits `tags` as part of its extraction output, so the dashboard and drill-down work on Haiku's own taxonomy. #309 becomes an optional augmentation (a second source of topic tags if we ever want finer filter resolution), not a gating dependency.

## Risk

Low. Read-only commands (dashboard, drill-down, search, review) carry no risk. Deletion by position uses an explicit per-chat listing state, so there is no way to accidentally delete "any memory shaped like X". Bulk delete requires a confirmation keyboard press. All operations scope by `user_id` (Telegram `chat_id`) via Mem0's built-in user isolation.

## Reference

- Extraction schema: `src/kai/memory_extraction.py:140-183`
- Available APIs: `src/kai/memory.py` (`get_all`, `search`, `delete_all`, `delete_by_source`, `get_stats`)
- MemPalace additions: `home/docs/memory-system/mempalace-inspired-additions.md`
- Discussion #301

dcellison:
## Summary

Implements spec 310: a Telegram-facing `/memory` command surface for browsing, searching, and forgetting Haiku-extracted facts. Closes #310.

Four focused commits:

- **C1 (`6357894`)** Add `MEMORY_SEARCH_FLOOR` env var (default 0.3) and consolidate the prior `_MIN_RELEVANCE_THRESHOLD` constant into it. One knob, two paths: both `format_context` (retrieval) and `/memory search` (UI) read the same value at every invocation, so a config edit + restart produces consistent behavior across both paths.

- **C2 (`99f9859`)** Pure-additive helpers and stats extensions in `memory.py`: `get_by_tag`, `delete_by_id`, a `limit` kwarg on `get_all`, plus nine extracted-only aggregate fields on `MemoryStats`. The new aggregates default to "no data" sentinels so the legacy two-arg construction still works. `delete_by_id` does a verify-before-delete on both `user_id` and `metadata.source == "extracted"` as defense-in-depth, even though Mem0's `delete` already scopes by user.

- **C3 (`1323760`)** New module `src/kai/memory_command.py` plus 64 unit tests. Pure builders return `(text, keyboard)` tuples; top-level handlers own the I/O. Per-chat ephemeral cache with lazy 30-minute TTL and single-entry-per-chat semantics keeps every callback under Telegram's 64-byte limit (memory ids are referenced by index, never embedded). `_TAG_ENUM` mirrors the extraction schema with a drift-guard test so future schema changes can't silently exclude tags from `/memory stats`.

- **C4 (`fd3663c`)** Wire the new handlers into `bot.py` and add a five-line `/memory` block to `/help`.

## UX

Plain text rendering throughout (spec §6.0); user-originated content like fact text and confirmation_quote routinely contains Markdown-reserved characters and rendering them as MarkdownV2 would corrupt the display. Forget flows use `answerCallbackQuery` toasts rather than edit-then-sleep-then-edit, keeping the event loop free.

Dashboard hides zero-count tags but stats screen surfaces them as a tuning signal (the spec §6.1 asymmetry, enforced by builders, not the data layer).

## Test plan

- [x] `pytest tests/test_memory_command.py` (64 new tests, all passing)
- [x] `pytest tests/test_memory.py tests/test_config.py` (full memory + config suites green; 244 passing, 9 deselected for Mem0 lock)
- [x] Full suite: 1977 passed, 9 deselected
- [ ] Manual smoke test against running bot once merged: `/memory`, `/memory search <q>`, `/memory stats`, tag drill-down + back, fact view + forget confirmation

## Notes

- No feature flag: `/memory` only reads or deletes the caller's own memories (spec §9). Safe to expose by default.
- No bot-wiring tests added; `test_bot.py` covers pure helpers only and the import smoke check verifies registration. The C3 suite already exercises every handler path through monkeypatched memory primitives.
- Spec 335 (Track 1 removal) ships separately. The `source == "extracted"` filter on every read/write path makes this PR correct regardless of #335 ordering.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -144,6 +144,13 @@ MEMORY_ENABLED=false
 # Models with different dims are rejected at startup. Supporting other
 # dimensions requires changing embedding_model_dims in memory.py.
 #MEMORY_EMBEDDING_MODEL=all-MiniLM-L6-v2
+# Minimum Mem0 similarity score for a memory to be returned. Float in
+# [0.0, 1.0]; default 0.3 matches Mem0's built-in default. Governs both
+# the context-injection path (format_context) and the /memory search UI
+# - one knob, two paths. Raise toward 0.5+ to reduce false positives at
+# the cost of recall; lower toward 0.0 if recall feels missing. Tunable
+# at runtime by editing this file and restarting the service.
+#MEMORY_SEARCH_FLOOR=0.3
 
 # Track 2 Haiku extraction. Requires MEMORY_ENABLED=true and a Claude
 # backend. Off by default at Phase 2 ship (self-reinforcing loop needs
--- a/src/kai/bot.py
+++ b/src/kai/bot.py
@@ -54,7 +54,7 @@ from telegram.ext import (
     filters,
 )
 
-from kai import github_api, services, sessions, webhook
+from kai import github_api, memory_command, services, sessions, webhook
 from kai.config import (
     DATA_DIR,
     MAX_CONTEXT_CEILING,
@@ -2845,6 +2845,12 @@ async def handle_help(update: Update, context: ContextTypes.DEFAULT_TYPE) -> Non
         "/github add <repo> - Watch a repo\n"
         "/github remove <repo> - Unwatch a repo\n"
         "\n"
+        "/memory - Browse remembered facts by tag\n"
+        "/memory search <q> - Semantic search over memories\n"
+        "/memory stats - Counts and confidence distribution\n"
+        "/memory forget <tag> - Delete all memories with a tag\n"
+        "/memory help - /memory subcommand reference\n"
+        "\n"
         "/voice - Toggle voice on/off\n"
         "/voice only - Voice only (no text)\n"
         "/voice on - Text + voice\n"
@@ -3705,12 +3711,14 @@ def create_bot(config: Config, *, use_webhook: bool = True) -> Application:
     app.add_handler(CommandHandler("voices", handle_voices))
     app.add_handler(CommandHandler("webhooks", handle_webhooks))
     app.add_handler(CommandHandler("github", handle_github))
+    app.add_handler(CommandHandler("memory", memory_command.handle_memory_command))
     app.add_handler(CommandHandler("stop", handle_stop))
 
     # Callback query handlers for inline keyboards (pattern-matched)
     app.add_handler(CallbackQueryHandler(handle_model_callback, pattern=r"^model:"))
     app.add_handler(CallbackQueryHandler(handle_voice_callback, pattern=r"^voice:"))
     app.add_handler(CallbackQueryHandler(handle_workspace_callback, pattern=r"^ws:"))
+    app.add_handler(CallbackQueryHandler(memory_command.handle_memory_callback, pattern=r"^mem:"))
 
     # Media handlers (must be before the catch-all text handler)
     app.add_handler(MessageHandler(filters.PHOTO, handle_photo))
--- a/src/kai/config.py
+++ b/src/kai/config.py
@@ -460,6 +460,19 @@ class Config:
     # an executor thread on a hung subprocess.
     memory_extraction_timeout_s: int = 10
 
+    # Minimum Mem0 similarity score for a memory to be returned by
+    # search-driven paths: both `format_context` (context injection
+    # at session start) and the `/memory search` UI surface in
+    # `memory_command.py`. Values below the floor are dropped before
+    # any ranking. Default 0.3 matches Mem0's built-in default and
+    # the prior hard-coded constant; raise toward 0.5+ to reduce
+    # false positives at the cost of recall. Spec 310 §7.5 documents
+    # the "one knob, two paths" decision: keeping the UI floor and
+    # the context-injection floor in lockstep prevents silent
+    # divergence between "what the user sees in /memory search" and
+    # "what Kai pulls into context" after a config change.
+    memory_search_floor: float = 0.3
+
     def get_workspace_config(self, workspace: Path) -> WorkspaceConfig | None:
         """
         Get per-workspace config for a path, or None for global defaults.
@@ -1400,6 +1413,21 @@ def load_config() -> Config:
     except ValueError:
         raise SystemExit("MEMORY_EXTRACTION_TIMEOUT_S must be an integer") from None
 
+    # Search relevance floor. Float in [0.0, 1.0]; default 0.3 matches
+    # Mem0's built-in default and the prior hard-coded constant. Same
+    # try/except pattern as the other memory_* numeric vars: bad input
+    # exits at startup rather than surfacing as a divide-by-zero or
+    # mysterious "no results" symptom at first query. Range is bounded
+    # at both ends because Mem0 cosine similarity is normalized to
+    # [0.0, 1.0]; a floor outside that range silently filters
+    # everything (>1.0) or nothing (<0.0), both of which are footguns.
+    try:
+        memory_search_floor = float(os.environ.get("MEMORY_SEARCH_FLOOR", "0.3"))
+        if memory_search_floor < 0.0 or memory_search_floor > 1.0:
+            raise SystemExit("MEMORY_SEARCH_FLOOR must be between 0.0 and 1.0")
+    except ValueError:
+        raise SystemExit("MEMORY_SEARCH_FLOOR must be a number") from None
+
     # Per-workspace configuration. Loaded after ALLOWED_WORKSPACES so
     # YAML-defined workspaces can be merged into the allowed set.
     workspace_configs = _load_workspace_configs()
@@ -1557,4 +1585,5 @@ def load_config() -> Config:
         memory_extraction_model=memory_extraction_model,
         memory_extraction_budget_usd=memory_extraction_budget_usd,
         memory_extraction_timeout_s=memory_extraction_timeout_s,
+        memory_search_floor=memory_search_floor,
     )
--- a/src/kai/memory.py
+++ b/src/kai/memory.py
@@ -29,10 +29,14 @@ log = logging.getLogger(__name__)
 
 # ── Data classes ────────────────────────────────────────────────────
 
-# Minimum similarity score for a memory to be included in context.
-# Based on smoke testing: clearly relevant results score 0.7+,
-# loosely relevant ~0.35, noise below 0.3.
-_MIN_RELEVANCE_THRESHOLD = 0.3
+# Minimum similarity score for a memory to be included in context
+# was previously a hard-coded 0.3 here. Promoted to a config field as
+# part of spec 310: see `Config.memory_search_floor` (env var
+# MEMORY_SEARCH_FLOOR). Both `format_context` below and the
+# `/memory search` UI in `memory_command.py` read the same value at
+# query time so a config change applies everywhere Kai consults memory.
+# Background: based on smoke testing, clearly relevant results score
+# 0.7+, loosely relevant ~0.35, noise below 0.3.
 
 # Maximum length (chars) for the assistant portion of an ingested
 # exchange. Long tool outputs (file dumps, stack traces, large diffs)
@@ -410,7 +414,14 @@ async def format_context(
     # Weighting happens AFTER this filter (spec §5.3) so a downweighted
     # legacy row cannot survive on raw score, and a boosted extracted fact
     # cannot be rescued below threshold.
-    results = [r for r in results if r.score >= _MIN_RELEVANCE_THRESHOLD]
+    #
+    # Read the floor from config at every call (not at module import) so a
+    # `MEMORY_SEARCH_FLOOR` change applied via service restart takes effect
+    # consistently here AND in the `/memory search` UI; spec 310 §7.5
+    # documents the one-knob-two-paths decision. _config is non-None inside
+    # this branch because is_enabled() returned True above.
+    floor = _config.memory_search_floor
+    results = [r for r in results if r.score >= floor]
     if not results:
         return ""
 
--- a/src/kai/memory.py
+++ b/src/kai/memory.py
@@ -21,7 +21,7 @@ import asyncio
 import logging
 import os
 import re
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 
 from kai.config import DATA_DIR, Config
 
@@ -166,14 +166,54 @@ class MemoryResult:
     memory_type: str  # From metadata["type"]: "exchange", "fact", etc.
     metadata: dict  # Full metadata dict from Mem0
     created_at: str  # ISO timestamp
+    # ISO timestamp of the most recent update; equal to created_at for
+    # rows that have never been refreshed. Surfaced for the /memory tag
+    # view (spec 310 §6.2), which sorts newest-updated first so a fact
+    # bubbled up by re-extraction lands at the top of its tag list.
+    # Defaults to "" so callers and tests that don't pass it remain
+    # source-compatible with the pre-spec-310 dataclass shape.
+    updated_at: str = ""
 
 
 @dataclass(frozen=True)
 class MemoryStats:
-    """Memory statistics for a user."""
+    """Memory statistics for a user.
+
+    Two distinct totals coexist:
+      - `total_count` covers ALL rows for the user (every source). This
+        is the original semantics; left untouched so existing callers do
+        not see a behavior change.
+      - `extracted_count` and every other field below cover ONLY rows
+        with `metadata.source == "extracted"`. The /memory stats UI
+        (spec 310 §6.6) operates on this restricted view: tier badges
+        and cross-source aggregates would clutter the tuning dashboard,
+        and Track 1 / legacy rows have no tags or confidence to report
+        on anyway.
+
+    All extracted-only aggregate fields default to a "no extracted rows"
+    sentinel (None for min/median/max, 0 for counts, empty dict for the
+    grouped fields) so a caller that constructs `MemoryStats(total_count=0,
+    by_type={})` (the legacy two-arg form) still produces a valid value.
+    """
 
     total_count: int
     by_type: dict[str, int]  # {"exchange": 42, "fact": 3, ...}
+
+    # ── Extracted-only aggregates (spec 310 §6.6 / §7.2) ────────────
+    # Every field below is computed over rows with
+    # metadata.source == "extracted". Track 1 and legacy rows do not
+    # contribute. min/median/max are None when extracted_count == 0
+    # because picking 0.0 would be misleading (it is a valid score
+    # value, not "no data"). Counts default to 0 / empty dict.
+    extracted_count: int = 0
+    by_tag: dict[str, int] = field(default_factory=dict)
+    confidence_min: float | None = None
+    confidence_median: float | None = None
+    confidence_max: float | None = None
+    confidence_below_0_7: int = 0
+    confidence_below_0_6: int = 0
+    confirmation_quote_count: int = 0
+    by_prompt_version: dict[str, int] = field(default_factory=dict)
 
 
 # ── Singleton state ─────────────────────────────────────────────────
@@ -233,6 +273,9 @@ def _wrap_result(raw: dict) -> MemoryResult:
     score, created_at, updated_at, user_id, role.
     """
     metadata = raw.get("metadata") or {}
+    # Mem0 surfaces both created_at and updated_at on every row; both
+    # default to "" if the upstream payload does not have them, which
+    # keeps callers (and the tag-view sort) from having to None-check.
     return MemoryResult(
         id=raw.get("id", ""),
         text=raw.get("memory", ""),
@@ -240,6 +283,7 @@ def _wrap_result(raw: dict) -> MemoryResult:
         memory_type=metadata.get("type", "unknown"),
         metadata=metadata,
         created_at=raw.get("created_at", ""),
+        updated_at=raw.get("updated_at", ""),
     )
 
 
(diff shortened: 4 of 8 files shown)


dcellison:
## Review by Kai

The `_is_authorized` reimplementation matches bot.py exactly - no auth bypass risk there.

---

## PR Review: feat(memory): /memory Telegram command (spec 310)

Overall this is a well-structured PR with thorough tests and good defensive coding. A few real issues worth addressing:

---

### Warning

**1. `assert` used for runtime validation in `_encode_callback` (`memory_command.py:228`)**

```python
assert len(encoded.encode("utf-8")) <= 64, f"callback_data too long: {encoded!r}"
```

`assert` statements are stripped when Python runs with `-O` (optimize flag). If the service ever starts with optimization enabled, Telegram would silently truncate over-limit callback data, producing confusing runtime failures. Replace with a proper `if`/`raise`:

```python
if len(encoded.encode("utf-8")) > 64:
    raise ValueError(f"callback_data too long: {encoded!r}")
```

---

**2. `_avg_confidence` is dead code (`memory_command.py:323`)**

The function is defined and documented but never called. `_build_dashboard` reads `stats.confidence_median` directly:

```python
median = stats.confidence_median  # line ~395
```

Either remove `_avg_confidence` or route `_build_dashboard` through it. As-is it creates misleading indirection for future readers.

---

**3. `ftd` verb accepts unvalidated tag from callback data (`memory_command.py:828`)**

```python
tag = args[0]
facts = memory.get_by_tag(user_id=str(chat_id), tag=tag)
```

An authorized user who crafts `mem:ftd:arbitrary-string` bypasses the `/memory forget <tag>` confirmation flow entirely. The behavior is safe (returns empty results → "Forgot 0 facts."), but for a known tag it allows skipping the "This cannot be undone" confirmation screen. A one-line guard closes it:

```python
if tag not in _TAG_ENUM:
    await query.answer(_MSG_SESSION_EXPIRED)
    return
```

---

### Suggestion

**4. `message_id` stored but never read (`memory_command.py` throughout)**

`_ScreenCache.message_id` is written in `_send_or_edit` (`sent.message_id`) but nothing in the module ever reads it. The edit path uses `callback_query.edit_message_text` which doesn't need it. If this is scaffolding for a future "single message per chat" design, note it; if not, remove the field and the write to keep the cache shape honest.

**5. `context` parameter on `_send_or_edit` is unused (`memory_command.py:1057`)**

The function accepts `context: ContextTypes.DEFAULT_TYPE | None` but never references it. Several callers already pass `None`. Drop the parameter - it's API surface with no behavior.

**6. Two full-corpus fetches per forget action**

`_send_fact_view` calls `get_all(limit=None)` to resolve a memory_id, then `_send_forget_fact_confirm` calls `get_all(limit=None)` again to fetch the same fact. For a user with thousands of extracted facts, that's two 100k-cap queries to find one row. A narrow `get_by_id` wrapper in `memory.py` (using the already-present `_memory.get(memory_id=...)` call from `delete_by_id`) would make both callers O(1).

**7. Misleading confidence display when all confidence values are malformed (`memory_command.py:736`)**

In `_build_stats`, when `extracted_count > 0` but every row has a non-numeric confidence, `stats.confidence_min/median/max` are all `None` and the display falls back to `0.00`. The `0.0` sentinel is indistinguishable from a true minimum score of zero. Using `"n/a"` as the fallback string (instead of `0.0`) would match the existing behavior in the fact detail view.

---

No issues found with the data-layer changes in `memory.py`, the config validation, or the test coverage. The `delete_by_id` ownership checks are correct and the `TOCTOU` window on the forget-count label is cosmetic only (delete re-fetches the live list).

dcellison:
## Review by Kai

Looking at only new issues; prior-review items that were addressed (assertion→raise, `ftd` validation, two-fetch pattern, `context` parameter) are not re-raised.

---

### Warning

**1. `tag` verb in callback handler accepts unvalidated tag (`memory_command.py`, `handle_memory_callback` `verb == "tag"` block)**

```python
tag = args[0]
# no _TAG_ENUM check
await _send_tag_view(update, context, chat_id, tag, page, edit=True)
```

`ftd` got the `_TAG_ENUM` guard in this PR, but the read-only `tag` verb did not. A crafted `mem:tag:arbitrary-string:0` callback reaches `memory.get_by_tag` with an off-enum string. The call is safe today (returns empty, renders "No memories with this tag."), but it silently accepted by a path that is documented as operating only over enum-defined tags, and the inconsistency with `ftd` is a maintenance trap. One-line fix mirrors the `ftd` guard:

```python
if tag not in _TAG_ENUM:
    await query.answer(_MSG_SESSION_EXPIRED)
    return
```

---

### Suggestion

**2. `_send_or_edit` silent no-op when `edit=True` but `callback_query is None` (`memory_command.py:1057` area)**

```python
if edit and update.callback_query is not None:
    ...  # only branch that sends when edit=True

if not edit:
    ...  # not reached when edit=True and callback_query is None
```

When `edit=True` and `callback_query is None`, neither branch fires and nothing is sent. This path is unreachable today (all `edit=True` callers are inside `handle_memory_callback` where `callback_query` is guaranteed), but the silence is a trap for future callers of the `_send_*` helpers. An `assert update.callback_query is not None` at the top of the `if edit` branch, or a final `else: raise` after both branches, would surface the contract violation immediately instead of producing a ghost action.

---

### Clean

Data layer (`memory.py` additions), config validation, `get_by_id` ownership scoping, the `ftd` enum guard, and the `_encode_callback` hardening all look correct. Test coverage for the new helpers is thorough.

dcellison:
## Review by Kai

## Review: feat(memory): /memory Telegram command (spec 310)

Prior review items (assertion→raise, `ftd` enum guard, two-fetch pattern, `context` parameter, `tag` verb guard, `_send_or_edit` silent no-op) are all correctly addressed. The data layer additions in `memory.py` are solid. A few new observations on the changed code:

---

### Suggestion

**1. `confidence_below_0_7` / `confidence_below_0_6` display when all confidence values are malformed (`memory_command.py`, `_build_stats`)**

The prior review caught that `confidence_min/median/max` should render `"n/a"` instead of `0.00` when no numeric confidence values exist - that fix is in. But `confidence_below_0_7 = 0` still displays as `0 (0.0%)` in the same state, which reads as "all facts scored ≥ 0.7" rather than "no confidence data". The `"n/a"` on min/median/max already signals the data issue, so this is low priority, but a parenthetical like `(n/a)` would be more honest when `len(confidences) == 0`.

**2. `assert update.message is not None` in `_send_forget_tag_confirm` (`memory_command.py`, two spots)**

```python
except Exception as exc:
    ...
    assert update.message is not None          # stripped by -O
    await update.message.reply_text(_MSG_QUERY_FAILED)
```

These follow the bot.py convention, but they're inconsistent with this module's own `_send_or_edit`, which was explicitly hardened to use `raise ValueError` over `assert` for the same `-O` reason. Since `_send_forget_tag_confirm` is only called from the command handler where `message` is guaranteed, this won't bite in practice, but the asymmetry is a maintenance trap.

**3. `_set_cache` TTL semantics comment vs. constant name**

`_CACHE_TTL_S` is named and documented as "30 minutes", but `_set_cache` resets `created_at` on every write, making it an *idle* timeout rather than an absolute session timeout. The constant name and the `.env.example` comment imply absolute; the implementation delivers idle. Worth clarifying in the constant's docstring.

---

### Clean

`get_by_id` ownership check, `get_by_id`-based O(1) fact fetch, `ftd`/`tag` enum guards, `_encode_callback` raise-vs-assert, `_send_or_edit` contract enforcement, config validation range checks, and the `source == "extracted"` filter layering all look correct. The 64-byte callback ceiling is comfortably met by every encoded callback in the visible diff (`mem:tag:confirmed_action:99` = 26 bytes). Auth scoping is consistent with `bot.py`.

dcellison:
## Review by Kai

Based on my analysis of the diff directly (the code doesn't exist locally - this is a pre-merge review):

---

## Review: feat(memory): /memory Telegram command (spec 310)

Prior items (assert→raise, `ftd`/`tag` enum guards, `get_by_id` O(1) path, `context` param, `_send_or_edit` contract, confidence n/a display, `_set_cache` TTL comment) are all correctly addressed. The data layer additions in `memory.py` are solid.

---

### Warning

**1. Double `query.answer()` in the outer error handler (`memory_command.py`, `handle_memory_callback`)**

Every verb block calls `query.answer()` before its `_send_*` call. If the subsequent `_send_*` raises (e.g., a network error escaping `_send_or_edit`'s fallback path, or the `assert update.effective_chat is not None` firing), the outer `except` handler calls `query.answer(_MSG_QUERY_FAILED)` on an already-answered query. Telegram rejects the second call, raising another exception inside the error handler itself - an unhandled exception that PTB logs but the user never sees.

Fix: track whether the query has been answered, or move `query.answer()` to after the risky operation:

```python
# Before:
await query.answer()
await _send_dashboard(update, context, chat_id, edit=True)

# After:
await _send_dashboard(update, context, chat_id, edit=True)
await query.answer()
```

The `ftd` verb is the one exception where the answer carries a count - that one could call `query.answer()` last too, after the delete loop and dashboard refresh complete.

---

### Suggestion

**2. `ftd` bulk delete: N extra Mem0 `get` calls (`memory_command.py`, `handle_memory_callback`, `verb == "ftd"` block)**

```python
facts = memory.get_by_tag(user_id=str(chat_id), tag=tag)
for fact in facts:
    if memory.delete_by_id(...):
        deleted += 1
```

`get_by_tag` already fetched and verified ownership + source for every row in `facts`. Each `delete_by_id` call re-fetches the same row via `get_by_id` (another Mem0 `get` round-trip) before deleting. For N facts, that's N redundant fetches on top of the initial corpus fetch. For typical tag sets this is invisible, but for a user who built up 200 `preference` facts the extra 200 `get` calls add meaningful latency. An internal `_delete_verified` path that skips re-verification for rows already in hand would avoid this; alternately, the `ftd` loop could call `_memory.delete(memory_id=fact.id)` directly and catch `ValueError` for the TOCTOU race, since `get_by_tag` already did the authorization work.

**3. `_build_stats` percentage denominator when mixed confidence data (`memory_command.py`, `_build_stats`)**

```python
pct_07 = (stats.confidence_below_0_7 / stats.extracted_count) * 100
```

`confidence_below_0_7` counts only rows with numeric confidence below 0.7. `extracted_count` counts all extracted rows including those whose confidence field is non-numeric (skipped by `get_stats`). When a fraction of rows have malformed confidence, the percentage is understated relative to rows-with-data. The `has_confidence` guard already handles the all-None case cleanly; this is the partial case. Low priority - requires deliberately malformed rows to manifest.

**4. Stale keyboard after `_send_or_edit` fallback-to-send (`memory_command.py`, `_send_or_edit`)**

When `edit_message_text` raises a non-"not modified" `BadRequest`, the function logs and falls through to `send_message`. This sends a new message but leaves the original message with its inline keyboard still visible in chat. Buttons on the stale message still trigger callbacks, which will produce odd navigation (double-dashboard, etc.). Wrapping the original message's keyboard with an empty edit before the fallback send would clean this up, though it adds a second Telegram call on what is already an error path.

---

### Clean

`get_by_id` ownership/source scoping, `delete_by_id` TOCTOU handling, the `has_confidence` n/a guard, callback size ceiling, `_encode_callback` hardening, cache TTL semantics comment, `tag`/`ftd` enum parity, and the full config validation chain all look correct. Auth scoping is consistent with `bot.py`.

dcellison:
## Review by Kai

Looking only at code that changed in this diff; prior-review items not re-raised.

---

## Warning

**1. `_send_search` does not filter by `source == "extracted"` (`memory_command.py`, `_send_search`)**

```python
results = memory.search(query, user_id=str(chat_id), limit=_SEARCH_LIMIT)
filtered = [r for r in results if r.score >= floor]
```

The floor is applied but `memory.search()` searches across all sources. Every other read path in this module goes through `get_by_tag` or `get_by_id`, both of which enforce `source == "extracted"`. Search does not. Non-extracted rows (Track 1 exchanges, legacy) appear in the results list; when the user taps one, `get_by_id` returns `None` (source mismatch) and the fact view shows "This memory no longer exists." - confusing for a result that was just displayed. The module docstring's claim that source filtering "lives in exactly one place" implicitly assumes search is out of scope, but that's not documented.

Fix: post-filter in `_send_search`:
```python
filtered = [r for r in results if r.score >= floor and r.metadata.get("source") == "extracted"]
```

---

## Suggestion

**2. `assert update.effective_chat is not None` in `_send_or_edit` is inconsistent with module hardening**

```python
if not edit:
    assert update.effective_chat is not None   # stripped by -O
    await update.effective_chat.send_message(...)
```

Every other defensive check in this module uses `if/raise` specifically because `-O` strips assertions (`_encode_callback`, `_send_or_edit`'s own `edit=True` guard, `_send_forget_tag_confirm`). This one was missed. Replace with:
```python
if update.effective_chat is None:
    raise ValueError("_send_or_edit: effective_chat is None")
```

**3. Grammar: "confirm forget 1 facts" (`memory_command.py`, `_build_forget_tag_confirm`)**

```python
f"confirm forget {count} facts"
```

When `count == 1` this renders "confirm forget 1 facts". Use `"fact" if count == 1 else "facts"`.

---

Data layer additions (`get_by_tag`, `get_by_id`, `delete_by_id`), config validation, answer-after-send ordering, stale keyboard cleanup, `_encode_callback` hardening, and enum guards all look correct.

dcellison:
## Review by Kai

Looking only at code present in this diff; all prior-review items are confirmed addressed.

---

## Review: feat(memory): /memory Telegram command (spec 310)

The PR is in good shape. Five rounds of review have cleared every substantive issue. One new finding and one clean call:

---

### Suggestion

**`_chat_id` and `_user_id` use `assert` inconsistently with the module's `-O` hardening (`memory_command.py`, bottom of file)**

```python
def _chat_id(update: Update) -> int:
    chat = update.effective_chat
    assert chat is not None   # stripped by python -O
    return chat.id
```

Every other defensive check in this module was explicitly converted to `if/raise` for `-O` safety (`_encode_callback`, `_send_or_edit`, `_send_forget_tag_confirm`). These two helpers and the `assert update.message is not None` at the top of `handle_memory_command` were missed. The risk is low: PTB routing guarantees these are non-None for `CommandHandler` and `CallbackQueryHandler`. But under `-O`, the assert strips and any unexpected None produces `AttributeError` instead of a meaningful `ValueError`, violating the module's own stated convention.

---

### Clean

`source == "extracted"` filter in `_send_search`, `_send_or_edit` effective_chat guard, `_encode_callback` raise-vs-assert, double-`query.answer()` ordering, `get_by_id` O(1) path, `_set_cache` TTL comment, `tag`/`ftd` enum parity, `_build_forget_tag_confirm` singular/plural, config validation range checks, and the `has_confidence` n/a guard all look correct. No security issues found; auth scoping is consistent with `bot.py`, and the ownership/source double-check in `delete_by_id` is structurally sound.

dcellison:
## Review by Kai

## Review: feat(memory): /memory Telegram command (spec 310)

All prior-review items are addressed. One new warning and one suggestion on code not previously flagged.

---

### Warning

**`_HELP_TEXT` passed to `query.answer(show_alert=True)` exceeds Telegram's 200-char limit (`memory_command.py`, `_HELP_TEXT` constant and `verb == "help"` branch)**

```python
_HELP_TEXT = (
    "/memory              Browse by tag\n"          # 35 chars
    "/memory search <q>   Semantic search\n"        # 37 chars
    "/memory stats        Counts and confidence distribution\n"  # 56 chars
    "/memory forget <tag> Delete all memories with a tag\n"      # 52 chars
    "/memory help         Show this help"            # 35 chars
)
# Total: ~215 chars. Telegram's answerCallbackQuery text field is limited to 200.
```

When the "Search" button on the dashboard is tapped, PTB passes this string to `answerCallbackQuery`. Telegram returns a `400 Bad Request`, PTB raises `BadRequest`, the outer `except Exception` handler catches it, and calls `query.answer(_MSG_QUERY_FAILED)` - the user sees "Memory query failed." instead of the help text.

Fix: strip the alignment padding; the monospace columns are invisible in Telegram's alert modal anyway:

```python
_HELP_TEXT = (
    "/memory - Browse by tag\n"
    "/memory search <q> - Semantic search\n"
    "/memory stats - Counts and confidence distribution\n"
    "/memory forget <tag> - Delete all memories with a tag\n"
    "/memory help - Show this help"
)
```

---

### Suggestion

**`edit_message_reply_markup` cleanup only catches `BadRequest` (`memory_command.py`, `_send_or_edit` fallback path)**

```python
try:
    await update.callback_query.edit_message_reply_markup(reply_markup=None)
except BadRequest as kb_exc:                 # narrow catch
    log.debug("clear stale keyboard failed (best-effort): %s", kb_exc)
```

The comment says "swallow any BadRequest" but a `NetworkError` or `TimedOut` from PTB propagates out of `_send_or_edit` into the outer `try/except` in `handle_memory_callback`, which calls `query.answer(_MSG_QUERY_FAILED)` on a query that was already answered by the fallback send. Since this is explicitly a best-effort path, `except Exception` is the appropriate catch width here.

---

### Clean

Config validation, `get_by_id` O(1) path, `delete_by_id` ownership/source scoping, `ftd`/`tag` enum guards, `_encode_callback` raise-vs-assert, `_send_or_edit` contract enforcement, `source == "extracted"` filter in `_send_search`, answer-after-send ordering, `has_confidence` n/a guard, `_set_cache` TTL semantics comment, and singular/plural in `_build_forget_tag_confirm` all look correct. The callback size ceiling is comfortably met. Auth scoping is consistent with `bot.py`.

dcellison:
## Review by Kai

## Review: feat(memory): /memory Telegram command (spec 310)

All prior-review items from the seven previous rounds are correctly addressed in this diff. Checking the changed code for new issues only.

---

### Clean

**`memory.py` additions** - `get_by_tag`, `get_by_id`, `delete_by_id`, and the extended `get_stats` are all correct. Ownership and source scoping are consistent. The `MemoryResult.updated_at: str = ""` default preserves backward compatibility for existing callers. The `get_all(limit=None)` → 100,000 ceiling translation is correctly documented and intentional per spec §7.2.1.

**`memory_command.py`** - The answer-after-send pattern is consistently applied across all verbs. The `_TAG_ENUM` guards on both `tag` and `ftd` verbs are in parity. The `_send_or_edit` contract (`edit=True` raises on missing `callback_query`; `effective_chat` check uses `if/raise`) is correctly hardened. `_chat_id` and `_user_id` use `if/raise`. The keyboard cleanup in the fallback path catches `Exception` (not just `BadRequest`). `_HELP_TEXT` is 198 chars, safely under the 200-char `answerCallbackQuery` limit. `source == "extracted"` filter is applied in `_send_search`. Singular/plural in `_build_forget_tag_confirm` is correct. TTL semantics (idle, not absolute) are documented in `_CACHE_TTL_S`.

**`_ScreenCache`** - `message_id` field is gone. `_avg_confidence` dead code is gone. The `context` parameter on `_send_or_edit` is gone.

**`config.py`** - `MEMORY_SEARCH_FLOOR` validation (range check, type check, startup exit) matches the pattern of other numeric memory vars.

**`bot.py`** - Handler registration looks correct. Pattern `r"^mem:"` correctly scopes the callback dispatcher.

**Security** - Auth scoping is consistent with `bot.py`. `delete_by_id` verifies ownership and source before deleting. No cross-user data exposure path visible. Callback enum guards prevent arbitrary tag injection in both read and write paths.

No new issues found. This is ready to merge.

PR #340, opened 2026-04-19
fix(claude): reset HOME on sudo -u for per-user credential isolation

dcellison:
## Summary
- `sudo -u <user>` preserves the caller's `HOME` by default. The bot was spawning `claude` under target users without resetting `HOME`, so `claude` read OAuth credentials from the bot user's `~/.claude/.credentials.json` instead of the target user's.
- Added `-H` to all three claude-spawning sites: `claude.py` (per-chat persistent), `review.py` (PR review), `triage.py` (issue triage). `-H` resets `HOME` from the target user's pw entry.
- Surfaced while onboarding a second OS user; without this fix every user would silently run under the bot user's Claude account (or fail entirely if those credentials were missing).

## Notes
- Each new OS user still needs a one-time interactive `claude login` to populate their own `~/.claude/.credentials.json`. This PR does not change that requirement, only ensures `claude` reads the correct file at spawn time.
- Considered `sudo -i -u` (full login shell) instead of `-H`, but rejected because `-i` chdirs to the target's home, fighting our `cwd=str(self.workspace)` and changing signal/process-group semantics. `-H` is the surgical one-flag fix.

## Test plan
- [ ] CI: `make test` and `make check` both green
- [ ] After deploy: second OS user (sellison) successfully runs `/start` and gets a real Claude response
- [ ] Existing single-user (daniel/admin) flow still works after deploy
- [ ] PR review subprocess (`review.py`) still completes successfully when invoked by the bot


dcellison:
## Review by Kai

**Clean fix. One suggestion worth noting.**

**The core change is correct.** `sudo -H` is the right tool here - it resets `HOME` from the passwd entry rather than inheriting from the caller. The alternative (`-i`) was correctly rejected; it would fight the existing `cwd=str(self.workspace)` and alter process-group semantics.

**No security issues.** The `-H` flag is a hardcoded string with no user input involved. The fix actually tightens the security boundary by ensuring per-user OAuth credential isolation.

**Tests are thorough and accurately updated** across all three modules, verifying positional argument order explicitly.

---

**suggestion** - `review.py:759`, `triage.py:475`

The comments "See claude.py for the same fix" are cross-file references that rot. If `claude.py` is later refactored or the three files diverge, the pointer becomes misleading. The why is already stated in the preceding sentence; drop the forward reference.

```python
# Before:
# and silently exits. See claude.py for the same fix.

# After:
# and silently exits.
```

---

That's the only finding. The fix is surgical, consistent across all three spawn sites, and the sudoers implications are unaffected (`-H` predates the `SETENV:` requirement which only applies to `--preserve-env`).

dcellison:
## Review by Kai

The prior suggestion (removing "See claude.py for the same fix" cross-references) was addressed - both `review.py` and `triage.py` comments now end at "and silently exits." without the forward reference.

**The PR is clean.** Nothing new to raise.

- Core fix is correct and consistent across all three spawn sites
- Tests accurately reflect the new positional indices in all affected assertions
- No security concerns; `-H` is hardcoded and actually tightens credential isolation
- Comment language is consistent across all three files

PR #342, opened 2026-04-20
fix(install): emit sudoers rule per users.yaml os_user

### LINKED ISSUE #341: install: emit sudoers rule per users.yaml os_user
## Problem

`_apply_sudoers` in `src/kai/install.py` writes `/etc/sudoers.d/kai` from a single legacy `CLAUDE_USER` env var. It does not iterate `/etc/kai/users.yaml` to emit rules for the per-user `os_user` values that the runtime now relies on (`pool.py`, `claude.py`).

Every `python -m kai install apply` (i.e., every `sudo make install`) truncates and rewrites the file. Any `kai ALL=(<os_user>) SETENV: NOPASSWD: <claude_bin>` rules added by hand for additional users get wiped on the next install. The bot then fails for those users with a silent claude exit within ~50 ms of spawn; the actual `sudo` rejection (`Sorry, user kai is not allowed to execute ... as <user>`) is only visible when claude is invoked outside the bot, because the daemon logs claude's stderr at DEBUG level only.

## Reproduction

1. Configure two users in `users.yaml`, the second with an `os_user` field distinct from the bot's service user.
2. Run `sudo make install` and inspect `/etc/sudoers.d/kai` — only the `kai ALL=(root) NOPASSWD: /bin/cat ...` lines are present; no per-user `(os_user)` rule is emitted.
3. The second user's Telegram messages produce `Claude process EOF` in the daemon log within ~50 ms of spawn.

## Proposed fix

`_apply_sudoers` should:

- Load `users.yaml` (or accept it as a parameter for testability).
- Collect the unique set of `os_user` values across `users` entries, skipping any user whose `os_user` matches the bot's service user (those use the self-sudo skip path and need no rule).
- Emit one `<service_user> ALL=(<os_user>) SETENV: NOPASSWD: <claude_bin>` line per collected user, in addition to the existing read-protected-config rules.
- Continue to honor the legacy single-`CLAUDE_USER` env var for backwards compatibility, but treat `users.yaml` as authoritative when both are present.

## Acceptance

- After `sudo make install`, `/etc/sudoers.d/kai` contains a per-user rule for every distinct `os_user` in `users.yaml`.
- Adding a new user to `users.yaml` and re-running `sudo make install` adds their rule without manual `visudo`.
- Removing a user from `users.yaml` and re-running `sudo make install` removes their rule.
- Tests cover: (a) zero `os_user` users — no per-user rules emitted; (b) one `os_user` — one rule; (c) multiple distinct `os_user` values — one rule each, deduplicated; (d) `os_user` matching service user — no rule emitted.

## Out of scope

- Memory env vars missing from the install wizard (separate issue).
- Broader "install should never destroy runtime state" refactor (separate consideration).

dcellison:
Fixes #341.

## Problem

`_apply_sudoers` only emitted a per-user sudo rule for the legacy single `CLAUDE_USER` env var. Per-user `os_user` values from `users.yaml` were ignored, so every `sudo make install` truncated `/etc/sudoers.d/kai` and wiped any hand-added rules. The runtime spawn for those users then failed silently with `Sorry, user kai is not allowed to execute ... as <os_user>`, surfacing in the daemon log only as `Claude process EOF` ~50 ms after spawn.

## Changes

- `_collect_os_users_from_yaml(path)` — lightweight reader returning distinct, non-empty `os_user` strings. Missing file returns `[]` (first install). Malformed YAML raises so install fails loudly rather than silently emitting no rules. Non-string values (PyYAML can parse `yes`/`no`/numerics as bool/int) are skipped.
- `_generate_sudoers(service_user, claude_user=None, os_users=())` — emits one `(target) SETENV: NOPASSWD: <claude_bin>` rule per distinct user across `claude_user` and `os_users`, dropping any that match `service_user` (those go through `resolve_claude_user()`'s self-sudo skip path at runtime).
- `_apply_sudoers` — loads `/etc/kai/users.yaml` by default; path parameterized for testability. Caller in `_cmd_apply` is unchanged.

## Acceptance coverage

Tests in `tests/test_install.py`:

- `test_no_per_user_rule_when_os_users_empty` — case (a)
- `test_one_os_user_emits_one_rule` — case (b)
- `test_multiple_os_users_emit_one_rule_each` — case (c)
- `test_duplicate_os_users_deduped` — case (c, dedupe)
- `test_os_user_matching_service_user_skipped` — case (d)
- `test_legacy_claude_user_combined_with_os_users` — backwards-compat dedupe

Plus a `TestCollectOsUsersFromYaml` class covering missing file, empty file, no `users:` key, non-list `users`, no `os_user` field, dedupe order, whitespace stripping, non-string values, and malformed YAML.

## Test plan

- [x] `make check` passes
- [x] `make test` passes (2034 tests, +23 new)
- [ ] `sudo make install` on the Mac mini and verify `/etc/sudoers.d/kai` contains one `kai ALL=(<os_user>) SETENV: NOPASSWD: <claude_bin>` line per distinct `os_user` in `users.yaml`
- [ ] Telegram message from the second user (sellison) processes without `Claude process EOF`

## Out of scope

Memory env vars missing from the install wizard (separate future issue).

### DIFF
--- a/src/kai/install.py
+++ b/src/kai/install.py
@@ -34,6 +34,7 @@ import sys
 import tempfile
 import textwrap
 import time
+from collections.abc import Iterable
 from pathlib import Path
 
 import yaml
@@ -997,7 +998,73 @@ def _generate_users_yaml(
     return "\n".join(lines)
 
 
-def _generate_sudoers(service_user: str, claude_user: str | None = None) -> str:
+def _collect_os_users_from_yaml(users_yaml_path: str | Path) -> list[str]:
+    """
+    Read users.yaml and return distinct, non-empty os_user values.
+
+    The runtime (pool.py / claude.py) spawns the inner Claude as each user's
+    `os_user` via `sudo -u`. That requires a matching sudoers rule for every
+    distinct os_user. Without this loader, only the legacy single CLAUDE_USER
+    env var got a rule, and additional users had to be hand-added with visudo
+    only to be wiped by the next `make install`.
+
+    The reader is intentionally lightweight: it does not validate roles,
+    models, or providers (config._load_user_configs already does that at
+    runtime). Install only needs the os_user strings.
+
+    Behavior:
+        - Missing file → []   (first install: users.yaml may not exist yet)
+        - Empty / non-dict / no `users:` key → []
+        - Malformed YAML → raises (yaml.YAMLError); install should fail loudly
+          rather than silently emit no per-user rules
+        - Non-string / empty / whitespace-only os_user values are skipped
+        - Duplicate os_user values are deduplicated, preserving first-seen order
+
+    Args:
+        users_yaml_path: Path to users.yaml (typically /etc/kai/users.yaml).
+
+    Returns:
+        Ordered list of unique non-empty os_user strings.
+    """
+    path = Path(users_yaml_path)
+    if not path.exists():
+        return []
+
+    text = path.read_text(encoding="utf-8")
+    if not text.strip():
+        return []
+
+    data = yaml.safe_load(text)
+    if not isinstance(data, dict):
+        return []
+
+    users = data.get("users")
+    if not isinstance(users, list):
+        return []
+
+    seen: set[str] = set()
+    result: list[str] = []
+    for entry in users:
+        if not isinstance(entry, dict):
+            continue
+        os_user = entry.get("os_user")
+        # PyYAML may parse "yes"/"no"/numeric os_user values as bool/int;
+        # only strings are valid OS usernames here.
+        if not isinstance(os_user, str):
+            continue
+        normalized = os_user.strip()
+        if not normalized or normalized in seen:
+            continue
+        seen.add(normalized)
+        result.append(normalized)
+    return result
+
+
+def _generate_sudoers(
+    service_user: str,
+    claude_user: str | None = None,
+    os_users: Iterable[str] = (),
+) -> str:
     """
     Generate sudoers rules for the service user to read protected config files.
 
@@ -1009,12 +1076,17 @@ def _generate_sudoers(service_user: str, claude_user: str | None = None) -> str:
     Falls back to /bin/cat and /usr/bin/tee if the binaries aren't found
     in the current PATH (e.g., when running in a minimal environment).
 
-    When claude_user is set, an additional rule allows the service user to
-    run the claude binary as that user (for process isolation via sudo -u).
+    Per-user `(target_user) SETENV: NOPASSWD:` rules for the claude binary are
+    emitted for every distinct user in `os_users` and `claude_user` combined.
+    Users matching `service_user` are skipped (the runtime detects self-sudo
+    via resolve_claude_user() and spawns claude directly without sudo).
 
     Args:
         service_user: The OS username that runs the Kai service.
-        claude_user: Optional OS username for the inner Claude process.
+        claude_user: Optional OS username for the inner Claude process
+            (legacy single CLAUDE_USER env var path; kept for backwards compat).
+        os_users: Distinct os_user values from users.yaml. Combined with
+            claude_user; duplicates and self-sudo entries are dropped.
 
     Returns:
         The sudoers file contents as a string.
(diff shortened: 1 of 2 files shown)


dcellison:
## Review by Kai

## PR Review: fix/install-sudoers-multiuser

The fix is correct and the test coverage is thorough. Three findings:

---

### WARNING - Unsanitized username written to sudoers

**`install.py`, `_generate_sudoers` (line ~1137)**

`target` values from `os_users` are inserted directly into the sudoers rule string with no validation:

```python
rules += f"{service_user} ALL=({target}) SETENV: NOPASSWD: {claude_bin}\n"
```

If an `os_user` value in `users.yaml` contains a `)` or newline (e.g. `alice) NOPASSWD: ALL\n`), it produces a syntactically valid but unintended sudoers rule. Unlike SQL, sudoers has no parameterized form, but the mitigation is a simple regex guard before the loop:

```python
import re
_VALID_USERNAME = re.compile(r'^[a-z_][a-z0-9_.-]{0,31}$')
# raise or skip if not _VALID_USERNAME.match(target)
```

Exploitability is low because `/etc/kai/users.yaml` is presumably root-owned, but files under `/etc/sudoers.d/` are included directly without automatic `visudo -c` validation, so a malformed entry can silently grant extra privileges or break sudo system-wide. Worth a hard error in `_collect_os_users_from_yaml` rather than a silent skip.

---

### SUGGESTION - TOCTOU in `_collect_os_users_from_yaml`

**`install.py`, lines ~1040-1042**

```python
if not path.exists():
    return []
text = path.read_text(encoding="utf-8")
```

The `exists()` + `read_text()` pair has a race. More Pythonic and race-free:

```python
try:
    text = path.read_text(encoding="utf-8")
except FileNotFoundError:
    return []
```

Low severity in practice (root-only install path), but the pattern is worth fixing since it's a one-liner improvement.

---

### SUGGESTION - `_apply_sudoers` silently swallows YAML errors on dry-run path

**`install.py`, `_apply_sudoers`**

`_collect_os_users_from_yaml` is called unconditionally before the `if dry_run` branch. That's correct - a malformed YAML should abort even a dry run. Just worth a brief comment explaining the intentional fail-fast behavior, since a reader might wonder why the YAML load isn't inside the `if not dry_run` block.

---

### Clean items (no issue)

- `yaml.safe_load` used throughout - no arbitrary code execution risk.
- Mutable default avoided correctly (`os_users: Iterable[str] = ()`).
- Dedup logic preserves insertion order correctly using parallel `seen` set + `result` list.
- `claude_user=None` backwards compat path is handled; combined dedup with `os_users` is correct.
- Test coverage of non-string PyYAML coercions (bool/int `os_user`) is exactly right.

dcellison:
## Review by Kai

All three findings from the prior review are addressed in this diff. Here's what I see in the new code:

**Verified addressed**
- Username injection: `_validate_os_user` called in `_collect_os_users_from_yaml` and again as defense-in-depth in `_generate_sudoers`. Parametrized test covers `)`, newline, whitespace, semicolon, and slash injection vectors.
- TOCTOU: `try/except FileNotFoundError` pattern used correctly (lines +1038-1041).
- Dry-run gate: YAML load happens before `if dry_run`, comment explains the intent.

**One new finding**

**SUGGESTION - misleading test comment, `tests/test_install.py`, `test_self_sudo_skip_runs_before_validation`**

```python
# If we passed the service_user as os_user with bad chars, the skip
# would happen first
```

This is factually wrong. The guard in `_generate_sudoers` is `candidate == service_user`, an exact string match. Passing `"kai) bad"` as an `os_user` when `service_user == "kai"` would *not* skip - it would fall through to `_validate_os_user` and raise. The test itself only uses a clean `"kai"` value, so it passes, but the comment will mislead future readers about the actual control flow. Either remove the comment or change it to: "service_user match is an exact-string comparison; a crafted value would not skip, it would raise."

**One observation (not a bug)**

`_validate_os_user` and `_OS_USER_RE` are referenced in this diff but not defined within it. Their correctness is load-bearing for the security fix and cannot be audited from the diff alone. The parametrized injection tests give reasonable confidence the regex covers the key cases, but whoever approves this should verify the regex in the source directly.

**Everything else is clean.** The dedup logic, `claude_user=None` backwards-compat path, `yaml.safe_load` usage, mutable-default avoidance, and test coverage of PyYAML bool/int coercions are all correct.

dcellison:
## Review by Kai

This is the third pass. Prior findings (injection, TOCTOU, dry-run gate comment) are all addressed. Here is what I see in the new diff:

---

**CLEAN - no new critical or warning findings.**

---

### Observation: `_validate_os_user` / `_OS_USER_RE` still not in the diff

The prior review flagged this. It remains true: these are load-bearing for the security fix and cannot be audited from the diff alone. The parametrized injection tests in `TestCollectOsUsersFromYaml.test_invalid_os_user_raises` cover `)`, newline, whitespace, semicolon, equals, slash, and at-sign, which gives reasonable confidence. Whoever approves should do a one-line grep in the source to confirm the regex is `^[a-z_][a-z0-9_.-]{0,31}$` or equivalent.

---

### Verified clean items

- `for candidate in (claude_user, *os_users)`: unpacking a generic `Iterable[str]` into a tuple is correct; generator inputs are handled.
- `if not candidate` guard correctly short-circuits empty-string `claude_user` before `_validate_os_user`, consistent with the YAML loader's behavior.
- `yaml.safe_load` on an empty string returns `None`; the `isinstance(data, dict)` guard handles it, but the earlier `text.strip()` check exits first - no gap.
- `test_os_user_is_stripped` correctly round-trips through `yaml.safe_dump` (PyYAML quotes strings with leading/trailing spaces, `safe_load` preserves them, `strip()` normalizes before validation).
- `test_no_per_user_rule_when_os_users_empty` asserting `"claude" not in result` is safe - the claude binary path is inside the `if target_users:` block, so with empty `os_users` and no `claude_user`, it is never reached.
- Mutable-default avoidance (`os_users: Iterable[str] = ()`), insertion-order dedup, `claude_user=None` backwards-compat path - all correct.

The implementation is sound. Ship pending the manual `sudo make install` verification in the test plan.

PR #344, opened 2026-04-20
fix(install): prompt for memory env vars in wizard (#343)

### LINKED ISSUE #343: install: capture MEMORY_ env vars in config wizard
## Problem

`_cmd_config()` in `src/kai/install.py` does not prompt for any `MEMORY_*` env vars. Result: every `sudo make install` writes a fresh `/etc/kai/env` without the memory configuration, the daemon restarts with `MEMORY_ENABLED` defaulting to false, and the semantic memory system silently goes dark for every user.

The dataclass field (`Config.memory_enabled` and siblings in `config.py`) and the env-var parsing in `load_config()` already exist. The persistence chain (`install.conf` → `_generate_env_file` → `/etc/kai/env`) also works generically. The only missing link is the wizard prompt that captures the value into the env dict in the first place.

This is the same regression class as #341 (sudoers rules wiped on reinstall): install destroys runtime state that the wizard never knew to capture. Hand-edits to `/etc/kai/env` survive only until the next `sudo make install`.

Affected env vars (every one is read in `config.py` `load_config()` but absent from `_cmd_config()`):

- `MEMORY_ENABLED` (toggle)
- `MEMORY_SEARCH_LIMIT`
- `MEMORY_TOKEN_BUDGET`
- `MEMORY_EMBEDDING_MODEL`
- `MEMORY_EXTRACTION_ENABLED`
- `MEMORY_EXTRACTION_MODEL`
- `MEMORY_EXTRACTION_BUDGET_USD`
- `MEMORY_EXTRACTION_TIMEOUT_S`

## Reproduction

1. Configure memory by hand-editing `/etc/kai/env` to include `MEMORY_ENABLED=true` and any related vars.
2. Reload the daemon and confirm memory is active (e.g., a memory-admin command returns the enabled state).
3. Run `sudo make install` (e.g., to deploy any unrelated change).
4. Inspect `/etc/kai/env` - the `MEMORY_*` lines are gone.
5. After daemon restart, memory is silently disabled for every user.

## Proposed fix

Extend `_cmd_config()` to prompt for memory vars, following the existing "if non-default, write" idiom already used at `install.py:780-792`:

- Single toggle prompt: `MEMORY_ENABLED` (default false). Skip the rest if false.
- If true, prompt only for the tunables operators actually want to set:
  - `MEMORY_EXTRACTION_ENABLED` (default false)
  - `MEMORY_EXTRACTION_BUDGET_USD` (default 0.01; suggest 0.05+ in the prompt help text)
  - `MEMORY_TOKEN_BUDGET` (default 2000)
- Other vars (`MEMORY_SEARCH_LIMIT`, `MEMORY_EMBEDDING_MODEL`, `MEMORY_EXTRACTION_MODEL`, `MEMORY_EXTRACTION_TIMEOUT_S`) keep their dataclass defaults; no prompts to avoid wizard fatigue. Operators wanting to override can hand-add to `install.conf` directly.
- Validation reuses the existing helpers (`_validate_positive_int`, `_validate_positive_float`).
- Each captured value lands in the `env` dict, which `_cmd_config()` already serializes to `install.conf`. From there `_apply_secrets()` writes them to `/etc/kai/env` on every apply - so the values survive subsequent `sudo make install` runs.
- Add `.env.example` entries and update `_CONFIG_ENV_VARS` in `tests/test_config.py`.

## Acceptance

- After `sudo make install`, `/etc/kai/env` contains `MEMORY_ENABLED` (and any other prompted vars) when the operator answered `y` in the wizard.
- Re-running `sudo make install` without re-running `make config` preserves the previously chosen memory values (they live in `install.conf` and round-trip through apply).
- Toggling `MEMORY_ENABLED` in the wizard from `y` back to `n` removes the var from the next-generated `/etc/kai/env` (consistent with the existing "if non-default, write" pattern).
- Tests cover: (a) wizard with `MEMORY_ENABLED=n` produces an env dict without `MEMORY_*` keys; (b) wizard with `MEMORY_ENABLED=y` and prompted tunables produces an env dict containing them; (c) round-trip through `_generate_env_file()` preserves the values; (d) re-running the wizard with an existing `install.conf` defaults to the previously chosen values.

## Out of scope

- Changing memory's per-user partitioning (already correct: `user_id = telegram_chat_id`).
- Promoting `MEMORY_EMBEDDING_MODEL` or `MEMORY_EXTRACTION_MODEL` to wizard prompts (operators rarely change these; keep dataclass defaults).
- Any audit of "what other env vars does install silently drop" - that belongs in a broader "install should never destroy runtime state" issue if needed (#341 and this one are the two known instances).

dcellison:
Closes #343.

## Summary

The install wizard never prompted for `MEMORY_*` env vars, so every `sudo make install` wrote a fresh `/etc/kai/env` without `MEMORY_ENABLED`, silently disabling Mem0 + Qdrant for every user. Same regression class as #341 (sudoers wiped on reinstall): the wizard fails to capture state that the install path then destroys.

The dataclass fields and `load_config()` parsing already existed; only the wizard capture was missing.

## Changes

`src/kai/install.py` - new `-- Semantic memory --` section in `_cmd_config()`:
- `MEMORY_ENABLED` toggle (default false)
- If enabled and `agent_backend == "claude"`: `MEMORY_EXTRACTION_ENABLED` toggle, and (when extraction is enabled) `MEMORY_EXTRACTION_BUDGET_USD` with positive-float validation
- If enabled: `MEMORY_TOKEN_BUDGET` with positive-int validation

Other tunables (`MEMORY_SEARCH_LIMIT`, `MEMORY_EMBEDDING_MODEL`, `MEMORY_EXTRACTION_MODEL`, `MEMORY_EXTRACTION_TIMEOUT_S`) keep their dataclass defaults; operators wanting to override can hand-edit `install.conf`.

Captured values land in the `env` dict via the existing "if non-default, write" idiom, so `install.conf` -> `_apply_secrets` -> `/etc/kai/env` round-trips them on every apply.

`tests/test_install.py` - four new acceptance tests matching the issue:
- `test_memory_disabled_omits_env_keys` - `MEMORY_ENABLED=false` produces no `MEMORY_*` keys
- `test_memory_enabled_writes_tunables` - enabled + tunables land in env
- `test_memory_round_trip_through_env_file` - values survive `_generate_env_file()`
- `test_memory_defaults_from_existing_install_conf` - re-running with existing config defaults to prior choices

Existing `TestCmdConfig` input lists updated with the new prompt answer.

## Test plan
- [x] `make check` (ruff + ruff format)
- [x] `make test` for `tests/test_install.py::TestCmdConfig` (10/10 pass)
- [x] Full suite (1920 pass excluding pre-existing memory tests blocked by live daemon's Qdrant lock; unrelated to this change)

### DIFF
--- a/src/kai/install.py
+++ b/src/kai/install.py
@@ -696,6 +696,49 @@ def _cmd_config() -> None:
         )
     print()
 
+    # -- Semantic memory --
+    # MEMORY_ENABLED is a single global toggle for Mem0 + Qdrant. Per-user
+    # partitioning happens at runtime via user_id = telegram_chat_id, so
+    # there is no users.yaml branching here. Without these prompts the
+    # wizard silently drops memory config on every reinstall (see #343).
+    print("-- Semantic memory --")
+    memory_enabled = _prompt_bool(
+        "Enable semantic memory (Mem0 + Qdrant)",
+        existing_env.get("MEMORY_ENABLED", "false").lower() in ("1", "true", "yes"),
+    )
+    # Defaults match the dataclass values in config.py. Only non-defaults
+    # (or memory_enabled=true itself) are written to the env dict below.
+    memory_extraction_enabled = False
+    memory_extraction_budget_usd = "0.01"
+    memory_token_budget = "2000"
+    if memory_enabled:
+        # Haiku extraction requires a Claude backend (config.py refuses
+        # to enable it otherwise). Skip the prompt for non-claude backends
+        # rather than offering an option the operator can't use.
+        if agent_backend == "claude":
+            memory_extraction_enabled = _prompt_bool(
+                "Enable Haiku extraction (proactive memory writes)",
+                existing_env.get("MEMORY_EXTRACTION_ENABLED", "false").lower() in ("1", "true", "yes"),
+            )
+            if memory_extraction_enabled:
+                while True:
+                    memory_extraction_budget_usd = _prompt(
+                        "Per-extraction USD budget (suggest 0.05 or higher)",
+                        existing_env.get("MEMORY_EXTRACTION_BUDGET_USD", "0.01"),
+                    )
+                    if _validate_positive_float(memory_extraction_budget_usd):
+                        break
+                    print("  Must be a positive number.")
+        while True:
+            memory_token_budget = _prompt(
+                "Memory context token budget per turn",
+                existing_env.get("MEMORY_TOKEN_BUDGET", "2000"),
+            )
+            if _validate_positive_int(memory_token_budget):
+                break
+            print("  Must be a positive integer.")
+    print()
+
     # -- External services --
     print("-- External services --")
     perplexity_key = _prompt(
@@ -790,6 +833,18 @@ def _cmd_config() -> None:
         env["PR_REVIEW_TIMEOUT_S"] = pr_review_timeout_s
     if pr_review_budget_usd != "1.0":
         env["PR_REVIEW_BUDGET_USD"] = pr_review_budget_usd
+
+    # Semantic memory: global env vars (per-user partitioning is runtime).
+    # Toggling memory_enabled from true back to false correctly drops
+    # MEMORY_* keys here, so the next /etc/kai/env reflects the new state.
+    if memory_enabled:
+        env["MEMORY_ENABLED"] = "true"
+        if memory_extraction_enabled:
+            env["MEMORY_EXTRACTION_ENABLED"] = "true"
+        if memory_extraction_budget_usd != "0.01":
+            env["MEMORY_EXTRACTION_BUDGET_USD"] = memory_extraction_budget_usd
+        if int(memory_token_budget) != 2000:
+            env["MEMORY_TOKEN_BUDGET"] = memory_token_budget
 
     # Build and write install.conf
     conf = {
--- a/src/kai/install.py
+++ b/src/kai/install.py
@@ -712,9 +712,10 @@ def _cmd_config() -> None:
     memory_extraction_budget_usd = "0.01"
     memory_token_budget = "2000"
     if memory_enabled:
-        # Haiku extraction requires a Claude backend (config.py refuses
-        # to enable it otherwise). Skip the prompt for non-claude backends
-        # rather than offering an option the operator can't use.
+        # Haiku extraction only fires when the active backend is Claude
+        # (bot.py:3609 silently skips it otherwise - no startup error,
+        # no log line). Skip the prompt for non-claude backends rather
+        # than offer an option whose effect is invisible at runtime.
         if agent_backend == "claude":
             memory_extraction_enabled = _prompt_bool(
                 "Enable Haiku extraction (proactive memory writes)",
@@ -847,6 +848,15 @@ def _cmd_config() -> None:
                 env["MEMORY_EXTRACTION_BUDGET_USD"] = memory_extraction_budget_usd
         if int(memory_token_budget) != 2000:
             env["MEMORY_TOKEN_BUDGET"] = memory_token_budget
+
+    # Drop stale extraction keys when the backend isn't Claude. Mirrors
+    # the CLAUDE_MODEL/CLAUDE_MAX_BUDGET_USD cleanup above: bot.py:3609
+    # silently ignores these on non-claude backends, so leaving them in
+    # /etc/kai/env is misleading without effect. A user who flips backend
+    # from claude to goose should not see lingering extraction config.
+    if agent_backend != "claude":
+        env.pop("MEMORY_EXTRACTION_ENABLED", None)
+        env.pop("MEMORY_EXTRACTION_BUDGET_USD", None)
 
     # Build and write install.conf
     conf = {
--- a/src/kai/install.py
+++ b/src/kai/install.py
@@ -837,12 +837,14 @@ def _cmd_config() -> None:
     # Semantic memory: global env vars (per-user partitioning is runtime).
     # Toggling memory_enabled from true back to false correctly drops
     # MEMORY_* keys here, so the next /etc/kai/env reflects the new state.
+    # Numeric comparisons (not string) so "0.010" or "2000.0" are not
+    # treated as non-default and spuriously written.
     if memory_enabled:
         env["MEMORY_ENABLED"] = "true"
         if memory_extraction_enabled:
             env["MEMORY_EXTRACTION_ENABLED"] = "true"
-        if memory_extraction_budget_usd != "0.01":
-            env["MEMORY_EXTRACTION_BUDGET_USD"] = memory_extraction_budget_usd
+            if float(memory_extraction_budget_usd) != 0.01:
+                env["MEMORY_EXTRACTION_BUDGET_USD"] = memory_extraction_budget_usd
         if int(memory_token_budget) != 2000:
             env["MEMORY_TOKEN_BUDGET"] = memory_token_budget
 
--- a/tests/test_install.py
+++ b/tests/test_install.py
@@ -1037,9 +1037,6 @@ class TestCmdConfig:
         _cmd_config()
 
         conf = json.loads((tmp_path / "install.conf").read_text())
-        # _generate_env_file is what _apply_secrets calls to render
-        # /etc/kai/env from the env dict. Round-tripping here proves the
-        # values reach the file the daemon actually reads at startup.
         rendered = _generate_env_file(conf["env"])
         assert 'MEMORY_ENABLED="true"' in rendered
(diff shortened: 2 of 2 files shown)


dcellison:
## Review by Kai

The PR is well-structured and the core fix is correct. Three findings:

**Warning - `MEMORY_EXTRACTION_BUDGET_USD` guard is misindented (`install.py` line ~847):**

```python
if memory_enabled:
    env["MEMORY_ENABLED"] = "true"
    if memory_extraction_enabled:
        env["MEMORY_EXTRACTION_ENABLED"] = "true"
    if memory_extraction_budget_usd != "0.01":        # <-- same level as extraction check
        env["MEMORY_EXTRACTION_BUDGET_USD"] = ...
```

Functionally safe today because `memory_extraction_budget_usd` can only deviate from `"0.01"` when extraction is enabled. But the guard should be nested under `if memory_extraction_enabled:` to reflect intent - as written, a future change to the prompt logic could silently write the key when extraction is off.

**Suggestion - float default comparison is lexicographic (`install.py` line ~847):**

`memory_extraction_budget_usd != "0.01"` treats `"0.010"` as non-default. A user who enters `"0.010"` (valid per `_validate_positive_float`) gets a spurious env entry. Use `float(memory_extraction_budget_usd) != 0.01` for consistency with the parallel `int(memory_token_budget) != 2000` check.

**Suggestion - multi-paragraph docstring in test helper (`tests/test_install.py` line ~990):**

```python
"""Build a default input list with a configurable memory block.

Captures the wizard's prompt order so each memory test only needs
to vary the section under test. Mirrors test_writes_install_conf.
"""
```

Project style (CLAUDE.md) caps docstrings at one short line. Collapse to one line or drop the body.

No security issues. No bugs in the happy path. The round-trip test and defaults-from-existing-conf test are the right acceptance tests for this regression class.

dcellison:
## Review by Kai

All three findings from the prior review have been addressed:

- `MEMORY_EXTRACTION_BUDGET_USD` guard is now correctly nested under `if memory_extraction_enabled:` (install.py line ~843)
- Float comparison uses `float(memory_extraction_budget_usd) != 0.01` (install.py line ~845)
- `_base_inputs` docstring collapsed to one line (test_install.py line ~947)

One new minor finding:

**Suggestion - multi-line comment block in test body (`tests/test_install.py`, `test_memory_round_trip_through_env_file`):**

```python
# _generate_env_file is what _apply_secrets calls to render
# /etc/kai/env from the env dict. Round-tripping here proves the
# values reach the file the daemon actually reads at startup.
```

Same style rule as the prior docstring note - this explains "what", not "why". The function name and test name already convey the intent. Worth collapsing or dropping.

No bugs, no security issues, no logic errors in the happy path or edge cases. The fix is correct and the test coverage - particularly the round-trip and defaults-from-existing-conf tests - is the right approach for this regression class.

dcellison:
## Review by Kai

The PR is clean. All prior findings have been addressed:

- `MEMORY_EXTRACTION_BUDGET_USD` guard correctly nested under `if memory_extraction_enabled:`
- Float comparison uses `float(...) != 0.01` not string equality
- `_base_inputs` docstring collapsed to one line
- `test_memory_round_trip_through_env_file` multi-line comment block removed

No new bugs, security issues, or logic errors. The validated-float/int inputs (`MEMORY_EXTRACTION_BUDGET_USD`, `MEMORY_TOKEN_BUDGET`) cannot carry shell metacharacters, so no injection risk in `_generate_env_file`. Test coverage is appropriate - disabled, enabled, round-trip, and defaults-from-existing all cover the regression class correctly.

dcellison:
## Review by Kai

The PR is clean. All prior findings are resolved. Two new suggestion-level findings:

**Suggestion - `test_memory_defaults_from_existing_install_conf` silently depends on `agent_backend` default (`tests/test_install.py`, ~line 1080)**

The seeded `existing` conf omits `AGENT_BACKEND`. The test presses Enter at every prompt, so the agent backend defaults to whatever `_prompt()` returns for an empty existing value. The test then asserts `MEMORY_EXTRACTION_ENABLED == "true"` - which only survives if the backend default is `"claude"`. If that default ever changes, the extraction keys get popped by the cleanup block and the assertion fails with no obvious connection to the backend default. Add `"AGENT_BACKEND": "claude"` to the seeded env to make the dependency explicit.

**Suggestion - no test for "was enabled, now disabled" round-trip (`tests/test_install.py`)**

The PR description explicitly claims "Toggling memory_enabled from true back to false correctly drops MEMORY_* keys." `test_memory_disabled_omits_env_keys` only covers a fresh install with memory off; it does not cover an existing conf with `MEMORY_ENABLED=true` where the user selects false. If `env` starts as a copy of `existing_env` anywhere in `_cmd_config`, this scenario would silently produce a stale `MEMORY_ENABLED=true` in the output. A fifth test seeding an existing conf with memory on and running the wizard with `"false"` for that prompt would close the gap.

No security issues, no logic errors in the tested paths, no critical or warning findings.

dcellison:
## Review by Kai

Looking at the diff with the context of the prior review thread.

Both prior suggestions are addressed: `AGENT_BACKEND: "claude"` is explicitly seeded in `test_memory_defaults_from_existing_install_conf`, and `test_memory_toggle_off_drops_existing_keys` covers the "was enabled, now disabled" scenario. The bonus `test_non_claude_backend_drops_stale_extraction_keys` is a solid addition.

**Suggestion - multi-line comment blocks in new tests (CLAUDE.md: "Never write... multi-line comment blocks - one short line max")**

Three new tests contain blocks the prior review flagged and had removed:

`test_memory_toggle_off_drops_existing_keys` (~line 1120):
```python
# Prior install had memory on with custom tunables. The wizard
# run below answers "false" to MEMORY_ENABLED. The output env
# must not retain any MEMORY_* keys - if it did, the daemon
# would silently keep memory live after the operator disabled it.
```

`test_memory_defaults_from_existing_install_conf` (expanded since last review):
```python
# Pre-seed install.conf with required fields plus memory tunables,
# then press Enter at every prompt...
# AGENT_BACKEND is seeded explicitly: the extraction-keys cleanup
# block pops MEMORY_EXTRACTION_* on non-claude backends...
```

`test_non_claude_backend_drops_stale_extraction_keys` (~line 1177):
```python
# Simulate a prior claude+extraction install. The wizard run
# below switches to goose; the extraction keys must not survive
# since bot.py:3609 silently ignores them on non-claude backends.
```

The content is legitimate WHY (non-obvious daemon behavior, AGENT_BACKEND invariant), but the format violates the project rule. Each block should collapse to one line. The `bot.py:3609` reference can stay as a single-line callout if needed.

No bugs, no security issues, no logic errors. The toggle-off test correctly exercises the non-obvious path the prior review identified as a gap.

dcellison:
## Review by Kai

Looking at the current diff against the prior review thread. All prior findings are resolved. One class of new findings:

**Suggestion - four multi-line comment blocks in `install.py` (CLAUDE.md: "Never write... multi-line comment blocks - one short line max")**

These were not raised in any prior round. All four appear in the install.py hunk:

`install.py` ~line 699 (section header block):
```python
# MEMORY_ENABLED is a single global toggle for Mem0 + Qdrant. Per-user
# partitioning happens at runtime via user_id = telegram_chat_id, so
# there is no users.yaml branching here. Without these prompts the
# wizard silently drops memory config on every reinstall (see #343).
```

`install.py` ~line 718 (Haiku extraction guard):
```python
# Haiku extraction only fires when the active backend is Claude
# (bot.py:3609 silently skips it otherwise - no startup error,
# no log line). Skip the prompt for non-claude backends rather
# than offer an option whose effect is invisible at runtime.
```

`install.py` ~line 836 (env-write block header):
```python
# Semantic memory: global env vars (per-user partitioning is runtime).
# Toggling memory_enabled from true back to false correctly drops
# MEMORY_* keys here, so the next /etc/kai/env reflects the new state.
# Numeric comparisons (not string) so "0.010" or "2000.0" are not
# treated as non-default and spuriously written.
```

`install.py` ~line 845 (stale-key cleanup):
```python
# Drop stale extraction keys when the backend isn't Claude. Mirrors
# the CLAUDE_MODEL/CLAUDE_MAX_BUDGET_USD cleanup above: bot.py:3609
# silently ignores these on non-claude backends, so leaving them in
# /etc/kai/env is misleading without effect. A user who flips backend
# from claude to goose should not see lingering extraction config.
```

The content is valid WHY reasoning, but each needs to collapse to one line. The `bot.py:3609` callout is worth keeping in the stale-key block.

The test file comment blocks from the prior round are all single-line in this diff - that's clean.

No bugs, no security issues, no logic errors. The fix is correct and the six-test suite covers the regression class thoroughly.

PR #346, opened 2026-04-20
fix(install): wizard prompts for MEMORY_EXTRACTION_TIMEOUT_S, MEMORY_SEARCH_LIMIT

### LINKED ISSUE #345: install wizard skips MEMORY_EXTRACTION_TIMEOUT_S, MEMORY_SEARCH_LIMIT prompts
PR #344 (closing #343) added wizard prompts for `MEMORY_ENABLED`, `MEMORY_EXTRACTION_ENABLED`, `MEMORY_EXTRACTION_BUDGET_USD`, and `MEMORY_TOKEN_BUDGET`, but explicitly skipped `MEMORY_EXTRACTION_TIMEOUT_S` and `MEMORY_SEARCH_LIMIT` with the rationale "operators wanting to override can hand-edit install.conf."

That rationale is wrong. `sudo make install` regenerates `/etc/kai/env` from `install.conf` entirely, so any value not captured by the wizard gets silently reset to its dataclass default on every reinstall. Same regression class as #343, just one layer deeper.

Confirmed reproduction (2026-04-20): production was tuned to `MEMORY_EXTRACTION_TIMEOUT_S=60` because extractions routinely take 20-30 seconds. After running `make config && sudo make install` on PR #344's merge, the timeout reset to the 10-second default and every extraction silently aborted at the timeout boundary.

## Scope

Add wizard prompts for the two operator-tunable vars under the existing `if memory_enabled:` guard in `_cmd_config()`:

| Env var | Default | Validation | Source |
|---|---|---|---|
| `MEMORY_EXTRACTION_TIMEOUT_S` | 10 | positive int | `config.py:461` |
| `MEMORY_SEARCH_LIMIT` | 10 | positive int | `config.py:439` |

Mirror the "if non-default, write" idiom already used for `MEMORY_TOKEN_BUDGET` and `MEMORY_EXTRACTION_BUDGET_USD`. Use numeric (not string) comparison so values like `"60"` vs `"60.0"` do not produce spurious env entries.

## Out of scope

`MEMORY_EMBEDDING_MODEL` and `MEMORY_EXTRACTION_MODEL` should NOT become wizard-tunable:
- Changing the embedding model breaks the existing Qdrant collection (different vector dimensions).
- The extraction model is code-coupled (Haiku-tuned prompts in `memory_extraction.py`).

These remain code-level decisions, not operator knobs. The wizard's silence on them is intentional.

## Tests

Round-trip tests mirroring the six added in PR #344:
- both vars omitted from `env` when `memory_enabled=false`
- both vars written to `env` when set to non-default values
- round-trip through `_generate_env_file()` and back
- defaults preserved when re-running wizard with existing `install.conf`

## Acceptance

After this fix, the only path to silently lose a `MEMORY_*` value across `sudo make install` should be the two model strings (intentional). All operator-tunable memory vars should survive reinstall.

dcellison:
Closes #345.

## Summary

Adds the two missing operator-tunable wizard prompts that #344 left out. After PR #344 merged, my own first reinstall silently reset `MEMORY_EXTRACTION_TIMEOUT_S` from 60 to 10, and every memory extraction (which routinely takes 20-30 seconds in production) immediately started timing out at the boundary. Same regression class as #343, just one layer deeper - the wizard fails to capture state that the install path then destroys.

## Changes

`src/kai/install.py`:
- `MEMORY_EXTRACTION_TIMEOUT_S` prompt nested under the existing `if memory_extraction_enabled:` guard. It is the LLM-call hard cap inside `memory_extraction.py:541`; irrelevant when extraction is off.
- `MEMORY_SEARCH_LIMIT` prompt at the top `if memory_enabled:` level. Retrieval (read path) runs regardless of extraction (write path), so this sits outside the extraction guard.
- Both use the validated `_validate_positive_int` loop and the "if non-default, write" idiom with numeric (not string) comparison.
- Non-claude cleanup pop also drops `MEMORY_EXTRACTION_TIMEOUT_S`, mirroring the existing pattern for `MEMORY_EXTRACTION_ENABLED` and `MEMORY_EXTRACTION_BUDGET_USD`.

`tests/test_install.py`:
- Existing six memory tests extended for the new prompt slots and seeded values.
- New `test_extraction_disabled_skips_timeout_prompt` guards the off-by-one trap: disabling extraction must consume exactly two fewer prompts (budget + timeout). If the gating drifts, the input iterator desynchronises and the wizard reads search limit as if it were the timeout.

## Out of scope

`MEMORY_EMBEDDING_MODEL` and `MEMORY_EXTRACTION_MODEL` stay non-tunable:
- Embedding-model changes break the existing Qdrant collection (different vector dimensions).
- Extraction-model is code-coupled (Haiku-tuned prompts in `memory_extraction.py`).

These are code-level decisions, not operator knobs. The wizard's silence on them is intentional.

## Test plan
- [x] `make check` (ruff + format)
- [x] `pytest tests/test_install.py::TestCmdConfig` - 13/13 pass (12 existing + 1 new)
- [x] Full suite passes outside the pre-existing `test_memory.py` failures (Qdrant lock held by the live daemon - same exclusion as #344, unrelated)

### DIFF
--- a/src/kai/install.py
+++ b/src/kai/install.py
@@ -741,7 +741,7 @@ def _cmd_config() -> None:
                 # rebuilds wholesale on every reinstall.
                 while True:
                     memory_extraction_timeout_s = _prompt(
-                        "Per-extraction timeout in seconds (suggest 60 or higher)",
+                        "Per-extraction timeout in seconds",
                         existing_env.get("MEMORY_EXTRACTION_TIMEOUT_S", "10"),
                     )
                     if _validate_positive_int(memory_extraction_timeout_s):
--- a/src/kai/install.py
+++ b/src/kai/install.py
@@ -710,7 +710,9 @@ def _cmd_config() -> None:
     # (or memory_enabled=true itself) are written to the env dict below.
     memory_extraction_enabled = False
     memory_extraction_budget_usd = "0.01"
+    memory_extraction_timeout_s = "10"
     memory_token_budget = "2000"
+    memory_search_limit = "10"
     if memory_enabled:
         # Haiku extraction only fires when the active backend is Claude
         # (bot.py:3609 silently skips it otherwise - no startup error,
@@ -730,12 +732,40 @@ def _cmd_config() -> None:
                     if _validate_positive_float(memory_extraction_budget_usd):
                         break
                     print("  Must be a positive number.")
+                # Extraction timeout is the LLM-call hard cap inside
+                # memory_extraction.py:541. Default 10s is too aggressive
+                # for production - real extractions routinely take 20-30s
+                # and silently abort at the boundary (#345). Operators
+                # MUST be able to raise this without hand-editing the
+                # generated install.conf, which `sudo make install`
+                # rebuilds wholesale on every reinstall.
+                while True:
+                    memory_extraction_timeout_s = _prompt(
+                        "Per-extraction timeout in seconds (suggest 60 or higher)",
+                        existing_env.get("MEMORY_EXTRACTION_TIMEOUT_S", "10"),
+                    )
+                    if _validate_positive_int(memory_extraction_timeout_s):
+                        break
+                    print("  Must be a positive integer.")
         while True:
             memory_token_budget = _prompt(
                 "Memory context token budget per turn",
                 existing_env.get("MEMORY_TOKEN_BUDGET", "2000"),
             )
             if _validate_positive_int(memory_token_budget):
+                break
+            print("  Must be a positive integer.")
+        # Search limit caps the retrieval set returned from Mem0/Qdrant
+        # before the token-budget filter. Lower values reduce noise but
+        # may miss relevant facts; higher values pay an embedding-search
+        # cost per turn. Operator-tunable for the same reason as the
+        # token budget: workload-specific, regression class #345.
+        while True:
+            memory_search_limit = _prompt(
+                "Memory search result limit per query",
+                existing_env.get("MEMORY_SEARCH_LIMIT", "10"),
+            )
+            if _validate_positive_int(memory_search_limit):
                 break
             print("  Must be a positive integer.")
     print()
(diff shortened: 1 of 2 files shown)


dcellison:
## Review by Kai

The PR is well-structured. Here are my findings:

**Warning - test coverage gap**

`test_memory_toggle_off_drops_existing_keys` seeds both new keys in the existing config but the diff does not show updated assertions confirming they are absent after the toggle-off. If the test's assertion block only checks the original keys, a regression where `MEMORY_SEARCH_LIMIT` or `MEMORY_EXTRACTION_TIMEOUT_S` survive a memory-disable would go undetected. Worth verifying the full assertion list in that test.

**Suggestion - misleading default UX**

`install.py:732` - the prompt reads "suggest 60 or higher" but defaults to `existing_env.get("MEMORY_EXTRACTION_TIMEOUT_S", "10")`. On a fresh install, the operator sees a suggestion of 60+ but the pre-filled default is 10. An operator who presses Enter accepts the very value the PR describes as too aggressive. Consider defaulting the fallback to "60" to align the hint with the default, or drop the parenthetical and let the range speak for itself.

**Suggestion - long inline comment**

`tests/test_install.py:1132` - the updated comment on the `# Switching claude -> goose...` line runs well past a readable column width. Split it into two comment lines.

**Everything else looks correct:**

- Nesting of the timeout prompt inside `if memory_extraction_enabled:` is right.
- `MEMORY_SEARCH_LIMIT` sitting outside that guard (but inside `if memory_enabled:`) is right.
- `int()` comparisons at write time are safe because `_validate_positive_int` already cleared the input.
- `MEMORY_SEARCH_LIMIT` correctly excluded from the non-claude cleanup (retrieval is backend-agnostic).
- The new `test_extraction_disabled_skips_timeout_prompt` test is a solid regression guard for iterator desync.

dcellison:
Re: finding #1 (test coverage gap) - the test at `test_install.py:1124` uses a loop assertion (`for key in env: assert not key.startswith("MEMORY_")`) which catches ALL `MEMORY_*` keys, including the newly seeded `MEMORY_EXTRACTION_TIMEOUT_S` and `MEMORY_SEARCH_LIMIT`. The seeding is the new coverage; the loop assertion is the gate. No regression on the new keys would slip through. Findings #2 (prompt UX) and #3 (long inline comment) addressed in 9c7cfbf.

dcellison:
## Review by Kai

Looking at the diff against the prior review thread to identify only new findings.

**Prior review status:**
- Comment line 1132: Addressed - the long line was correctly split into two.
- Default "10" vs "suggest 60+" mismatch: Not addressed, but author has seen it.
- `test_memory_toggle_off_drops_existing_keys` assertion gap: Not addressed, but author has seen it.

**New findings:**

**Suggestion - test/test_install.py, `test_extraction_disabled_skips_timeout_prompt` docstring**

Every other test in `TestCmdConfig` uses a single-line docstring (e.g., `"""Wizard-captured memory vars survive _generate_env_file()."""`). The new test's docstring runs to six lines across two paragraphs, which is inconsistent with the file's established style. The second paragraph restates information already present in the comment block and test name. Trimming to a single line (e.g., `"""Extraction off skips budget + timeout prompts; search limit is still asked."""`) and keeping the iterator-desync explanation as an inline comment above the assertion would match style.

**Everything else looks correct:**

- Prompt nesting (`MEMORY_EXTRACTION_TIMEOUT_S` inside extraction guard, `MEMORY_SEARCH_LIMIT` outside it) is right.
- `MEMORY_SEARCH_LIMIT` correctly excluded from the non-claude cleanup pop.
- `int()` comparisons at write time are safe - `_validate_positive_int` cleared the input already.
- Input sequences in all updated tests match the new prompt ordering.
- `test_non_claude_backend_drops_stale_extraction_keys` input sequence is correct: extraction + timeout prompts are skipped, token budget and search limit are consumed in the right order.
- `test_extraction_disabled_skips_timeout_prompt` is a solid regression guard for the iterator desync trap.

One style nit on the docstring; logic and coverage are sound.

PR #348, opened 2026-04-20
fix(memory): per-user MEMORY.md scoped by chat_id

### LINKED ISSUE #347: Per-user MEMORY.md under memory/<chat_id>/
Multi-user broke the MEMORY.md write path silently. `/var/lib/kai/memory/MEMORY.md` is single-global, owned by the service user mode 644. When Kai spawns inner Claude as a non-service `os_user` via `sudo -H -u <os_user>`, the subprocess can read the file but cannot write it. The subprocess reports this in-session as "I can't write to the memory file ... (owned by the kai user, I'm running as `<os_user>`)".

Mem0 facts and chat history are already per-user (`user_id=chat_id` filters at `memory.py:380/949/1038`, history at `history/<chat_id>/`). MEMORY.md is the only plaintext memory surface still single-global.

## Scope

Move MEMORY.md to `DATA_DIR/memory/<chat_id>/MEMORY.md`, mirroring the history layout.

- `backend.py:226` - build `memory_path` from `chat_id`, same scoping as `history_dir` on the line below.
- `main.py:93 _bootstrap_memory` - becomes lazy per-chat-id bootstrap invoked from the per-message path, not at startup. Startup cannot know the full chat_id set.
- `install.py` migration - existing `memory/MEMORY.md` relocates to the primary operator's chat_id dir (first user in `users.yaml`) on next `sudo make install`. One-time, idempotent.
- `install.py:1715` ownership - each `memory/<chat_id>/` subdir chowned to that user's `os_user`, resolved from `users.yaml`. `mem0_history.db` and `qdrant/` stay service-owned (the service writes them, not the subprocess).

## Out of scope

- Mem0/Qdrant: already per-user by code-layer filter; filesystem ownership unchanged.
- `history/<chat_id>/`: already per-user on disk.
- Cross-user memory sharing: not a feature; each user's MEMORY.md is theirs alone.

## Tests

- `_build_context` reads `memory/<chat_id>/MEMORY.md` for the given chat_id; does not read other users' files.
- Bootstrap creates the per-chat dir with correct ownership on first message.
- Migration moves legacy `memory/MEMORY.md` to the primary chat_id dir and leaves no stub.
- Two chat_ids get independent MEMORY.md files; writes to one do not appear in the other.

## Acceptance

Inner Claude running as any `os_user` can write its own chat's MEMORY.md. A user's notes do not appear in another user's context.

dcellison:
Closes #347.

## Summary
- MEMORY.md was the only memory surface still single-global. The path `/var/lib/kai/memory/MEMORY.md` is owned by the service user mode 644, so an inner Claude subprocess spawned via `sudo -H -u <os_user>` could read but not write it. The subprocess surfaced this in-session as "I can't write to the memory file."
- This PR scopes MEMORY.md per chat_id under `DATA_DIR/memory/<chat_id>/MEMORY.md`, mirroring the existing `history/<chat_id>/` layout. Each per-user subdirectory is chowned to the matching `os_user` from `users.yaml` at install time. The `qdrant/` tree and `mem0_history.db` stay service-owned because the service writes them, not the subprocess.
- Mem0 facts and chat history were already per-user (code-layer `user_id=chat_id` filters at `memory.py:380/949/1038`, history at `history/<chat_id>/`). MEMORY.md is now consistent with both.

## Changes
- `src/kai/backend.py` - `_build_context` derives `memory_path` from `chat_id`. Falls back to legacy `memory/MEMORY.md` when `chat_id is None` so `python -m kai` single-user dev continues to work unchanged.
- `src/kai/backend.py` - new `ensure_user_memory(chat_id, data_dir)` lazy bootstrap. Startup cannot know the full chat_id set in advance, so per-user MEMORY.md is created on first message instead.
- `src/kai/claude.py`, `src/kai/goose.py` - call `ensure_user_memory` before `build_session_context` in the per-message path.
- `src/kai/main.py` - removed `_bootstrap_memory()`. Startup no longer pre-creates any MEMORY.md.
- `src/kai/install.py` - new `_collect_user_memory_owners(users_yaml_path)` reads `users.yaml` and returns `[(chat_id, os_user)]`. `_apply_migrate` now: (1) copies source-tree `home/.claude/MEMORY.md` into the primary user's dir, (2) MOVES any legacy global `memory/MEMORY.md` into the primary's dir (move, not copy, so a stale global cannot shadow per-user reads), (3) seeds additional users from `MEMORY.md.example`, (4) chowns each `memory/<chat_id>/` subdir to the resolved `os_user`.

## Migration semantics
- First-ever install (no `users.yaml`): no-op. Runtime falls back to legacy global path.
- Existing single-global install: legacy `memory/MEMORY.md` is MOVED to the primary user's subdirectory. Move (not copy) is required; a stale global file would shadow per-user reads once additional users join.
- Reinstall: idempotent. Existing per-user MEMORY.md files are not overwritten.
- Multi-user: additional users beyond primary get a fresh `MEMORY.md.example`-seeded file. The primary's notes never become another user's starting state.

## Tests
- `tests/test_backend.py` - three new classes covering per-user read scoping, lazy bootstrap (creation, idempotence, OSError tolerance, fallback when chat_id is None), and cross-user isolation (writes to one chat_id never appear in another).
- `tests/test_install.py` - `TestCollectUserMemoryOwners` (8 tests for users.yaml parsing edge cases) and `TestApplyMigratePerUserMemory` (5 tests for the migration paths above). The new migration class deliberately lives outside `TestApplyMigrate` so its `_isolate_users_yaml` autouse fixture (which stubs `_collect_user_memory_owners` to `[]`) does not neuter them.
- `tests/test_main.py` - removed `TestBootstrapMemory` since `_bootstrap_memory()` no longer exists.
- Local: 2060 pytest pass; ruff check + format clean. The 2 failures + 5 errors in `test_memory.py` are a pre-existing environmental flake (the running Kai service holds an exclusive flock on the global `~/.mem0/migrations_qdrant/.lock`); none of those tests reference symbols this PR touches and they fail in isolation against `main`.

## Test plan
- [ ] CI green
- [ ] `sudo make install` on a host with multi-user `users.yaml`: confirm `memory/<chat_id>/MEMORY.md` exists per user and is owned by the correct `os_user`
- [ ] Inner Claude as a non-service `os_user` can write its own chat's MEMORY.md
- [ ] A user's notes do not appear in another user's session context

### DIFF
--- a/src/kai/backend.py
+++ b/src/kai/backend.py
@@ -15,12 +15,13 @@ prompt assembly that is identical across all backends.
 """
 
 import logging
+import shutil
 from abc import ABC, abstractmethod
 from collections.abc import AsyncIterator
 from dataclasses import dataclass, field
 from pathlib import Path
 
-from kai.config import WorkspaceConfig
+from kai.config import PROJECT_ROOT, WorkspaceConfig
 from kai.history import get_recent_history
 
 log = logging.getLogger(__name__)
@@ -223,7 +224,20 @@ def build_session_context(
     # so it survives make install. Available regardless of which
     # workspace the inner Claude is operating in. try/except guards
     # against race and permission errors (same pattern as identity).
-    memory_path = data_dir / "memory" / "MEMORY.md"
+    #
+    # Per-user scoping (#347): when chat_id is set, the file lives under
+    # memory/<chat_id>/MEMORY.md so each user has their own writable
+    # copy. The inner Claude subprocess runs as that user's os_user
+    # (via sudo -H -u), so ownership of the subdirectory is set to
+    # match. A non-service user cannot write the legacy single-global
+    # file, which was the bug this scoping fixes. Falls back to the
+    # legacy global path when chat_id is None (local-dev backends
+    # with no multi-user config) so nothing regresses on single-user
+    # setups that never hit the multi-user pool path.
+    if chat_id is not None:
+        memory_path = data_dir / "memory" / str(chat_id) / "MEMORY.md"
+    else:
+        memory_path = data_dir / "memory" / "MEMORY.md"
     try:
         memory = memory_path.read_text().strip()
         if memory:
@@ -344,6 +358,60 @@ def build_session_context(
     # No trailing \n\n here - prepend_to_prompt() adds the separator
     # between the context block and the user's message.
     return "\n\n".join(parts)
+
+
+def ensure_user_memory(chat_id: int | None, data_dir: Path) -> None:
+    """
+    Lazily create the per-user MEMORY.md directory and seed file.
+
+    Idempotent and cheap: a stat, a possible mkdir, a possible copy2.
+    Called on every send() path before build_session_context() so that
+    a user who appeared after install (e.g., a new entry in users.yaml
+    without a reinstall) still gets a writable memory surface on their
+    first message.
+
+    In production the install step (install.py ~line 1715) pre-creates
+    and chowns `memory/<chat_id>/` to that user's os_user, so mkdir
+    here is a no-op. For dev/test runs without a prior install, this
+    function does the whole job - mkdir + seed copy - inheriting the
+    caller's ownership. That matches the single-user dev case where
+    the bot process is the only identity in play.
+
+    Intentionally no-ops when chat_id is None. That path reads and
+    writes `memory/MEMORY.md` directly (legacy global location) for
+    backends that never see a chat_id - dev harnesses, smoke scripts.
+
+    Silent on OSError: a permissions issue creating per-user memory
+    must not prevent the bot from answering a message. The read path
+    in build_session_context already returns a placeholder string on
+    missing files, so the downstream session still boots cleanly.
+    """
+    if chat_id is None:
+        return
+
+    user_memory_dir = data_dir / "memory" / str(chat_id)
+    user_memory_file = user_memory_dir / "MEMORY.md"
+
+    try:
+        user_memory_dir.mkdir(parents=True, exist_ok=True)
+        if not user_memory_file.exists():
+            # Seed from the example template if one ships with the
+            # source tree. Copy2 preserves mode bits so a deliberately
+            # permissive example stays that way.
+            example = PROJECT_ROOT / "home" / ".claude" / "MEMORY.md.example"
+            if example.is_file():
+                shutil.copy2(example, user_memory_file)
+            else:
+                # No template available (unusual - only happens when the
+                # install tree is incomplete). Create a minimal placeholder
+                # so the subprocess has a writable file to edit.
+                user_memory_file.write_text("# Memory\n")
+    except OSError:
+        log.warning(
+            "ensure_user_memory: could not bootstrap %s",
+            user_memory_file,
+            exc_info=True,
+        )
 
 
 def build_foreign_workspace_reminder(workspace: Path, home_workspace: Path) -> str | None:
--- a/src/kai/backend.py
+++ b/src/kai/backend.py
@@ -388,20 +388,32 @@ def ensure_user_memory(chat_id: int | None, data_dir: Path) -> None:
       bootstrap does the whole job - mkdir + seed copy - inheriting
       the caller's ownership, which IS the writer. Good.
 
-    Intentionally no-ops when chat_id is None. That path reads and
-    writes `memory/MEMORY.md` directly (legacy global location) for
-    backends that never see a chat_id - dev harnesses, smoke scripts.
+    When chat_id is None (legacy/dev path - no users.yaml, no real
+    Telegram session), the function ensures `data_dir/memory/MEMORY.md`
+    exists with the same seed-or-placeholder semantics. This mirrors
+    the removed main._bootstrap_memory() so the legacy global path
+    continues to be writable from a fresh install where memory is
+    disabled and users.yaml is absent.
 
     Silent on OSError: a permissions issue creating per-user memory
     must not prevent the bot from answering a message. The read path
     in build_session_context already returns a placeholder string on
     missing files, so the downstream session still boots cleanly.
     """
+    # When chat_id is None we are on the legacy/dev path: there is no
+    # users.yaml to drive a per-user subdir, so the file lives at
+    # data_dir/memory/MEMORY.md. This mirrors the behavior of the
+    # removed main._bootstrap_memory() function so a fresh
+    # `python -m kai` (no users.yaml, no prior install, memory disabled)
+    # still has a writable memory_root for the inner Claude to update.
+    # Without this branch a write attempt from the subprocess would
+    # FileNotFoundError on the missing parent directory.
     if chat_id is None:
-        return
-
-    user_memory_dir = data_dir / "memory" / str(chat_id)
-    user_memory_file = user_memory_dir / "MEMORY.md"
+        user_memory_dir = data_dir / "memory"
+        user_memory_file = user_memory_dir / "MEMORY.md"
+    else:
+        user_memory_dir = data_dir / "memory" / str(chat_id)
+        user_memory_file = user_memory_dir / "MEMORY.md"
 
     try:
         user_memory_dir.mkdir(parents=True, exist_ok=True)
--- a/src/kai/backend.py
+++ b/src/kai/backend.py
@@ -366,16 +366,27 @@ def ensure_user_memory(chat_id: int | None, data_dir: Path) -> None:
 
     Idempotent and cheap: a stat, a possible mkdir, a possible copy2.
     Called on every send() path before build_session_context() so that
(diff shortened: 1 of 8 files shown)


dcellison:
## Review by Kai

## PR Review: fix(memory): per-user MEMORY.md scoped by chat_id

Overall the PR is well-structured, the migration semantics are sound, and the test coverage is thorough. Three actionable findings below.

---

### Warning 1 - `ensure_user_memory` overclaims ownership fix for post-install users

**`src/kai/backend.py`, `ensure_user_memory` docstring (~line 363)**

The docstring states: *"a user who appeared after install (e.g., a new entry in users.yaml without a reinstall) still gets a writable memory surface on their first message."*

This is false when `os_user != service_user`. `user_memory_dir.mkdir()` inherits the calling process's identity (the service user). The inner Claude subprocess runs as `os_user` via `sudo -H -u`, so the newly-created directory is still service-owned and the write still fails - the exact bug #347 reports. Only the `install.py` chown path correctly fixes ownership. Post-install add-user without a reinstall remains broken for any user with an explicit `os_user`.

The docstring should be corrected to state that lazy bootstrap only fixes the dev case (no `os_user` separation), and that the multi-user production case requires a reinstall.

---

### Warning 2 - `pwd.getpwnam` fails late with a bare `KeyError` after files are already written

**`src/kai/install.py`, `_apply_migrate`, ownership section (~line 1845 in the post-diff file)**

```python
for chat_id, os_user in memory_owners:
    if os_user is None:
        continue
    pwd_entry = pwd.getpwnam(os_user)          # raises KeyError, not ValueError
    per_user_ids[str(chat_id)] = (pwd_entry.pw_uid, pwd_entry.pw_gid)
```

`per_user_ids` is built in the ownership block, which runs **after** the MEMORY.md migration block has already created and possibly moved files. If a `users.yaml` entry names an OS user that doesn't exist on the host, `getpwnam` raises a bare `KeyError("<username>")` with no mention of `users.yaml`, the chat_id, or the fact that files were already modified.

Two fixes needed:
1. Move the `per_user_ids` construction to **before** the MEMORY.md migration block so a misconfiguration hard-fails before touching disk.
2. Wrap `pwd.getpwnam` in a `try/except KeyError` and re-raise as `ValueError` with a message like `f"os_user {os_user!r} (chat_id={chat_id}) in {users_yaml_path} does not exist on this host"`.

---

### Suggestion - Dry-run for secondary users doesn't simulate the missing-template branch

**`src/kai/install.py`, secondary users loop in `_apply_migrate` (~line 1795)**

```python
if dry_run:
    print(f"[DRY RUN] Would seed {user_dst} from example template")
    continue
```

The real path falls back to `user_dst.write_text("# Memory\n")` when `example_template` doesn't exist. The dry-run always prints "from example template" regardless, so an operator on a host missing the template gets misleading output. Minor - add a conditional on `example_template.is_file()` to match the real branch.

---

### Suggestion - `ensure_user_memory` is never retried after a silent OSError

**`src/kai/claude.py:409`, `src/kai/goose.py:353`**

The call is inside `if self._fresh_session:` which flips to `False` immediately after. A transient permission error on the first message silently logs a warning, and no subsequent message will attempt creation again until the session object is recreated. Low risk in practice, but worth a code comment acknowledging this limitation so future readers don't assume it self-heals.

---

The test design (especially the `_isolate_users_yaml` autouse fixture and the deliberate separation of `TestApplyMigratePerUserMemory`) is clean and the isolation rationale is well-documented. No issues with the test logic itself.

dcellison:
## Review by Kai

All three prior warnings and both suggestions are addressed in this revision. Two new findings:

---

### Warning - `DATA_DIR/memory/` never created for single-user/no-`users.yaml` installs

**`src/kai/main.py`** (removal of `_bootstrap_memory`) + **`src/kai/backend.py`** (`ensure_user_memory` no-ops on `chat_id=None`)

`_bootstrap_memory()` was the only call that created `DATA_DIR/memory/` for fresh single-user installations. The PR removes it and replaces it with `ensure_user_memory`, which explicitly no-ops when `chat_id is None`. `_apply_migrate` only creates `memory/<chat_id>/` when `users.yaml` has entries - when `memory_owners` is empty, the `if primary_chat_id is not None:` block is skipped and `memory_root` is never `mkdir`-ed.

Result for a fresh `python -m kai` dev box with no `users.yaml`:
- `_apply_migrate` creates nothing under `memory/`
- At runtime `build_session_context` resolves to `data_dir / "memory" / "MEMORY.md"`, reads a placeholder (the directory doesn't exist)
- If the inner Claude tries to `write_text` its memory updates, the `FileNotFoundError` on the non-existent parent will propagate

The PR description claims "single-user dev continues to work unchanged" - that claim requires `DATA_DIR/memory/` to exist before the first run. Either `init_memory()` (not visible in the diff) creates it as a side effect, or this is a regression. If `init_memory` doesn't create `memory/`, a minimal `memory_root.mkdir(parents=True, exist_ok=True)` needs to survive somewhere for the legacy path.

---

### Suggestion - Dry-run primary user has a silent no-op branch

**`src/kai/install.py`**, `_apply_migrate`, primary user dry-run block (~line 1780 in post-diff)

When `primary_chat_id is not None` and `not primary_dst.exists()` but none of `legacy_global`, `legacy_src_tree`, or `example_template` exist, the `if dry_run:` block prints nothing - it falls off the bottom of the `elif` chain. The real (non-dry-run) path hits `primary_dst.write_text("# Memory\n")` and prints `"Created empty ..."`. An operator running dry-run on a minimal host gets silence for this branch. Add an `else: print(f"[DRY RUN] Would create empty {primary_dst} (no template found)")` to match.

---

Everything else looks correct. The `per_user_ids` validation hoisted before disk mutation, the move-not-copy semantics for `legacy_global`, the autouse fixture isolation in tests, and the secondary-user dry-run template-presence check all look sound.

dcellison:
## Review by Kai

This is the third review pass. All prior warnings and suggestions have been addressed. Here are my findings on the current diff:

---

### Clean

**Migration semantics** are correct end-to-end: `legacy_global` is moved (not copied), `legacy_src_tree` is copied, secondary users get the template only, and `primary_dst.exists()` guards idempotence. The ownership tier split (service-owned root, per-user subdirs) is correctly implemented.

**`ensure_user_memory` `chat_id=None` branch** properly replaces the removed `_bootstrap_memory()` for the dev path. The docstring accurately scopes what this function can and cannot fix (post-install add-user with a distinct `os_user` still requires a reinstall).

**Validation hoisting** in `_apply_migrate` correctly aborts with a clear `ValueError` before any disk mutation when `pwd.getpwnam` fails. The test `test_unknown_os_user_aborts_before_disk_mutation` guards against regressions.

**`_isolate_users_yaml` autouse fixture** in `TestApplyMigrate` correctly prevents the per-user migration from running in tests that don't care about it, and `TestApplyMigratePerUserMemory` is correctly placed outside that class to bypass it.

---

### Suggestion - Loop variable `chat_id` reused across two unrelated loops in `_apply_migrate`

`install.py`, `_apply_migrate`

```python
for chat_id, os_user in memory_owners:          # per_user_ids construction
    ...

for chat_id, _os_user in memory_owners[1:]:     # secondary users seeding
    ...
```

Both loops bind the name `chat_id` in the same function scope. There is no actual bug since neither loop's residual value is read after its own block, but it is visually misleading - a reader skimming the secondary loop could momentarily wonder whether `chat_id` from the first loop is intended. Using a distinct name like `sec_chat_id` in one of the loops would remove the ambiguity for no cost.

---

No critical issues, no security issues, no logic errors. The implementation is correct and the test coverage is thorough.

dcellison:
## Review by Kai

This is the fourth pass. All findings from prior reviews are addressed.

### Clean

**Migration semantics** remain correct: validation hoisted before disk mutation, `legacy_global` moved not copied, `legacy_src_tree` copied, secondary users seeded from template only, dry-run has a complete `else` branch for the no-template case (including the primary path).

**`ensure_user_memory` `chat_id=None` branch** correctly replaces the removed `_bootstrap_memory()`. The docstring accurately scopes what lazy bootstrap can and cannot fix for post-install add-user cases.

**`_collect_user_memory_owners`** validation is consistent with `_collect_os_users_from_yaml`: missing/empty/malformed files return `[]`, invalid `os_user` raises `ValueError`, non-int `telegram_id` skipped silently.

**Test coverage** is thorough: `test_unknown_os_user_aborts_before_disk_mutation` correctly guards the hoisted-validation invariant, `test_moves_legacy_global_memory_to_primary` asserts `not legacy_global.exists()` (verifying move semantics), and the `_isolate_users_yaml` autouse fixture is correctly scoped to `TestApplyMigrate` only.

**No new bugs, security issues, or logic errors found.** The loop variable `chat_id` reused across the two `memory_owners` loops in `_apply_migrate` (prior suggestion, third pass) remains unaddressed but is cosmetic and was the author's call.

PR #352, opened 2026-04-20
fix: drop redundant tag-row text from /memory dashboard

### LINKED ISSUE #351: Drop redundant tag-row text from /memory dashboard
## Summary

The `/memory` dashboard renders each tag's count in two places: a text bar-chart block and the inline keyboard buttons immediately below. The text rows duplicate what the buttons already show, so the dashboard is roughly twice as tall as it needs to be without giving the user any data they couldn't read off the buttons.

## Current behavior

```
Memories: 73 facts across 8 tags.

  project            18   ▓▓▓▓▓▓▓▓
  fact               16   ▓▓▓▓▓▓▓
  preference         15   ▓▓▓▓▓▓▓
  constraint         10   ▓▓▓▓
  decision            9   ▓▓▓▓
  location            3   ▓
  confirmed_action    1   ▓
  relationship        1   ▓

Tap a tag to browse. Confidence: median 0.86, min 0.52.
[project (18)] [fact (16)] [preference (15)] [constraint (10)] ...
[Search] [Stats]
```

The counts in the bar-chart rows (`18`, `16`, `15`, ...) are the same counts shown in the button labels (`(18)`, `(16)`, `(15)`, ...).

## Proposed change

Drop the per-tag text rows. Keep the summary line, the confidence footer, and the inline keyboard exactly as they are.

```
Memories: 73 facts across 8 tags.

Tap a tag to browse. Confidence: median 0.86, min 0.52.
[project (18)] [fact (16)] [preference (15)] [constraint (10)] ...
[Search] [Stats]
```

## Tradeoff

We lose the visual bar chart, which gave at-a-glance relative magnitude. The button labels still carry the counts in parentheses so the relative magnitudes remain readable; only the bar visualization itself is gone. Telegram's inline buttons are short text only, so there is no equivalent surface for the bars - they exist purely because the buttons can't render them.

## Cleanup

`_bar`, `_BAR_CHAR`, and `_BAR_WIDTH` in `memory_command.py` become unused once the rows are gone. Their tests in `TestBar` (`test_zero_count_empty_bar`, `test_max_count_full_width`, `test_nonzero_min_one_block`) become unnecessary. The `test_renders_summary_and_tags` test asserts the order of tag rows in the dashboard text - it needs to be reworked to assert the order of buttons instead.

## Spec drift

Spec 310 §6.1 specified the bar-chart block as part of the dashboard mock-up. This issue intentionally diverges from that part of the spec based on operator feedback during real use. The other §6.1 elements (summary header, confidence footer, asymmetry between dashboard and stats around zero-count tags) are unchanged.

## Related

- #306 / spec 310 - original dashboard design
- PR #337 - dashboard implementation

dcellison:
Closes #351.

## Summary

The `/memory` dashboard renders each tag's count in two places: a text bar-chart block and the inline keyboard buttons below. The text rows duplicate what the buttons already show, so the dashboard is roughly twice as tall as needed. Operator feedback on real data confirmed the bars added no value the button labels weren't already carrying through their `(N)` suffixes.

## Changes

**`src/kai/memory_command.py`:**
- `_build_dashboard` no longer emits the per-tag rows. The text now stops at the summary header (`Memories: N facts across M tags.`) and the confidence footer (`Tap a tag to browse. Confidence: median X, min Y.`).
- `_bar`, `_BAR_CHAR`, `_BAR_WIDTH` removed - unused after the rows are gone.
- `_build_stats` is unchanged; it has its own per-tag rendering that never called `_bar`.

**`tests/test_memory_command.py`:**
- `TestBar` (3 tests) deleted along with `_bar`.
- `test_renders_summary_and_tags` reworked: asserts the bar character (▓) is gone, the per-tag row prefixes are gone, and the buttons carry the counts in descending order (`preference (5)`, `fact (3)`, `decision (2)`).

Net diff: +23 / -56 (-33 lines).

## Before / after

Before:
```
Memories: 73 facts across 8 tags.

  project            18   ▓▓▓▓▓▓▓▓
  fact               16   ▓▓▓▓▓▓▓
  preference         15   ▓▓▓▓▓▓▓
  constraint         10   ▓▓▓▓
  decision            9   ▓▓▓▓
  location            3   ▓
  confirmed_action    1   ▓
  relationship        1   ▓

Tap a tag to browse. Confidence: median 0.86, min 0.52.
[project (18)] [fact (16)] ...
[Search] [Stats]
```

After:
```
Memories: 73 facts across 8 tags.

Tap a tag to browse. Confidence: median 0.86, min 0.52.
[project (18)] [fact (16)] ...
[Search] [Stats]
```

## Spec drift

This intentionally diverges from spec 310 §6.1's bar-chart mock-up. All other §6.1 elements (summary header, confidence footer, asymmetry between dashboard and stats around zero-count tags) are preserved. Issue #351 documents the rationale.

## Test plan

- [x] `make check` passes (ruff check + ruff format).
- [x] `pytest tests/test_memory_command.py` - 83 passed.
- [ ] After deploy, send `/memory` and confirm the dashboard renders without the per-tag rows but with summary + footer + tag buttons + Search/Stats.
- [ ] Confirm `/memory stats` still renders its full per-tag table (not affected by this change).

## Related

- #351 - feature request
- #306 / spec 310 - original dashboard design (§6.1 bar chart now superseded)
- PR #337 - dashboard implementation

### DIFF
--- a/src/kai/memory_command.py
+++ b/src/kai/memory_command.py
@@ -101,17 +101,6 @@ _LIST_TRUNCATE_LEN = 80
 # hard ceiling on session duration.
 _CACHE_TTL_S = 30 * 60
 
-# Unicode block character for the dashboard bar chart (spec §6.1
-# mock-up uses ▓). Plain text mode renders this fine on every
-# Telegram client.
-_BAR_CHAR = "\u2593"
-
-# Width of the dashboard bar chart in characters, scaled to the
-# largest tag count. Eight is wide enough to show contrast between
-# 38 and 5 without overflowing the typical phone screen.
-_BAR_WIDTH = 8
-
-
 # ── Per-chat navigation cache ───────────────────────────────────────
 
 
@@ -315,20 +304,6 @@ def _format_date(iso_str: str, with_time: bool = False) -> str:
     return iso_str[:10]
 
 
-def _bar(count: int, max_count: int, width: int = _BAR_WIDTH) -> str:
-    """Render a unicode-block bar of length proportional to count.
-
-    A bar is at least one block wide for any nonzero count so a tag
-    with a single fact is visually distinguishable from a tag with
-    zero facts (which is hidden from the dashboard anyway, but the
-    builder is also reused for stats).
-    """
-    if max_count <= 0 or count <= 0:
-        return ""
-    n = max(1, round((count / max_count) * width))
-    return _BAR_CHAR * n
-
-
 # ── Builder: dashboard ──────────────────────────────────────────────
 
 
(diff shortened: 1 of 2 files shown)


dcellison:
## Review by Kai

The PR looks clean. Here is what I checked:

**Logic (no issues)**

The removal of `max_count = nonzero[0][1]` and `label_width = ...` is safe - both variables were only used by the deleted loop. The early-return guard for `nonzero` being empty was already in place before those lines, so no new IndexError risk is introduced.

The blank line between summary and footer is preserved: the `lines.append("")` that follows `lines.append(summary)` is context, not a deletion. Output matches the "after" example in the description.

**Security (no issues)**

Purely a UI text change. No user input processing, auth, or data exposure changes.

**Tests (well-done)**

The comment in the updated test explaining why the assertion uses `"  {tag}"` prefix instead of bare tag names is exactly the right kind of "non-obvious why" comment. The `tag_button_labels` assertion gives clear ordering coverage.

**Style (no issues)**

The in-code comment block explaining the spec drift from §6.1 is long but justified - intentional spec divergence with a documented rationale is exactly the scenario that warrants the explanation.

**One minor note (suggestion)**

The comment added to `_build_dashboard` mentions issue #351 and spec §6.1 twice (once at the comment, once again in the PR description). The comment itself is slightly redundant ("Per #351... operator feedback on real data showed the bars duplicated what the buttons already telegraph through the parenthesized counts"). Could trim to the core fact, but it is not wrong, and the project style favors thorough comments for non-obvious decisions.

No blockers. Approve.

PR #355, opened 2026-04-21
feat: per-user home workspace under DATA_DIR

### LINKED ISSUE #353: Replace shared /opt/kai/home/ with per-user home workspaces
## Summary

Replace the shared `/opt/kai/home/` default workspace with a per-user directory at `DATA_DIR/home/<chat_id>/`. `/workspace home` resolves per-user: first to `users.yaml` if set, then to the chat_id-keyed Kai-managed directory. No user ever lands in a directory shared with other users.

## Current behavior

- `config.claude_workspace` (src/kai/config.py:339) defaults to `PROJECT_ROOT / "home"`, which resolves to `/opt/kai/home/` in production.
- Session landing (src/kai/pool.py:113) uses `user.home_workspace` when set, else falls through to `claude_workspace`.
- `/workspace home` (src/kai/bot.py:2034, 2052) ignores `user.home_workspace` entirely. It always sends the user to `config.claude_workspace`, so multi-user installs pile everyone into the same shared directory.

## Problems

1. Multiple users share one filesystem location. Content written by one user is visible to every other user. The name "home" reads as "personal" to anyone with Unix intuition, making the shared reality a surprise.
2. `home_workspace` in users.yaml is partially honored. Pool init reads it; `/workspace home` does not. Two sources of truth that silently disagree.
3. `/opt/kai/` is wiped and rewritten by `sudo make install`. Anything under `/opt/kai/home/` is unsafe to treat as durable user data.
4. Telegram-only users (no `os_user` in users.yaml, subprocess runs as the bot user) have no non-shared landing spot without an admin setting `home_workspace` by hand for every user.

## Proposed change

Introduce a per-user workspace root that matches the existing chat_id-keyed pattern under DATA_DIR:

```
DATA_DIR/
  files/<chat_id>/       (already exists)
  history/<chat_id>/     (already exists)
  memory/<chat_id>/      (per-user as of PR #348)
  home/<chat_id>/        NEW
```

Resolution order for both session landing and `/workspace home`:

1. `user.home_workspace` from users.yaml if set.
2. `DATA_DIR/home/<chat_id>/` - auto-created lazily on first access (same bootstrap pattern as memory/ in #348), owned by the bot user, mode 0700.

Remove `config.claude_workspace` as a fallback. Drop the default of `PROJECT_ROOT / "home"` entirely.

## Scope

- `src/kai/config.py`: drop `claude_workspace` default; decide whether the field stays (e.g., for dev PROJECT_ROOT-relative workflows) or goes entirely.
- `src/kai/pool.py:113-114`: update fallback chain to use `DATA_DIR/home/<chat_id>/` instead of `claude_workspace`. Lazy bootstrap on first access.
- `src/kai/bot.py:2034, 2052`: `/workspace home` reads `user.home_workspace` first, then the per-user default.
- `src/kai/install.py`: do not create or reference `/opt/kai/home/`. Installer ensures `/var/lib/kai/home/` exists with bot-user ownership.
- Tests: adjust any that assumed the shared default.
- Docs: README, wiki multi-user setup, and home CLAUDE.md references.

## Out of scope

- Renaming `home_workspace`. The name is accurate once the shared fallback is gone.
- The `home/` directory inside the dev repo (per-developer spec/scratch convention) is not an install artifact and is untouched.

## Migration

Existing installs may have real user content under `/opt/kai/home/`. The upgrade does not auto-delete. Release notes instruct operators to:

1. Inspect `/opt/kai/home/` for anything worth keeping.
2. Move kept content into `/var/lib/kai/home/<chat_id>/` (or a specific user's `home_workspace`).
3. Remove `/opt/kai/home/` manually.

dcellison:
Closes #353.

## Summary

- Replaces the shared `/opt/kai/home/` fallback with per-chat_id `DATA_DIR/home/<chat_id>/`, closing the multi-user privacy hazard
- Honors `user.home_workspace` from `users.yaml` when explicitly set (opt-out from Kai-managed directory)
- Mirrors the per-user MEMORY.md migration pattern from PR #348

## Changes

**Backend** (`src/kai/backend.py`)
- `ensure_user_home(chat_id, data_dir)` - lazy `mkdir` for `DATA_DIR/home/<chat_id>/`, idempotent, 0755
- `resolve_home_workspace(chat_id, config)` - prefers `users.yaml` override, falls back to `ensure_user_home`
- Single resolver used by `pool.py` and `bot.py`

**Config** (`src/kai/config.py`)
- Removes `claude_workspace` field from `Config`; validator logs "falling back to per-user default" instead of the old global path

**Pool / bot / webhook**
- `pool.py`: 3 call sites now use `resolve_home_workspace(chat_id, self._config)`
- `bot.py`: 5 call sites (including `/workspace home`, `/workspace` keyboard) use the resolver
- `webhook.py`: `_app["workspace"]` and `update_workspace()` removed entirely (post-review cleanup - see below). Send-file confinement reads `pool.get_workspace(chat_id)` at request time; `_resolve_local_repo` resolves GitHub webhook repos via WORKSPACE_BASE / ALLOWED_WORKSPACES / workspace_history

**Install wizard** (`src/kai/install.py`)
- Drops `home_workspace` prompt (admin lands in per-user directory like everyone else)
- `_apply_migrate` creates each user's `home/<chat_id>/` slot with correct ownership, skipping overrides that pin a path outside `DATA_DIR`
- `_collect_user_home_overrides()` helper reads `users.yaml` to detect opt-outs

**Tests**
- `test_backend.py`: `TestEnsureUserHome` + `TestResolveHomeWorkspace` (idempotency, None chat_id, users.yaml precedence)
- `test_bot.py`: `test_workspace_home_lands_in_per_user_directory` - pins the multi-user privacy guarantee at the `/workspace home` flow
- `test_install.py`: `test_wizard_defaults_do_not_reintroduce_shared_home` - regression guard across both basic and advanced wizard paths
- 1948 tests pass

## Migration steps (release notes)

Existing installations have content under `/opt/kai/home/` (or equivalent `PROJECT_ROOT/home/`). After deploying this change, that directory is **no longer read by Kai at runtime**. Operators should:

1. **Stop the service**: `sudo launchctl bootout system/com.syrinx.kai` (macOS) or systemd equivalent.
2. **Decide per user whether to migrate content**:
   - For each active `chat_id`, move desired files from `/opt/kai/home/` into `/var/lib/kai/home/<chat_id>/`.
   - `sudo chown -R <os_user>:<os_user> /var/lib/kai/home/<chat_id>` if that user has a distinct `os_user` in `users.yaml`; otherwise leave service-owned.
   - Users who want a workspace outside `DATA_DIR` (a clone of a dev tree, a synced volume, etc.) can hand-add `home_workspace: /absolute/path` to their `users.yaml` entry.
3. **Reinstall**: `sudo make install`. The `_apply_migrate` step will lazily create any still-missing `home/<chat_id>/` directories for known users.
4. **Restart the service**.
5. **Archive or remove** `/opt/kai/home/` once you've confirmed nothing important remained. It will not be re-created by Kai.

Users with `home_workspace` already set in `users.yaml` are unaffected - they continue to use their pinned path.

## Post-review cleanup (v6)

After the automated reviewer approved at round 5, an independent audit found that `webhook.update_workspace()` had been silently decoupled from `_do_switch_workspace` earlier in the spec revision history but left in the codebase for backward compatibility. With `_do_switch_workspace` no longer calling it and send-file now reading `pool.get_workspace(chat_id)` at request time, the `_app["workspace"]` slot and the `update_workspace()` setter contributed to nothing - including to the "load-bearing for send-file confinement" rationale that v5 cited for keeping them. A state labeled "update_workspace looks live but never runs, and a spec that describes behavior the code no longer has" is the worst of both worlds.

v6 removes:

- `update_workspace()` function definition in `webhook.py`
- `_app["workspace"] = None` initialization
- `_resolve_local_repo` step 1 (home-workspace-via-parent heuristic - only fired when the slot was populated, which it never is post-v6)
- The `request.app.get("workspace")` fallback in the send-file handler (redundant: pool-based path already covers every real request)
- `TestUpdateWorkspace` test class, `test_home_workspace` / `test_priority_order` from `TestResolveLocalRepo`, and 11 now-unused `patch("kai.bot.webhook.update_workspace")` guards in `test_bot.py`

Net diff: 22 insertions, 146 deletions. No user-visible behavior change.

## Test plan

- [x] `make test` - 1948 passing
- [ ] Deploy to Mac mini, verify `/workspace home` for admin lands in `/var/lib/kai/home/<admin_chat_id>/`
- [ ] Verify a second user's `/workspace home` lands in their own `home/<chat_id>/`, not the admin's
- [ ] Confirm `users.yaml` `home_workspace` overrides still route correctly
- [ ] `sudo make install` twice in a row is idempotent (no duplicate chown churn)

### DIFF
--- a/src/kai/backend.py
+++ b/src/kai/backend.py
@@ -15,6 +15,7 @@ prompt assembly that is identical across all backends.
 """
 
 import logging
+import os
 import shutil
 from abc import ABC, abstractmethod
 from collections.abc import AsyncIterator
@@ -498,17 +499,28 @@ def ensure_user_home(chat_id: int | None, data_dir: Path) -> Path:
     # but dev paths and the tests go straight through lazy bootstrap).
     # exist_ok=True means this is safe to call on every session init.
     path.mkdir(parents=True, exist_ok=True, mode=0o755)
+    # mkdir's mode= argument is masked by the process umask, so on a
+    # service configured with umask 0o027 the directory can end up 0o750
+    # instead of 0o755. That blocks group traversal and can break the
+    # inner subprocess when sudo -u targets a different identity. Force
+    # the intended bits explicitly - this is the same pattern install.py
+    # `_apply_migrate` uses after its mkdir calls. Idempotent on reuse.
+    os.chmod(path, 0o755)
     return path
 
 
-def resolve_home_workspace(chat_id: int | None, config: Config) -> Path:
+def resolve_home_workspace(
+    chat_id: int | None,
+    config: Config,
+    data_dir: Path | None = None,
+) -> Path:
     """
     Return the user's home workspace path.
 
     Resolution order:
     1. `user.home_workspace` from users.yaml when set (admin override,
        returned as-is with no second-guessing).
-    2. `DATA_DIR/home/<chat_id>/`, auto-created via ensure_user_home().
+    2. `<data_dir>/home/<chat_id>/`, auto-created via ensure_user_home().
 
     All call sites (pool.py session init / get_workspace fallback,
     bot.py `/workspace home` handler and workspace listings) MUST go
@@ -521,6 +533,17 @@ def resolve_home_workspace(chat_id: int | None, config: Config) -> Path:
     Passing `config` (not just its admin home field) keeps the
     resolution decision local: the caller does not need to know whether
     to prefer users.yaml or the per-user default.
+
+    The `data_dir` parameter is exposed for tests that want to thread a
+    tmp_path in cleanly (beats monkeypatching the module attribute,
+    because a test that forgets the patch would silently write to the
+    real `/var/lib/kai`). When None, we resolve from the module-level
+    DATA_DIR at CALL time (not definition time): a `data_dir=DATA_DIR`
+    default would freeze the value at import, defeating the
+    `monkeypatch.setattr("kai.backend.DATA_DIR", ...)` pattern that
+    pool.py and bot.py tests still depend on. The underlying
+    ensure_user_home already takes data_dir as a parameter; this
+    mirrors that signature.
     """
     # Admin-authored override from users.yaml wins outright. No check
     # that the path exists - the config-load validator (config.py
@@ -532,9 +555,13 @@ def resolve_home_workspace(chat_id: int | None, config: Config) -> Path:
         return user.home_workspace
 
     # No users.yaml override: use the per-user Kai-managed directory
-    # under DATA_DIR. ensure_user_home is idempotent so calling it on
-    # every resolution is cheap.
-    return ensure_user_home(chat_id, DATA_DIR)
+    # under data_dir. Resolve DATA_DIR from the module here (not as a
+    # default arg value) so monkeypatching `kai.backend.DATA_DIR` in
+    # tests still flows through. ensure_user_home is idempotent so
+    # calling it on every resolution is cheap.
+    if data_dir is None:
+        data_dir = DATA_DIR
+    return ensure_user_home(chat_id, data_dir)
 
 
 def build_foreign_workspace_reminder(workspace: Path, home_workspace: Path) -> str | None:
--- a/src/kai/backend.py
+++ b/src/kai/backend.py
@@ -21,7 +21,7 @@ from collections.abc import AsyncIterator
 from dataclasses import dataclass, field
 from pathlib import Path
 
-from kai.config import PROJECT_ROOT, WorkspaceConfig
+from kai.config import DATA_DIR, PROJECT_ROOT, Config, WorkspaceConfig
 from kai.history import get_recent_history
 
 log = logging.getLogger(__name__)
@@ -435,6 +435,106 @@ def ensure_user_memory(chat_id: int | None, data_dir: Path) -> None:
             user_memory_file,
             exc_info=True,
         )
+
+
+def ensure_user_home(chat_id: int | None, data_dir: Path) -> Path:
+    """
+    Lazily create the per-user home workspace directory.
+
+    Returns the path that should be used as the user's home workspace
+    when users.yaml does not set an explicit home_workspace override.
+    Idempotent and cheap: a stat and a possible mkdir. Called on every
+    session init (and on `/workspace home`) so that any user without a
+    pre-created `home/<chat_id>/` - single-user dev, test runs, or a
+    user added to users.yaml after install - still lands in a valid
+    directory on their first message.
+
+    This is the companion to ensure_user_memory() above. Same scope
+    caveats apply (copied verbatim because the multi-user ownership
+    semantics are identical):
+
+    * Production multi-user (users.yaml entry has an explicit os_user
+      different from the service user): the install step (install.py
+      `_apply_migrate`) pre-creates and chowns `home/<chat_id>/` to the
+      user's os_user. mkdir here is then a no-op and the directory is
+      already writable by the subprocess identity.
+    * A user added to users.yaml AFTER install with a distinct os_user:
+      lazy bootstrap is NOT enough. mkdir here runs as the bot/service
+      identity, so the new directory is service-owned. The inner
+      subprocess (sudo -H -u <os_user>) can read but not write - the
+      same #347 regression that memory hit. A reinstall (or a manual
+      chown) is required for that case.
+    * Dev / single-user / users without a distinct os_user: lazy
+      bootstrap does the whole job. mkdir inherits the caller's
+      ownership, which IS the subprocess identity.
+
+    When chat_id is None (admin-less startup paths: tests, webhook
+    health checks, dev smoke tests) the function returns a fixed
+    `data_dir/home/anon` path. That path is only used when there is no
+    real user; it is not shared between actual Telegram users because
+    any real message carries a chat_id.
+
+    Mode 0755 with per-user chown at install time is what isolates
+    users from each other. Matches memory/ exactly (install.py
+    `_apply_migrate` uses the same ownership rules). Isolation comes
+    from ownership, not mode bits.
+
+    Unlike ensure_user_memory this does NOT seed a template file. The
+    home workspace is an empty directory that the user populates
+    themselves (cloning a repo, dropping notes, etc.).
+    """
+    # chat_id of None means we have no real Telegram session to key on
+    # (test runs, health checks, smoke runs without users.yaml). Use a
+    # fixed "anon" subdir so the resolver always returns a concrete
+    # path rather than None; avoids defensive None-checks at every call
+    # site.
+    if chat_id is None:
+        path = data_dir / "home" / "anon"
+    else:
+        path = data_dir / "home" / str(chat_id)
+
+    # mkdir parents=True is load-bearing: on a brand-new install the
+    # data_dir/home/ root may not exist yet (the installer creates it,
+    # but dev paths and the tests go straight through lazy bootstrap).
+    # exist_ok=True means this is safe to call on every session init.
+    path.mkdir(parents=True, exist_ok=True, mode=0o755)
+    return path
+
+
+def resolve_home_workspace(chat_id: int | None, config: Config) -> Path:
+    """
+    Return the user's home workspace path.
+
+    Resolution order:
+    1. `user.home_workspace` from users.yaml when set (admin override,
+       returned as-is with no second-guessing).
+    2. `DATA_DIR/home/<chat_id>/`, auto-created via ensure_user_home().
+
+    All call sites (pool.py session init / get_workspace fallback,
+    bot.py `/workspace home` handler and workspace listings) MUST go
+    through this function rather than reading any global home field on
+    Config directly - that field was removed by issue #353. The old
+    global fallback (PROJECT_ROOT / "home" / shared "/opt/kai/home/")
+    was a multi-user privacy hazard (every user landed in a directory
+    every other user could read).
+
+    Passing `config` (not just its admin home field) keeps the
+    resolution decision local: the caller does not need to know whether
+    to prefer users.yaml or the per-user default.
+    """
+    # Admin-authored override from users.yaml wins outright. No check
+    # that the path exists - the config-load validator (config.py
+    # load_config) already filtered out bad paths, and a path that
+    # became invalid post-load (unmounted drive, etc.) is the admin's
+    # problem to fix, not ours to paper over.
+    user = config.get_user_config(chat_id) if chat_id is not None else None
+    if user and user.home_workspace:
+        return user.home_workspace
+
+    # No users.yaml override: use the per-user Kai-managed directory
+    # under DATA_DIR. ensure_user_home is idempotent so calling it on
+    # every resolution is cheap.
+    return ensure_user_home(chat_id, DATA_DIR)
 
 
 def build_foreign_workspace_reminder(workspace: Path, home_workspace: Path) -> str | None:
--- a/src/kai/bot.py
+++ b/src/kai/bot.py
@@ -55,6 +55,7 @@ from telegram.ext import (
 )
 
 from kai import github_api, memory_command, services, sessions, webhook
+from kai.backend import resolve_home_workspace
 from kai.config import (
     DATA_DIR,
     MAX_CONTEXT_CEILING,
@@ -1259,7 +1260,10 @@ async def _do_switch_workspace(context: ContextTypes.DEFAULT_TYPE, chat_id: int,
     """
     pool = _get_pool(context)
     config: Config = context.bot_data["config"]
-    home = config.claude_workspace
+    # "home" for this user is per-user post-#353. Same resolver pool.py
+    # uses, so the equality check below matches the directory the user
+    # would land in on `/workspace home` or session restart.
+    home = resolve_home_workspace(chat_id, config)
 
     # Layer DB overrides (from /workspace config) on top of YAML baseline.
     yaml_config = config.get_workspace_config(path)
@@ -1288,7 +1292,9 @@ async def _switch_workspace(update: Update, context: ContextTypes.DEFAULT_TYPE, 
     assert update.message is not None
     pool = _get_pool(context)
     config: Config = context.bot_data["config"]
-    home = config.claude_workspace
+    # Per-user resolution: this is what `/workspace home` would route to
+    # for THIS user, not a shared default.
+    home = resolve_home_workspace(_chat_id(update), config)
 
     if path == pool.get_workspace(_chat_id(update)):
         await update.message.reply_text("Already in that workspace.")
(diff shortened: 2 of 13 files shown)


dcellison:
## Review by Kai

## PR Review: feat/per-user-home-workspace-353

### Warning

**`ensure_user_home` doesn't force mode 0o755 after `mkdir`** (`backend.py`, `ensure_user_home`)

`mkdir(mode=0o755)` is subject to umask masking. `install.py`'s `_apply_migrate` correctly follows `mkdir` with `os.chmod(user_dir, 0o755)` to force the intended bits, but `ensure_user_home` doesn't. With a service umask of 0o027, the lazily-created directory gets 0o750, leaving group members unable to traverse. The test `test_directory_is_0755` papers over this - it only asserts `mode & 0o022 == 0`, which passes on 0o750. Add `path.chmod(0o755)` after the `mkdir` call to match the installer.

**Comment/code mismatch in `_apply_migrate` for DATA_DIR-internal overrides** (`install.py`, new home-provisioning block)

The leading comment says:
```
Override path under DATA_DIR (rare, but valid): provision the override path itself
```

But the code always provisions `home_root / str(chat_id)`, never `override`. For a user with `home_workspace: /var/lib/kai/custom_dir`:
- The installer creates `/var/lib/kai/home/<chat_id>/` (never used at runtime)
- `/var/lib/kai/custom_dir` is never provisioned by installer or runtime (`resolve_home_workspace` returns the override directly, skipping `ensure_user_home`)
- Result: the user's actual runtime workspace doesn't exist; first write will fail

Fix: when `override` is set and inside `data_path`, use `override` as `user_dir` instead of `home_root / str(chat_id)`.

**`_apply_migrate` home block silently no-ops if `home_root` doesn't exist yet** (`install.py`)

```python
if home_root.is_dir():
    for chat_id, _os_user in memory_owners:
```

If `_apply_directories` hasn't run (or is reordered), per-user dirs are silently skipped with no warning. `_apply_migrate` for memory doesn't have this guard - it creates unconditionally. Either add a `home_root.mkdir(parents=True, exist_ok=True)` fallback, or document the ordering requirement explicitly.

---

### Suggestion

**`resolve_home_workspace` captures `DATA_DIR` at import time** (`backend.py:~495`)

Using the module-level `DATA_DIR` rather than threading it through `config` (e.g., `config.data_dir`) means tests must monkeypatch the module attribute. This works but is fragile - a test that forgets the patch will silently write to the real `/var/lib/kai`. The existing `ensure_user_home(chat_id, data_dir)` signature already takes `data_dir` as a parameter; extending the same pattern to `resolve_home_workspace(chat_id, config, data_dir=DATA_DIR)` would make the coupling explicit.

**`test_directory_is_0755` test name misrepresents its assertion** (`test_backend.py:~379`)

The test accepts any mode where `mode & 0o022 == 0`, including 0o750 or 0o700. The name implies 0o755 is guaranteed. This directly masks the umask issue above.

---

### Clean

- The `anon` sentinel path for `chat_id=None` is documented and intentionally shared (health checks only). Acceptable.
- `_collect_user_home_overrides` validation is appropriately defensive and mirrors `_collect_user_memory_owners` conventions.
- `_app["workspace"] = None` with 403 fallback is a safe change - slightly more restrictive than the old behavior, not less.
- `_make_config` back-compat alias in `test_bot.py` and `test_pool.py` is well-structured; the alias wins only when `user_configs` isn't explicitly passed.
- No path-traversal risk: `str(chat_id)` is always an int or the literal `"anon"`.

The DATA_DIR-internal override provisioning mismatch is the most actionable bug; the mode issue is the most likely to cause a subtle production failure.

dcellison:
## Review by Kai

## PR Review: feat/per-user-home-workspace-353

All three warnings and both suggestions from the W1 review have been addressed correctly. The fixes are clean and the tests now properly assert exact 0o755 mode.

---

### Warning

**`home_root` defensive fallback in `_apply_migrate` is missing `os.chmod`** (`install.py`, new home-provisioning block)

```python
if not dry_run:
    home_root.mkdir(parents=True, exist_ok=True, mode=0o755)
# ← no os.chmod(home_root, 0o755) here
```

Every other `mkdir` in this block is followed by `os.chmod` to defeat umask masking - `user_dir` gets it on the next line. `home_root` doesn't. In the edge case this fallback exists to handle (i.e., `_apply_directories` hasn't run), a service umask of 0o027 leaves `home_root` at 0o750. A distinct `os_user` subprocess that isn't in the service group can't traverse into `home_root` to reach its own `home/<chat_id>/` directory, even though that subdirectory has correct 0o755 mode. Fix: add `os.chmod(home_root, 0o755)` after the `mkdir`.

---

### Suggestion

**`user_dir.exists()` idempotency guard skips mode re-application on reinstall** (`install.py`)

```python
if user_dir.exists():
    continue  # idempotent across reinstalls
```

A directory created by `ensure_user_home` at runtime under a hostile umask (0o750) won't be corrected when the operator later runs `sudo make install`. The memory provisioning block has the same behavior, so this is consistent, and the risk is low - correct mode is enforced on the write path in `ensure_user_home`. Noted for awareness rather than as a blocker.

---

### Clean

- W1 fixes are complete: `ensure_user_home` adds `os.chmod(path, 0o755)` after `mkdir`; `_apply_migrate` now provisions the override path rather than always `home_root / str(chat_id)` for DATA_DIR-internal overrides; `home_root.mkdir` fallback eliminates the silent no-op on missing parent.
- `resolve_home_workspace` resolves `DATA_DIR` at call time, not import time - monkeypatching works correctly.
- `test_directory_mode_is_exactly_0755` now asserts `mode == 0o755` rather than the weaker `mode & 0o022 == 0`.
- `_collect_user_home_overrides` mirrors `_collect_user_memory_owners` conventions - appropriate and defensive.
- `webhook.py` `_app["workspace"] = None` with 403 fallback is safe; the pool is always present in production when send-file fires.
- New regression tests (`test_wizard_defaults_do_not_reintroduce_shared_home`, `TestApplyMigratePerUserHome`) cover the three provisioning cases exhaustively.

The missing `os.chmod` on `home_root` is the only new finding. Everything else is solid.

dcellison:
## Review by Kai

## PR Review: feat/per-user-home-workspace-353 (W3)

All three W1 warnings and the W2 `home_root` chmod finding are correctly addressed. The code is solid. One new finding:

---

### Suggestion

**`is_relative_to(data_path)` uses an unresolved `data_path` while `override` is always `.resolve()`'d** (`install.py`, home block in `_apply_migrate`)

```python
if override is not None and not override.is_relative_to(data_path):
    continue
```

`override` comes from `_collect_user_home_overrides` as `Path(home_ws).expanduser().resolve()` - canonicalized. `data_path` is whatever the caller passes in, which may not be resolved. On macOS, `/var/lib/kai` is a symlink to `/private/var/lib/kai`; if the operator writes `home_workspace: /private/var/lib/kai/custom` in `users.yaml`, `override.resolve()` is `/private/var/lib/kai/custom` but `data_path` is `/var/lib/kai`, so `is_relative_to` returns False and the block falls through to Case 3 (skip), leaving the runtime directory unprovisioned.

Impact is install-time only - `resolve_home_workspace` still returns the override at runtime, and `ensure_user_home` is not called for overrides - so a single-os_user setup would work (service can create the dir on first write). A distinct-os_user setup would fail on first write because the pre-provisioned chown step was skipped.

Fix: `override.is_relative_to(data_path.resolve())`.

---

### Clean

Everything from W1 and W2 is correctly resolved:
- `ensure_user_home`: `os.chmod(path, 0o755)` after `mkdir` forces mode past umask.
- `_apply_migrate` Case 2: provisions `override` path rather than `home_root / str(chat_id)` for DATA_DIR-internal overrides.
- `_apply_migrate` fallback: `home_root.mkdir(parents=True, exist_ok=True)` + `os.chmod(home_root, 0o755)` eliminates the silent no-op.
- `test_directory_mode_is_exactly_0755`: asserts `mode == 0o755`, not the weaker `mode & 0o022 == 0`.
- `resolve_home_workspace`: resolves `DATA_DIR` at call time, not import time.
- `_collect_user_home_overrides`: defensive and mirrors `_collect_user_memory_owners` conventions.
- `TestApplyMigratePerUserHome` covers all three provisioning cases with hostile-umask verification.
- `test_wizard_defaults_do_not_reintroduce_shared_home` pins both wizard paths.

The symlink issue is the only new finding, and it only affects installer pre-provisioning on hosts with symlinked `DATA_DIR`.

dcellison:
## Review by Kai

## PR Review: feat/per-user-home-workspace-353 (W4)

All three W1 warnings, the W2 `home_root` chmod finding, and the W3 `data_path.resolve()` symlink fix are correctly addressed. The code is solid.

---

### Suggestion

**Double call to `resolve_home_workspace` in `pool.py _create_instance`** (`pool.py`, `_create_instance`)

`resolve_home_workspace(chat_id, self._config)` is called twice in the same method - once to set `workspace` (the initial landing directory), and again to set `home_ws` (the "home" reference for the foreign-workspace reminder). Each call invokes `ensure_user_home`, which does a `stat` + possible `mkdir` + `chmod` per call.

The two values have different semantics (`workspace` can be overridden by DB/workspace_config; `home_ws` is always the canonical home), so the separate calls are intentionally correct. The redundancy is structural, not a bug. Worth a comment noting why they are separate rather than `home_ws = workspace`.

---

### Clean

- All prior findings resolved: `ensure_user_home` has `os.chmod` post-`mkdir`; Case 2 provisions the override path rather than `home_root / str(chat_id)`; `home_root.mkdir` fallback with `os.chmod` eliminates the silent no-op; `resolve_home_workspace` resolves `DATA_DIR` at call time; `test_directory_mode_is_exactly_0755` asserts `mode == 0o755`; `data_path.resolve()` fixes the symlink containment check.
- `_apply_migrate` dry-run behavior with a missing `home_root` is correct - `user_dir.exists()` returns False, the loop prints what would be created, and no filesystem mutations happen.
- `webhook.py _app["workspace"] = None` with `if workspace:` guard is safe; the 403 fallback is more restrictive than the old behavior, never less.
- `test_triage.py` sensitive-path substitution is reasonable - `/etc/kai/env` is a more accurate sensitive-path example post-#353 than the removed shared home.
- `_make_config` back-compat aliases in `test_bot.py` and `test_pool.py` are correctly scoped to the specific chat IDs each test suite uses.
- `TestApplyMigratePerUserHome` covers all five cases (default, DATA_DIR-internal override, external override, missing `home_root`, symlinked `DATA_DIR`) with hostile-umask verification where it matters.

No new findings beyond the minor suggestion above. Approved.

dcellison:
## Review by Kai

## PR Review: feat/per-user-home-workspace-353 (W5)

The W4 suggestion - documenting why `resolve_home_workspace` is called twice in `pool.py _create_instance` - has been addressed with a thorough block comment. All prior findings (W1-W4) are correctly resolved.

---

### Clean

- **W4 comment addressed**: The block comment on `home_ws = resolve_home_workspace(chat_id, self._config)` at `pool.py:~183` correctly explains the intentional separation: `workspace` is the initial landing dir (overridable by DB switch / `workspace_config`), `home_ws` is the canonical home used for foreign-workspace reminder detection. The redundant stat is acknowledged and justified.
- **`ensure_user_home`**: `os.chmod(path, 0o755)` after `mkdir` - umask defeated.
- **`_apply_migrate` Case 2**: provisions `override` path rather than `home_root / str(chat_id)` for DATA_DIR-internal overrides.
- **`_apply_migrate` fallback**: `home_root.mkdir(parents=True, exist_ok=True)` + `os.chmod(home_root, 0o755)` - no silent no-op, mode forced past umask.
- **`data_path_resolved = data_path.resolve()`**: symlink-safe containment check for the macOS `/var/lib` → `/private/var/lib` case.
- **`_apply_directories`**: `data_path / "home"` added to the pre-creation list, so the defensive fallback in `_apply_migrate` is the exception rather than the rule.
- **`resolve_home_workspace`**: resolves `DATA_DIR` at call time via `if data_dir is None: data_dir = DATA_DIR`; monkeypatching `kai.backend.DATA_DIR` continues to work in all three call sites (`pool.py`, `bot.py`, `get_workspace`).
- **`_app["workspace"] = None`**: safe initialization; 403 fallback is more restrictive than the old behavior.
- **`_make_config` back-compat aliases** in `test_bot.py` and `test_pool.py`: correctly scoped to the specific chat IDs each suite uses; `user_configs` explicit override wins.
- **`TestApplyMigratePerUserHome`**: covers all five provisioning cases including hostile-umask and symlinked DATA_DIR.
- **`test_directory_mode_is_exactly_0755`**: asserts `mode == 0o755`, not the weaker `mode & 0o022 == 0`.
- **`test_wizard_defaults_do_not_reintroduce_shared_home`**: covers both basic and advanced wizard paths; `CLAUDE_WORKSPACE` env absence is pinned.

No new findings. Approved.

dcellison:
## Review by Kai

## PR Review: feat/per-user-home-workspace-353 (W6)

All W1-W5 findings are correctly resolved. After a fresh pass over the diff, the code is solid.

---

### Clean

- **`ensure_user_home`**: `os.chmod(path, 0o755)` after `mkdir` forces mode past umask. `parents=True` handles cold dev paths. `anon` sentinel for `chat_id=None` is correct and documented.
- **`resolve_home_workspace`**: `DATA_DIR` resolved at call time (not import time); `user.home_workspace` wins outright; `data_dir` parameter threads cleanly for tests.
- **`_apply_migrate` home block**: all five cases covered (`TestApplyMigratePerUserHome`), symlink-safe containment via `data_path.resolve()`, `home_root` defensive `mkdir` + explicit `chmod`, correct case 2 provisions override path not `home_root/<chat_id>`.
- **`_handle_send_file`**: pool-gated confinement (`pool is None` → 403) is strictly more restrictive than the prior global fallback. `pool.get_workspace(chat_id)` gives per-user confinement at request time.
- **`_resolve_local_repo`**: home workspace check removed correctly - with per-user homes the webhook has no user context to key on; history fallback (`workspace_history`) covers the realistic case where a user has previously switched to the repo.
- **`update_workspace()` removal**: call sites cleaned from `bot.py` (7 patches removed) and `test_webhook_api.py`; `TestUpdateWorkspace` deleted; no dangling callers.
- **Back-compat aliases** in `_make_config` (`test_bot.py`, `test_pool.py`): `claude_workspace=` → `UserConfig.home_workspace` for the fixture's default chat_id. Correctly scoped; explicit `user_configs` wins.
- **`test_workspace_home_lands_in_per_user_directory`**: pins the spec-353 privacy guarantee - `monkeypatch.setattr("kai.backend.DATA_DIR", tmp_path)` + asserts `change_workspace` receives `tmp_path/home/12345`.
- **`test_wizard_defaults_do_not_reintroduce_shared_home`**: covers both basic and advanced wizard paths; `CLAUDE_WORKSPACE` env absence pinned.

---

No new findings. Approved.

PR #356, opened 2026-04-21
fix(memory): /memory stats crashes on int prompt_version

### LINKED ISSUE #338: `/memory stats` crashes with TypeError when extracted facts exist
## Symptom

Tapping the **Stats** button on the `/memory` dashboard does nothing visible to the user. Running `/memory stats` directly produces the same silent failure. The service log shows the same traceback on both paths:

```
TypeError: object of type 'int' has no len()
  File "src/kai/memory_command.py", line 687, in <genexpr>
    version_width = max(len(v) for v, _ in sorted_versions)
```

The error repeats on every tap, so neither the callback path (`handle_memory_callback` -> `_send_stats`) nor the slash-command path (`handle_memory_command` -> `_send_stats`) is usable.

## Root cause

Type mismatch between the extraction write and the stats read. The write side stores `prompt_version` as an `int`; the read side assumes a `str`.

- `src/kai/memory_extraction.py:41` declares `_EXTRACTION_PROMPT_VERSION = 1` (integer).
- `src/kai/memory_extraction.py:618` writes it unchanged into Mem0 metadata:
  ```python
  "prompt_version": _EXTRACTION_PROMPT_VERSION,
  ```
- `src/kai/memory.py:216` types the aggregate as `by_prompt_version: dict[str, int]`.
- `src/kai/memory.py:1158-1163` populates it with the raw metadata value, and the comment claims the value is "stored as a string". The comment is wrong; the value arrives as `int`.
- `src/kai/memory_command.py:687` then runs `max(len(v) for v, _ in sorted_versions)`, expecting string keys. `len(1)` raises `TypeError`.

The `.get("prompt_version", "")` default at `memory.py:1162` only guards against a missing key, not a mistyped value, so the int flows straight through.

## Impact

Any fact already written to Qdrant has `int` in its `prompt_version` metadata. Fixing only the write side will not rescue the existing rows: `_build_stats` will keep crashing until the read side is defensive, or the old rows are migrated / wiped.

## Suggested fix

Fix both sides so write and read agree with the type annotation and the existing comment.

1. **Read side (handles existing data).** In `src/kai/memory_command.py`:
   - Line 687: `version_width = max(len(str(v)) for v, _ in sorted_versions)`
   - Line 689: `f"  {str(version).ljust(version_width)}  {count:>3}"`
   This makes the stats view resilient to whatever `prompt_version` type is on disk.

2. **Write side (prevents recurrence).** In `src/kai/memory_extraction.py:41`:
   - Change to `_EXTRACTION_PROMPT_VERSION = "1"`.
   This matches the `dict[str, int]` annotation and the `memory.py:1158` comment.

Doing only (2) leaves existing rows broken. Doing only (1) leaves the documented contract (`dict[str, int]`) a lie.

## Test gap worth noting

PR #337 went through eight review rounds focused on spec conformance. No round exercised the Stats button against a store that had extracted facts in it: empty-store hits the early-return at `_build_stats` line 626 and never reaches line 687. End-to-end smoke against a populated store would have caught this before merge. Worth adding a unit test for `_build_stats` with a non-empty `by_prompt_version` and a fixture that simulates the real metadata shape.

## Not in scope for this issue

The **Search** button on the dashboard is not broken but is misleading. It fires `query.answer(_HELP_TEXT, show_alert=True)`, a transient toast, because Telegram inline keyboards cannot accept free-text input. Users reasonably expect it to open a search. Either relabel to "Help" or drop the button. This is a UX issue, not part of this stats bug; file separately if it warrants tracking.

dcellison:
## Summary

Fixes #338. Tapping the **Stats** button on the `/memory` dashboard (or running `/memory stats`) raised `TypeError: object of type 'int' has no len()` inside `_build_stats` and silently swallowed the reply. The stats view was unreachable on any non-empty store.

Root cause was a type mismatch between the write and read sides of `prompt_version`:

- Write side: `_EXTRACTION_PROMPT_VERSION = 1` (int) was written straight into Mem0 metadata.
- Aggregation side: `MemoryStats.by_prompt_version` is annotated `dict[str, int]` with an inline comment claiming "stored as a string", but the read site cast nothing.
- Render side: `_build_stats` called `len(v)` and `v.ljust(...)` on the dict keys. `len(1)` raises.

## Changes

Three-point fix per the spec breakdown:

1. **Write side** (`src/kai/memory_extraction.py`). `_EXTRACTION_PROMPT_VERSION` is now typed `str = "1"`. Chose `"1"` over `"v1"` so legacy int rows (which stringify to `"1"` via the aggregation cast) bucket together with new rows; `"v1"` would have produced two visible buckets in the stats view for the same prompt revision until legacy rows aged out.
2. **Aggregation side** (`src/kai/memory.py`). `pv = m.metadata.get("prompt_version", "")` becomes `pv = str(...)`. Legacy int-typed rows in Qdrant now bucket under the stringified key. The dict annotation is no longer a lie.
3. **Render side** (`src/kai/memory_command.py`). Belt-and-suspenders `str()` wraps around `len()` and `ljust()` targets in `_build_stats`. The aggregation cast already guarantees strings, but a caller that constructs `MemoryStats` directly (tests do this; a future admin endpoint might) bypasses that guarantee.

The existing `memory.py` comment is rephrased from "stored as a string" to "bucketed as a string (the cast at the aggregation boundary handles legacy int rows)" so it remains accurate for both legacy and new rows.

No data migration. Legacy int-typed rows in Qdrant stay as-is; the read-side cast handles them in perpetuity.

## Tests

Two new regression tests, both verified to fail on `main` before the fix and pass after:

- `tests/test_memory_command.py::TestBuildStats::test_renders_int_keyed_prompt_version_bucket` - hands `_build_stats` a `MemoryStats` with `by_prompt_version={1: 1}` (int key, the load-bearing fixture - `len("1")` works on `main`, `len(1)` raises).
- `tests/test_memory.py::TestGetStatsExtended::test_by_prompt_version_normalizes_int_to_str` - drives `get_stats` over a mixed legacy-int and post-fix-str row and asserts both bucket to `{"1": 2}`. Pre-fix the dict has mixed keys `{1: 1, "1": 1}`.

Existing `_build_stats` and `test_by_prompt_version_counts` tests use `"v3"`/`"v2"` str fixtures and continue to pass; they document the str-input happy path that was already covered.

## Test Plan

- [x] Both new regression tests fail on `main` (verified before applying fixes).
- [x] Both new regression tests pass after applying fixes.
- [x] `make check` passes (ruff lint + format).
- [x] `make test` passes 2072 tests; the 2 failures + 5 errors in `TestInitMemory` and `TestMemoryIntegration` are pre-existing on `main` (verified via `git stash`), unrelated to this change.
- [ ] Manual smoke: open `/memory` on a store with at least one extracted fact, tap **Stats**, confirm the "Prompt versions:" section renders without raising.

### DIFF
--- a/src/kai/memory.py
+++ b/src/kai/memory.py
@@ -1162,13 +1162,17 @@ def get_stats(*, user_id: str) -> MemoryStats:
         # annotation would be a lie and downstream renderers would
         # crash on len(int).
         #
-        # The `... or ""` guard collapses both absent keys and
-        # null-valued keys into the same empty-string sentinel
-        # bucket. Without it, str(None) would create a literal
-        # "None" bucket that looks like a real version label.
-        # Empty-string bucket is itself meaningful: it surfaces a
-        # regression in extraction (forgetting to stamp the version)
-        # rather than silently merging into other counts.
+        # The `... or ""` guard collapses any falsy stored value
+        # (absent key, explicit None, 0, etc.) into the same
+        # empty-string sentinel bucket. Without it, str(None) would
+        # create a literal "None" bucket that looks like a real
+        # version label. Falsy version numbers like 0 are not used
+        # in practice, but folding them in keeps the sentinel
+        # behavior uniform across all "no usable version" shapes.
+        # The empty-string bucket itself is meaningful - it
+        # surfaces a regression in extraction (forgetting to stamp
+        # the version) rather than silently merging into other
+        # counts.
         pv = str(m.metadata.get("prompt_version") or "")
         by_prompt_version[pv] = by_prompt_version.get(pv, 0) + 1
 
--- a/src/kai/memory.py
+++ b/src/kai/memory.py
@@ -1155,11 +1155,19 @@ def get_stats(*, user_id: str) -> MemoryStats:
         for t in m.metadata.get("tags") or []:
             by_tag[t] = by_tag.get(t, 0) + 1
 
-        # prompt_version: stored as a string; "" treated as a distinct
-        # bucket so a regression in extraction (forgetting to stamp the
-        # version) shows up rather than silently merging into other
-        # counts. Bucket key is the value itself.
-        pv = m.metadata.get("prompt_version", "")
+        # prompt_version: bucketed as a string (the cast at the
+        # aggregation boundary handles legacy int rows). "" treated
+        # as a distinct bucket so a regression in extraction
+        # (forgetting to stamp the version) shows up rather than
+        # silently merging into other counts. Bucket key is the value
+        # itself, stringified.
+        #
+        # Cast defensively: legacy rows written before the write-side
+        # fix in spec 338 carry an int here, and the dict[str, int]
+        # annotation would be a lie without this. Stringifying at the
+        # aggregation boundary means any downstream consumer sees the
+        # documented shape.
+        pv = str(m.metadata.get("prompt_version", ""))
         by_prompt_version[pv] = by_prompt_version.get(pv, 0) + 1
 
         # confidence: the schema enforces [0.5, 1.0], but a bad row
--- a/src/kai/memory.py
+++ b/src/kai/memory.py
@@ -1155,19 +1155,21 @@ def get_stats(*, user_id: str) -> MemoryStats:
         for t in m.metadata.get("tags") or []:
             by_tag[t] = by_tag.get(t, 0) + 1
 
-        # prompt_version: bucketed as a string (the cast at the
-        # aggregation boundary handles legacy int rows). "" treated
-        # as a distinct bucket so a regression in extraction
-        # (forgetting to stamp the version) shows up rather than
-        # silently merging into other counts. Bucket key is the value
-        # itself, stringified.
+        # prompt_version: bucketed as a string. The cast handles
+        # legacy rows whose metadata stores prompt_version as a
+        # non-string value (older revisions of the extraction code
+        # wrote an int); without the cast the dict[str, int]
+        # annotation would be a lie and downstream renderers would
+        # crash on len(int).
         #
-        # Cast defensively: legacy rows written before the write-side
-        # fix in spec 338 carry an int here, and the dict[str, int]
-        # annotation would be a lie without this. Stringifying at the
-        # aggregation boundary means any downstream consumer sees the
-        # documented shape.
-        pv = str(m.metadata.get("prompt_version", ""))
+        # The `... or ""` guard collapses both absent keys and
+        # null-valued keys into the same empty-string sentinel
+        # bucket. Without it, str(None) would create a literal
+        # "None" bucket that looks like a real version label.
+        # Empty-string bucket is itself meaningful: it surfaces a
+        # regression in extraction (forgetting to stamp the version)
+        # rather than silently merging into other counts.
+        pv = str(m.metadata.get("prompt_version") or "")
         by_prompt_version[pv] = by_prompt_version.get(pv, 0) + 1
 
         # confidence: the schema enforces [0.5, 1.0], but a bad row
--- a/src/kai/memory_command.py
+++ b/src/kai/memory_command.py
@@ -655,9 +655,15 @@ def _build_stats(stats: MemoryStats) -> tuple[str, InlineKeyboardMarkup]:
             stats.by_prompt_version.items(),
             key=lambda item: (-item[1], item[0]),
         )
-        version_width = max(len(v) for v, _ in sorted_versions)
+        # Belt-and-suspenders: aggregation in memory.py already casts
+        # prompt_version to str, but a caller that constructs
+        # MemoryStats by a different path (tests do this directly,
+        # and a future admin endpoint might) would not be caught by
+        # that. Keep the renderer resilient on its own so a stray int
+        # key cannot reproduce the spec 338 TypeError.
+        version_width = max(len(str(v)) for v, _ in sorted_versions)
         for version, count in sorted_versions:
-            lines.append(f"  {version.ljust(version_width)}  {count:>3}")
+            lines.append(f"  {str(version).ljust(version_width)}  {count:>3}")
 
     text = "\n".join(lines)
     kb = InlineKeyboardMarkup([[InlineKeyboardButton("back", callback_data=_encode_callback("dash"))]])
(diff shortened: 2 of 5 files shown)


dcellison:
## Review by Kai

The fix is correct and the root-cause analysis is accurate. A few findings:

---

**Warning - Convention violation: issue/spec references in source code comments**

`memory.py:1163`, `memory_command.py:658-664`, and both test files (`test_memory.py:1907`, `test_memory_command.py:628`) all contain comments like `# Spec 338 regression:`, `# spec 338 TypeError`, and `# the spec 338 TypeError`. Per CLAUDE.md conventions, code comments must not reference issue/ticket numbers - those rot as the codebase evolves and belong in the PR description, not the source. The WHY should be expressed in terms of invariants, not external identifiers.

---

**Suggestion - Sort comparator not protected against mixed-type keys**

In `memory_command.py`, the fix wraps `len()` and `ljust()` with `str()`, but the sort key still compares the raw dict key:

```python
sorted_versions = sorted(
    stats.by_prompt_version.items(),
    key=lambda item: (-item[1], item[0]),  # item[0] still raw
)
```

If `by_prompt_version` has mixed `int`/`str` keys (e.g., `{1: 1, "2": 3}`), Python 3 raises `TypeError: '<' not supported between instances of 'str' and 'int'` on the comparison - before `str()` ever gets called. The PR's stated rationale for the renderer fix is that callers constructing `MemoryStats` directly bypass the aggregation cast; that scenario can produce mixed-type dicts with more than one entry. The sort key should be `key=lambda item: (-item[1], str(item[0]))` to match the rendering fix's intent.

The test uses a single-key dict `{1: 1}`, which never triggers a comparison, so it passes without catching this.

---

**Suggestion - `str(None)` creates a spurious bucket**

`memory.py:1174`:
```python
pv = str(m.metadata.get("prompt_version", ""))
```

If a metadata dict has `"prompt_version": None` (key present, value null), `.get()` returns `None` and `str(None)` produces the string `"None"` - a bucket that looks like a real version label. The default `""` only fires when the key is absent. Guarding with `str(m.metadata.get("prompt_version") or "")` would collapse both absent and null to the empty-string sentinel bucket. Low-priority since `None` values are unlikely in practice, but it's the one remaining shape the cast doesn't cleanly handle.

---

The core three-point fix (write side, aggregation cast, render-side belt-and-suspenders) is sound. The `"1"` vs `"v1"` bucketing rationale is correct. Tests are well-targeted. The sort comparator gap is the only functional hole.

dcellison:
All three findings addressed in `24cebb9`.

**Convention violation - spec number references in comments**

Removed from all four files. Comments now express the WHY in invariant terms (what shape the data can take, why the cast belongs where it does), no `Spec 338` / `# spec 338` references anywhere in source or tests.

**Sort comparator gap**

Right call - the `len()`/`ljust()` wraps don't help if the sort raises first. Tiebreaker is now `str(item[0])` in `memory_command.py`. New regression test `TestBuildStats::test_renders_mixed_int_str_keyed_prompt_versions` pins it - uses `{1: 1, "2": 1}` (equal counts force the tiebreaker to fire on int<->str). Verified to raise `TypeError: '<' not supported between instances of 'str' and 'int'` against the round-1 commit before this push.

**`str(None)` phantom bucket**

Fixed in `memory.py` - `str(m.metadata.get("prompt_version") or "")` collapses both absent and null cases to the empty-string sentinel. New regression test `TestGetStatsExtended::test_by_prompt_version_collapses_null_with_missing` mixes a `prompt_version=None` row with a key-absent row and asserts `{"": 2}` plus `"None" not in stats.by_prompt_version`. Also verified to fail on the round-1 commit (it produced `{"None": 1, "": 1}`).

Test count now 2074 (up from 2072). The 2 failures + 5 errors in `TestInitMemory`/`TestMemoryIntegration` are the Mem0 telemetry-Qdrant lock collision tracked separately as #357; unaffected by this push.

dcellison:
## Review by Kai

All three findings from the prior review are addressed:

- **Sort comparator** - fixed with `str(item[0])` at `memory_command.py:658`
- **`str(None)` spurious bucket** - fixed with `... or ""` at `memory.py:1174`
- **Spec/issue refs in comments** - removed; all comments now describe invariants only

A new test (`test_renders_mixed_int_str_keyed_prompt_versions`) was added specifically to cover the sort comparator gap, which the prior review noted was invisible to single-key fixtures.

---

**Suggestion (low priority) - falsiness of `0` in `... or ""`**

`memory.py:1174`:
```python
pv = str(m.metadata.get("prompt_version") or "")
```

If `prompt_version` is integer `0`, `0 or ""` evaluates to `""`, so it buckets with "no version" rows rather than under `"0"`. This mirrors the `None` behavior but without explicit documentation. In practice prompt versions won't be zero, but the comment documents `None` and absent keys as the motivation for the guard without mentioning this edge case. Adding a brief note like "falsy values (None, 0) collapse to the empty-string sentinel" would make the intent complete. Not a functional concern given the domain.

---

Otherwise the PR is clean. Logic is sound, convention compliance is good, and the three-layer fix (write side, aggregation cast, render guard) is correctly scoped. No new issues.

dcellison:
Comment expanded in `e7bee83`. The `... or ""` guard now reads as a general "any falsy stored value collapses to the empty-string sentinel" rather than a `None`-specific note, with `0` called out explicitly so a future reader doesn't have to rediscover the truthiness behavior. No code change.

dcellison:
## Review by Kai

The PR is clean. All findings from both prior review rounds are addressed in the current diff.

The round-2 low-priority suggestion about documenting the falsiness of `0` is resolved - the comment at `memory.py:1163` now explicitly says "any falsy stored value (absent key, explicit None, 0, etc.)" which covers the undocumented edge case the prior reviewer flagged.

Spot-checking the remaining logic:

- **Write side** (`memory_extraction.py:41`): `str = "1"` is correctly typed and the `"1"` vs `"v1"` bucketing rationale holds.
- **Aggregation cast** (`memory.py`): `str(m.metadata.get("prompt_version") or "")` correctly handles absent, `None`, `0`, and int-typed legacy rows. The `or ""` guard precedes `str()` so `str(None)` is never reached.
- **Sort comparator** (`memory_command.py:658`): `str(item[0])` wraps the tiebreaker before `sorted()` runs - the only place it can prevent the `TypeError`. Correct placement.
- **Render guards** (`memory_command.py:664,666`): `str(v)` on `len()` and `ljust()` targets is belt-and-suspenders; harmless and correctly motivated by the direct-construction bypass path.
- **Tests**: Four regression tests, all using invariant-term comments with no issue/spec references. The mixed-type sort test (`{1: 1, "2": 1}` with equal counts) correctly forces the tiebreaker to fire, which a single-key fixture cannot.

No new issues. Approve.

PR #358, opened 2026-04-21
fix(tests): isolate Mem0 telemetry Qdrant via MEM0_DIR

### LINKED ISSUE #357: tests: isolate Mem0 telemetry Qdrant via MEM0_DIR env var
## Symptom

`tests/test_memory.py::TestInitMemory` and `tests/test_memory.py::TestMemoryIntegration` reliably fail on a dev machine where the production Kai service is running, with:

```
RuntimeError: Storage folder $HOME/.mem0/migrations_qdrant is already
accessed by another instance of Qdrant client. If you require concurrent
access, use Qdrant server instead.
```

Affected tests (2 failures + 5 errors locally; 0 on CI):

- `TestInitMemory::test_init_creates_qdrant_dir`
- `TestInitMemory::test_init_dimension_mismatch_disables`
- `TestMemoryIntegration::test_add_and_search`
- `TestMemoryIntegration::test_search_relevance_ranking`
- `TestMemoryIntegration::test_multi_user_isolation`
- `TestMemoryIntegration::test_get_all_and_delete_all`
- `TestMemoryIntegration::test_format_context_integration`

## Why CI is green and dev is red

CI (`.github/workflows/ci.yml`) runs `make test` on a fresh Ubuntu container with no Kai service running, so nothing holds the lock and the tests pass. On a dev machine where the production service is up (LaunchDaemon `com.syrinx.kai`), the lock conflict fires every time. This produces the worst possible failure mode: green on the gate that protects `main`, red on every developer who ever runs the suite.

## Root cause

The fixture `real_memory_instance` in `tests/test_memory.py:91-127` already does the right thing for the user-facing vector store - it points `vector_store.config.path` at `tmp_path_factory.mktemp("qdrant")`. But Mem0's `Memory.__init__` opens a *second* Qdrant local-mode client for telemetry/migration tracking, and that one is hardcoded outside the user-configured path:

```python
# .venv/lib/python3.13/site-packages/mem0/memory/main.py:378-379
provider_path = f"migrations_{self.config.vector_store.provider}"
telemetry_config_dict['path'] = os.path.join(mem0_dir, provider_path)
```

`mem0_dir` comes from `mem0/memory/setup.py:8`:

```python
mem0_dir = os.environ.get("MEM0_DIR") or os.path.join(home_dir, ".mem0")
```

So Mem0 always opens `$HOME/.mem0/migrations_*` for its own bookkeeping, regardless of where the user-configured collection lives. Both production Kai and the test process try to open the same `$HOME/.mem0/migrations_qdrant` folder; Qdrant local mode permits only one writer at a time via portalocker.

## Fix

Set `MEM0_DIR` to an isolated tmp path before any Mem0 import in the integration fixture, and (separately) inside the per-test fixtures `TestInitMemory` uses. One option:

```python
@pytest.fixture(scope="module")
def real_memory_instance(tmp_path_factory):
    if not _HAS_MEM0:
        pytest.skip("mem0ai not installed")

    # Mem0 opens a second Qdrant client at $MEM0_DIR/migrations_qdrant
    # for its own telemetry/migration tracking. Without isolation, that
    # collides with the production service's lock on the same folder.
    mem0_root = tmp_path_factory.mktemp("mem0_root")
    os.environ["MEM0_DIR"] = str(mem0_root)

    from mem0 import Memory
    ...
```

For `TestInitMemory::test_init_*` the same `MEM0_DIR` patch needs to wrap the `init_memory(config)` call. Either make it a session-level autouse fixture for the integration class or apply it per-test.

## Acceptance

- [ ] All 7 affected tests pass on a dev machine while production Kai is running.
- [ ] All 7 still pass on CI.
- [ ] No test writes to `$HOME/.mem0/` during the suite (verify with `lsof` or by stashing the directory before the run).

## Out of scope

The spec 338 PR (#356) that surfaced this issue. That PR's `make test` output included these failures and they were verified pre-existing on `main`; this issue tracks the underlying test-isolation defect separately.

dcellison:
## Summary

Mem0 opens a hardcoded telemetry/migration Qdrant client at `$MEM0_DIR/migrations_qdrant`, outside the user's `vector_store.config`. Default is `$HOME/.mem0/migrations_qdrant`, which collides with the production Kai service's portalocker lock on dev machines and kills 7 memory tests with `Storage folder ... is already accessed by another instance of Qdrant client`. CI was green because containers don't have the production daemon running.

Set `MEM0_DIR` to a per-run `tempfile.mkdtemp` at `tests/conftest.py` module-top so the override is in `os.environ` before pytest collects `test_memory.py` and triggers its top-level `import mem0`. A fixture-scoped override would be too late: `mem0_dir` is frozen into `mem0.memory.main`'s namespace at first import via the chain `mem0/__init__.py:6` → `mem0/memory/main.py:25` → `mem0/memory/setup.py:8`.

`atexit` cleans up the per-run temp dir at interpreter shutdown.

## Regression test

Adds `TestInitMemory::test_mem0_telemetry_path_is_isolated` with three assertions:

1. `MEM0_DIR` is set - catches conftest simplification that drops the override.
2. Not `$HOME/.mem0` - catches a literal-default replacement.
3. Matches `mem0.memory.setup.mem0_dir` - catches the load-ordering trap where the env var is set after `import mem0` has already cached the wrong path. This is the freeze-detector that pinpoints the failure mode the spec is preventing.

Gated by `@integration` so CI without the `[memory]` extra skips rather than errors on the `mem0.memory.setup` import.

## Test plan

- [x] All 7 originally-failing tests now pass with the production Kai service running on the same machine
- [x] New `test_mem0_telemetry_path_is_isolated` passes
- [x] Full suite green: 2082 passed (up from 2072 - 9 newly-passing + 1 new regression test)
- [x] `$HOME/.mem0/migrations_qdrant` mtime unchanged before and after the suite (still `Apr 16 10:28:25`, the production service's last touch)
- [x] `/tmp/kai-test-mem0-*` removed by `atexit` after the process exits

fixes #357

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -5,9 +5,61 @@ Fixtures here apply to ALL tests automatically, providing safety
 guarantees that individual tests don't need to remember to set up.
 """
 
+# ── Session-wide Mem0 isolation ─────────────────────────────────────
+#
+# Mem0 opens a telemetry/migration-tracking Qdrant client at
+# `$MEM0_DIR/migrations_qdrant` (see mem0/memory/main.py:378-379 and
+# the parallel construction at line 1819). That path is hardcoded
+# outside the user's `vector_store.config` and defaults to
+# `$HOME/.mem0/migrations_qdrant`. If the production Kai service is
+# running on the same machine, it holds a portalocker lock on that
+# folder and any test-suite run racing against it dies with
+# `Storage folder ... is already accessed by another instance of
+# Qdrant client`. See issue #357.
+#
+# The path is resolved at module-import time in
+# `mem0/memory/setup.py:8` (and again, independently, in
+# `mem0/configs/base.py:13`), which means the override must be in
+# `os.environ` BEFORE any test module imports `mem0`.
+# `tests/test_memory.py:80` does `import mem0` at module top to
+# detect availability, so a fixture-level override is too late -
+# pytest collects the test module after conftest runs but before
+# any fixture body executes. conftest.py module-top is the only
+# hook that fires before test collection, making this the right
+# place for the env-var set.
+import atexit
+import os
+import shutil
+import tempfile
 from unittest.mock import patch
 
 import pytest
+
+# Hard override (NOT setdefault). A dev who happens to run the
+# suite with `MEM0_DIR=$HOME/.mem0` exported in their shell is the
+# exact scenario this fix targets; respecting an inherited value
+# would silently reopen the bug. tempfile.mkdtemp is used instead
+# of pytest's tmp_path_factory because that factory is only
+# accessible from inside fixtures - module-top code has no clean
+# way to reach it without poking at private pytest internals.
+_MEM0_TEST_DIR = tempfile.mkdtemp(prefix="kai-test-mem0-")
+os.environ["MEM0_DIR"] = _MEM0_TEST_DIR
+
+
+@atexit.register
+def _cleanup_mem0_test_dir() -> None:
+    """Best-effort removal of the per-run Mem0 telemetry root.
+
+    `tempfile.mkdtemp` does not auto-clean; without this hook every
+    `make test` invocation leaks one directory under /tmp/. Errors
+    are swallowed (ignore_errors=True) so a stray open file handle
+    from a crashed worker cannot mask a real test failure with a
+    secondary traceback at interpreter shutdown.
+    """
+    shutil.rmtree(_MEM0_TEST_DIR, ignore_errors=True)
+
+
+# ── Per-test history isolation ──────────────────────────────────────
 
 
 @pytest.fixture(autouse=True)
--- a/tests/test_memory.py
+++ b/tests/test_memory.py
@@ -142,6 +142,46 @@ class TestInitMemory:
         assert not is_enabled()
 
     @integration
+    def test_mem0_telemetry_path_is_isolated(self):
+        """Mem0's hardcoded telemetry Qdrant path must live under the
+        test-only MEM0_DIR set by conftest, never under $HOME/.mem0.
+
+        Regression guard for issue #357. Mem0 opens a second Qdrant
+        local-mode client at $MEM0_DIR/migrations_qdrant for its own
+        telemetry/migration tracking, with the path resolved at module
+        import time (mem0/memory/setup.py:8). If MEM0_DIR is not set
+        before any `import mem0` runs, that path freezes to the
+        production default $HOME/.mem0 and collides with the running
+        production service's portalocker lock - making the entire
+        TestMemoryIntegration suite unrunnable on a dev machine.
+
+        If a future refactor removes the conftest-level env override,
+        this test fails fast on CI and on dev machines alike, before
+        the heavier integration tests hit the RuntimeError.
+        """
+        import os
+        from pathlib import Path
+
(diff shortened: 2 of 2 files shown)


dcellison:
## Review by Kai

**Overall: Clean fix, well-reasoned. A few minor observations below.**

---

**Suggestion - conftest.py, `tempfile.mkdtemp` call (line ~43)**

If `mkdtemp` fails (full disk, permission error), it throws at conftest import time. The failure message will be an unhelpful OSError with no indication it came from test setup. Consider wrapping with a more informative error:

```python
try:
    _MEM0_TEST_DIR = tempfile.mkdtemp(prefix="kai-test-mem0-")
except OSError as e:
    raise RuntimeError(f"Could not create Mem0 isolation dir: {e}") from e
```

Low probability, but would save debugging time if it ever hits.

---

**Suggestion - test_memory.py, local `import os` inside test method (line ~155)**

`import os` at the top of `test_mem0_telemetry_path_is_isolated` is redundant if `os` is already imported at module level in `test_memory.py` (the diff doesn't show the file header, so I can't confirm). If it is, ruff won't flag it as an error but it's noise. Worth checking.

---

**Suggestion - atexit with pytest-xdist**

If the suite ever gains parallel workers via `pytest-xdist`, each worker process imports conftest.py independently and calls `mkdtemp`, creating `n_workers + 1` temp dirs (including the orchestrator). Each cleans up its own via atexit. This is fine as-is - just worth knowing if parallelism is added later so someone doesn't wonder why multiple `kai-test-mem0-*` dirs exist mid-run.

---

**Logic: correct.** The three-assertion regression test is well-structured - assertion 3 (the freeze-detector via `mem0.memory.setup.mem0_dir`) is the load-bearing one and catches exactly the subtle ordering bug the PR is preventing. The `setdefault` vs hard-override decision is correctly motivated.

**Security: none.** Pure test infrastructure, no user input, no external exposure.

**Conventions: clean.** Thorough commenting matches the project standard. `make check` reportedly passes. No direct-push to main.

dcellison:
## Review by Kai

**Overall: Clean. No new issues beyond what the prior review covered.**

The prior review raised three points - none were addressed, which per the instructions I won't re-raise. Here is my read of the current diff only:

---

**Suggestion - test_memory.py, `@integration` gate on assertions 1 and 2**

Assertions 1 and 2 (`MEM0_DIR` is set; not pointing at `$HOME/.mem0`) have no dependency on mem0 being installed - only assertion 3's `from mem0.memory.setup import mem0_dir` does. Gating all three on `@integration` means the conftest override goes untested on any CI tier that skips the `[memory]` extra. Minor: the current design still catches the regression wherever mem0 is installed, which is the machine where the bug manifests.

---

**Logic: correct.** The hard override (not `setdefault`) is the right call and well-motivated. The atexit pattern is correct - `@atexit.register` on a no-arg function is valid, `ignore_errors=True` is appropriate for cleanup code, and LIFO atexit ordering means this runs after pytest's own hooks without interfering.

**Ordering guarantee: sound.** Module-top conftest sets `os.environ["MEM0_DIR"]` before any test module is collected, which is the only hook that satisfies the constraint from `mem0/memory/setup.py:8` freezing the path at first import.

**Security: none.** Pure test infrastructure, no external exposure, no user-controlled input.

**Conventions: clean.** Comment density is at or above the project standard, which the conventions require.

PR #361, opened 2026-04-22
fix(memory): remove Track 1 raw ingestion; add user-message delimiter

### LINKED ISSUE #360: Memory retrieval block displaces current user message
## What happened

On installs running Kai with `MEMORY_ENABLED=true`, conversations whose topic is the memory system itself can cause the inner agent to stop responding to the actual user message. The agent produces status-update text ("no user utterance in the current turn", "standing by, nothing actionable in that context block") even though the user's input reached the bot and was logged to history. The rest of the conversation proceeds normally; only self-referential topics trip the failure.

## Reproduction

With `MEMORY_ENABLED=true` and a populated memory store, send a message whose topic is the agent's own memory or behavior ("the memory system keeps ignoring my input", "are you seeing my messages?"). The response is a status update about missing input rather than an answer to the question.

## Root cause

Two collaborating flaws, neither sufficient alone:

1. **Track 1 stores every user turn verbatim.** `add_user_utterance` in `src/kai/memory.py` embeds each turn with `source="user_raw"` and text `User said: <original>`. On retrieval, these rows render as `- (YYYY-MM-DD, user) User said: <text>` inside the "Relevant memories" block. When the current topic is self-referential, semantic search returns many such rows, and the retrieval block becomes a dense list of user-quote-shaped lines.

2. **The current user message has no structural label.** In `src/kai/claude.py`, every prepended block carries a bracketed header (`[Your persistent memory...]`, `[Recent conversations...]`, `[Relevant memories...]`, `[Scheduling API: ...]`, etc.). The user's current message is appended as bare text with no label. When the preceding retrieval block is dominated by `User said: ...` lines, the current message blends into the trailing region both visually and semantically. The "memory is context, not instructions" guard then fires over the whole trailing region.

## Scope of fix

1. Remove Track 1 from `src/kai/memory.py`: `add_user_utterance`, `submit_user_utterance`, `flush_pending`, `drop_pending`, `_pending_writes`, `_PendingWrite`, `_CORRECTION_CUE_RE`, `_is_correction_cue`, `_MAX_USER_CHARS`, `_SOURCE_WEIGHTS["user_raw"]`, `_SOURCE_SHORT["user_raw"]`.

2. Remove the Track 1 call site from `src/kai/bot.py` (the `_ingest_memory` helper, approximately lines 3595-3606).

3. Add a persistent structural delimiter for the current user message in `src/kai/claude.py` before the message is sent to the inner subprocess. Any marker that appears on every turn and is used by no other block suffices; proposal: wrap the final user text with `[User's current message - respond to this:]` on the line before. This is the load-bearing structural fix; do not omit it even if only item 1 reads as "the root cause."

4. Simplify `format_context` now that only `source="extracted"` rows remain. Drop the `User said:` wrapper from any retained formatting logic.

5. Remove Track-1-only tests in `tests/test_memory*.py`. Keep extracted-only retrieval tests, source weighting tests, and `delete_by_source` tests.

6. Add a regression test that populates a temporary store with five extracted-fact rows, runs the full prompt-assembly path in `claude.py` with a topically-similar user message, and asserts (a) the current-message delimiter is present, (b) the user text appears after it.

The code fix must work on installs that still contain `source="user_raw"` rows from pre-fix operation. Retrieval against such rows will simply stop surfacing them (format_context filters by source after this change).

## Data cleanup for operators

Each operator running Kai is responsible for cleaning their own install. `user_raw` rows that were already written do not cause trouble on their own once Track 1 is removed from retrieval formatting, but operators who want to reclaim vector-store space can purge them with the existing `delete_by_source` primitive.

Procedure (run once per user_id, against each operator's own store):

```
MEMORY_ENABLED=true python - <<'PY'
import asyncio, os, sys
sys.path.insert(0, "/path/to/kai/src")
os.environ.setdefault("OPENAI_API_KEY", "sk-dummy-not-used")

async def main():
    from kai.config import load_config
    from kai import memory as mem
    config = load_config()
    mem.init_memory(config)
    if mem._memory is None:
        print("memory init failed"); return
    # List user_ids for your install (Telegram chat IDs, typically).
    for user_id in ("<chat_id_1>", "<chat_id_2>"):
        deleted = await mem.delete_by_source(user_id, "user_raw")
        print(f"{user_id}: deleted {deleted} rows")

asyncio.run(main())
PY
```

Safe to run while memory is disabled; it force-enables for the one-shot. Only touches rows whose metadata source equals `user_raw`; extracted facts are untouched.

## Mitigation before the fix lands

Operators seeing the failure mode on their installs can mitigate in one of two ways:

- **Disable memory** (set `MEMORY_ENABLED=false` and restart). Eliminates the retrieval block entirely. The cleanest mitigation; recommended for operators who are not actively using memory yet.
- **Periodic purge** of `user_raw` rows via the script above. Reduces the chance that any given turn's retrieval will be dominated by user-quote lines. Does not fully prevent the bug because a recently-stored single turn can still collide.

Neither mitigation is a substitute for the code fix.

## References

- Epic #306 (memory system)
- Earlier decision to drop Track 1 (April 2026), which specified `delete_by_source` as the cleanup primitive but did not land the code removal or the delimiter change.

dcellison:
## Summary

Implements spec 360. Removes Track 1 raw user ingestion and adds a structural delimiter for the current user message, fixing the failure mode where the inner Claude refused to respond to memory-adjacent turns.

## What changed

- **Track 1 deletion**: removed `add_user_utterance`, `submit_user_utterance`, `flush_pending`, `drop_pending`, `_pending_writes`, `_PendingWrite`, `_CORRECTION_CUE_RE`, `_is_correction_cue`, the `user_raw` entries in `_SOURCE_WEIGHTS` and `_SOURCE_SHORT`, plus the verification-window machinery and the surrounding stale comment prose.
- **`USER_MESSAGE_MARKER`** added in `claude.py` and prepended FIRST in the prompt chain, so the existing `session_ctx`, `memory_ctx`, and `reminder` prepends stack above it. The marker ends up as the closest prefix to the user's actual text — the layout invariant the spec exists to enforce.
- **`_MAX_USER_CHARS`** moved from `memory.py` to `memory_extraction.py` (now its sole consumer).
- **`_end_session`** simplified to a thin wrapper around `sessions.clear_session` (the Track 1 flush call is gone; nothing else lives at session-end yet).
- **Tests**: deleted five Track 1 test classes, rewrote five `TestMemoryIntegration` tests to use `add_structured` (the surviving Track 2 ingestion primitive), trimmed the source-weighting test from a 3-way to a 2-way ordering, added `test_format_context_extracted_only` (post-Track-1 retrieval has no `User said:` lines), `test_prompt_contains_current_message_delimiter`, and `test_delimiter_is_closest_prefix_to_user_text` (regression test for the spec evaluation's B-1 finding).

Net change: 331 insertions, 802 deletions across 7 files.

## Verification

- `make check` passes (ruff check + format).
- `make test` passes (2037 tests, including the 5 real-Mem0 integration tests).
- Acceptance #2 grep clean: zero matches for the Track 1 symbols across `src/kai/`; zero `user_raw` matches in `memory.py`, `bot.py`, `claude.py`, `memory_extraction.py`, `webhook.py`, `cron.py`. `memory_admin.py`'s `_KNOWN_SOURCES` and `--source` help still list `user_raw` per the spec exemption (operators must still be able to purge legacy rows).
- Acceptance #3: cleaned up the `user_raw` references in the `_SOURCE_WEIGHTS` block-comment and in the `delete_by_source` / `get_by_tag` docstring prose.
- The B-1 regression test asserts (a) marker appears exactly once, (b) all three other context blocks appear before the marker, and (c) only whitespace separates the marker from the user's actual text.

## Test plan

- [x] `make check` clean
- [x] `make test` passes (2037 tests)
- [x] All three new `test_claude.py` tests pass
- [x] Real-Mem0 integration suite passes after the `add_structured` rewrites
- [ ] Manual smoke per acceptance #4: re-enable memory, send three memory-adjacent messages, confirm normal responses (post-merge step)
- [ ] Manual delimiter-layout verification per acceptance #5 (local-only, not in the diff)

fixes #360

### DIFF
--- a/src/kai/bot.py
+++ b/src/kai/bot.py
@@ -339,34 +339,17 @@ async def handle_start(update: Update, context: ContextTypes.DEFAULT_TYPE) -> No
 
 
 async def _end_session(chat_id: int) -> None:
+    """Session-end hook: clear the session record.
+
+    Spec 360 removed the Track 1 verification-window flush that used to
+    run here. Track 2 extraction (`memory_extraction.extract_and_store`)
+    is fire-and-forget per turn and has no end-of-session drain step,
+    so the only thing left to do at session end is the session-state
+    cleanup itself. Kept as a thin wrapper rather than inlining the
+    `sessions.clear_session` call at every site so future session-end
+    hooks (typing-indicator drain, metric emission, etc.) have a
+    natural home.
     """
-    Session-end hook: flush any pending memory write, then clear the
-    session record.
-
-    Implements the Phase 3 session-end half of spec §6.3 P3. When
-    `memory_extraction_enabled` is off (the default through Phase 2
-    and Phase 3 ship), the flush call still runs but `flush_pending`
-    is a cheap dict-pop + no-op when there is nothing queued. The
-    memory check avoids importing the memory layer at all when it is
-    globally disabled, which also avoids loading Mem0's optional deps
-    on installs that do not have them.
-
-    Failures in the flush are logged at WARNING and never propagate:
-    session-clear MUST always proceed so the bot does not wedge. This
-    mirrors the try/except pattern around _ingest_memory.
-    """
-    from kai.memory import flush_pending
-    from kai.memory import is_enabled as memory_is_enabled
-
-    if memory_is_enabled():
-        try:
-            await flush_pending(str(chat_id))
-        except Exception:
-            # Never block a session clear on a memory hiccup. The
-            # pending dict entry (if any) has already been popped by
-            # flush_pending before the Mem0 write, so a crash here
-            # does not leave orphaned state.
-            log.warning("flush_pending failed during session end", exc_info=True)
     await sessions.clear_session(chat_id)
 
 
@@ -3577,7 +3560,6 @@ async def _handle_response(
         async def _ingest_memory() -> None:
             try:
                 from kai import memory_extraction
-                from kai.memory import submit_user_utterance
 
                 # Extract user text from prompt. For multimodal prompts
                 # (image + text), pull the first text block rather than
@@ -3589,26 +3571,22 @@ async def _handle_response(
                         (block["text"] for block in prompt if block.get("type") == "text"),
                         "",
                     )
-                # Skip image-only exchanges - no meaningful text to embed
+                # Skip image-only exchanges - no meaningful text to embed.
                 if not user_text:
                     return
-                # Track 1: user-only raw embedding, gated by the Phase 3
-                # verification window. `submit_user_utterance` queues the
-                # turn instead of flushing it immediately; the previous
-                # turn's queued write either flushes now (non-correction)
-                # or gets dropped (correction cue). The assistant reply
-                # is never stored here. Any assistant-derived facts come
-                # from Phase 2's extractor below.
-                await submit_user_utterance(
-                    user_text=user_text,
-                    user_id=str(chat_id),
-                    session_id=final_response.session_id,
-                )
 
                 # Track 2: Haiku extraction. Runs only under the Claude
                 # backend and only when explicitly enabled. Fire-and-
                 # forget INSIDE the existing fire-and-forget task - the
                 # subprocess latency never blocks reply delivery.
+                #
+                # Spec 360 removed the previous Track 1 raw-user write
+                # that used to run alongside this call. The verbatim
+                # user storage was producing retrieval blocks dense
+                # with `User said:` lines that mimicked real user input
+                # and confused the inner agent on memory-adjacent
+                # topics. Extracted facts (Track 2) remain the only
+                # write path.
                 #
                 # Backend check uses the same per-user fall-through
                 # pattern as bot.py:370 (user override wins, else global).
--- a/src/kai/bot.py
+++ b/src/kai/bot.py
@@ -3589,9 +3589,10 @@ async def _handle_response(
                 # write path.
                 #
                 # Backend check uses the same per-user fall-through
-                # pattern as bot.py:370 (user override wins, else global).
-                # A global-only check would miss users with a per-user
-                # override, so the explicit fall-through is mandatory here.
+                # pattern as `_get_user_provider` (user override wins,
+                # else global). A global-only check would miss users
+                # with a per-user override, so the explicit fall-through
+                # is mandatory here.
                 user_config = config.get_user_config(chat_id)
                 effective_backend = (
                     user_config.agent_backend if user_config and user_config.agent_backend else config.agent_backend
--- a/src/kai/bot.py
+++ b/src/kai/bot.py
@@ -341,14 +341,13 @@ async def handle_start(update: Update, context: ContextTypes.DEFAULT_TYPE) -> No
 async def _end_session(chat_id: int) -> None:
     """Session-end hook: clear the session record.
 
-    Spec 360 removed the Track 1 verification-window flush that used to
-    run here. Track 2 extraction (`memory_extraction.extract_and_store`)
+    The previous Track 1 verification-window flush that used to run
+    here is gone. Track 2 extraction (`memory_extraction.extract_and_store`)
     is fire-and-forget per turn and has no end-of-session drain step,
     so the only thing left to do at session end is the session-state
     cleanup itself. Kept as a thin wrapper rather than inlining the
-    `sessions.clear_session` call at every site so future session-end
-    hooks (typing-indicator drain, metric emission, etc.) have a
-    natural home.
+    `sessions.clear_session` call because it has multiple call sites
+    (grep `_end_session(`); inlining would touch every one.
     """
     await sessions.clear_session(chat_id)
 
--- a/src/kai/claude.py
+++ b/src/kai/claude.py
@@ -41,6 +41,26 @@ from kai.config import DATA_DIR, WorkspaceConfig, parse_env_file, resolve_claude
 log = logging.getLogger(__name__)
 
 
+# Persistent structural delimiter for the current user message. Spec
+# 360 §1.2 / §3.3: when retrieval blocks contain quote-shaped lines
+# that mimic real user input (legacy `User said:` rows or sufficiently
+# user-voiced extracted facts), the inner agent can fail to recognize
+# the trailing user text as the actual message. This marker is the
+# one structural signal that says "the message below this line is
+# the real one — respond to it." Module-level so tests can import
+# and assert against it without string duplication.
+#
+# Bracket-label format chosen to match the visual shape of the other
+# context blocks (`[Recent conversations ...]`, `[Your persistent
+# memory ...]`, `[Relevant memories ...]`) without colliding with any
+# of them — a retrieval row could in principle contain any of those
+# header strings verbatim, but cannot accidentally produce this exact
+# label. NOT gated by a feature flag: spec §5.1 requires the marker
+# to be a permanent fixture so prompt shape is consistent
+# session-to-session.
+USER_MESSAGE_MARKER = "[User's current message - respond to this:]"
+
+
 # ── Claude Code backend ─────────────────────────────────────────────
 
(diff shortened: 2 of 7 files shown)


dcellison:
## Review by Kai

This PR is well-structured and largely clean. The core logic is sound. Here are my findings:

---

**SUGGESTION** - `memory_extraction.py`, `_MAX_USER_CHARS` comment (line ~57)

> "that one stays in memory.py because it is referenced from `_build_extraction_payload` **below** alongside legacy comment context"

"Below" implies `_build_extraction_payload` is in `memory.py` below `_MAX_ASSISTANT_CHARS`, but it's actually in `memory_extraction.py` (the current file). The sentence means "referenced by `_build_extraction_payload` over in `memory_extraction.py`," but the word "below" actively misleads. Worth a one-word fix: "referenced from `_build_extraction_payload` in this module."

---

**SUGGESTION** - `tests/test_memory.py`, `test_format_context_extracted_only` (line ~609)

```python
assert "fact)" in line, f"non-(fact) tag on line: {line!r}"
```

`"fact)" in line` would also match if a memory text itself contained the substring `"fact)"` (e.g. a stored fact about "a known fact)"). The check passes here only because the seeded memory texts are controlled. A tighter assertion like `re.search(r'\((?:\d{4}-\d{2}-\d{2}, )?fact\)', line)` would be more precise. Low risk given the test controls its own data, but worth noting for future test data changes.

---

**SUGGESTION** - `tests/test_memory.py`, integration test `test_format_context_integration` (line ~1808)

The test method is `async def` but now calls `mem_mod.add_structured(...)` synchronously (previous version was `await mem_mod.add_user_utterance(...)`). This is correct if `add_structured` is sync, but the change is invisible in this diff since `add_structured` isn't shown. If `add_structured` were ever made async, this call would silently return a coroutine without awaiting it. No current bug, but a documentation gap.

---

**No issues found in:**
- The Track 1 deletion itself - complete and consistent (`re` import removed, `_pending_writes` fixture cleanup removed, all five deleted test classes appropriately removed)
- The `USER_MESSAGE_MARKER` ordering logic - the prepend-first-to-land-closest invariant is correct, and `test_delimiter_is_closest_prefix_to_user_text` is a well-designed regression guard
- `_end_session` simplification - thin wrapper rationale is sound
- `_MAX_USER_CHARS` migration - value unchanged, only consumer updated, cross-module `_MAX_ASSISTANT_CHARS` asymmetry is acknowledged and justified
- Legacy `user_raw` row handling - falls through to `_UNKNOWN_SOURCE_WEIGHT` / `"legacy"` label correctly
- Security posture - no new injection surfaces, no auth changes, no data exposure introduced

The PR does what it says. The three findings above are all clarity/robustness suggestions, not bugs.

dcellison:
## Review by Kai

This PR is clean. The prior review's three suggestions were all addressed: the `tag_re` regex replaces the `"fact)" in line` substring check, the misleading "below" word was removed, and the `add_structured` sync contract is documented in the integration test docstring. No findings from prior rounds need re-raising.

---

**SUGGESTION** - `bot.py`, comment inside `_ingest_memory` (around the `# Backend check uses the same per-user fall-through` line)

The comment references `bot.py:370` as an anchor for the backend-check pattern. The Track 1 deletion removed ~30 lines above that point, so the line number is likely stale. Minor, but stale line-number references in comments tend to mislead during future archaeology.

---

**SUGGESTION** - `memory_extraction.py`, `_MAX_USER_CHARS` comment (line ~57-65)

The sentence "that one stays in memory.py because `_build_extraction_payload` in this module reads it alongside legacy comment context tied to memory.py" is harder to parse than the parallel explanation on the `memory.py` side, which reads "kept here on the assistant side because moving it would be churn unrelated to spec 360." The two explanations say the same thing but the `memory_extraction.py` version conflates "reads it" with "tied to context" in a way that requires two passes. Worth aligning with the cleaner phrasing across the street.

---

Nothing else to raise. The Track 1 deletion is complete and consistent (`re` import gone, `_pending_writes` fixture teardown removed, all five test classes deleted, all six acceptance checks described in the PR description accounted for). The `USER_MESSAGE_MARKER` ordering invariant is sound - prepend-first lands closest, subsequent prepends stack above - and `test_delimiter_is_closest_prefix_to_user_text` is a well-constructed regression guard for the inverted form. The `kai.memory.format_context` patch target in that test is correct for a lazy-import alias. No bugs, no security issues.

dcellison:
## Review by Kai

This PR is clean. The prior round's suggestions were all addressed - the stale `bot.py:370` line number is replaced with `_get_user_provider`, the `tag_re` regex replaces the `"fact)" in line` substring check, and the `memory_extraction.py` comment is rewritten cleanly.

No new findings to raise.

---

**Confirmed correct:**

- `USER_MESSAGE_MARKER` prepend ordering in `claude.py:417` - first prepend lands closest to user text because each subsequent prepend pushes it down; `test_delimiter_is_closest_prefix_to_user_text` is a well-constructed regression guard for the inverted form.
- `_MAX_USER_CHARS` migration to `memory_extraction.py:57` - value unchanged (2000), sole consumer updated, cross-file asymmetry with `_MAX_ASSISTANT_CHARS` is documented and justified.
- `re` import removal from `memory.py` is correct (the only consumer was `_CORRECTION_CUE_RE`, now deleted); `re` import addition in `test_memory.py` is correct (new `tag_re` usage).
- `_reset_memory_module` fixture cleanup in `tests/test_memory.py:62` - `_pending_writes.clear()` correctly removed along with the attribute itself.
- `test_format_context_integration` sync/async contract - pinned in the docstring with a clear "what breaks if this changes" note. Adequate for now.
- `kai.memory.format_context` patch target in `test_delimiter_is_closest_prefix_to_user_text` - correct for a lazy-import alias resolved at call time.
- Track 1 deletion completeness - `re` import gone, five test classes deleted, fixture teardown line removed, all six acceptance checks from the PR description accounted for.

dcellison:
## Review by Kai

This PR is clean. Three prior review rounds addressed all findings; nothing new to raise.

Verified correct:

- **Track 1 deletion is complete** - `re` import removed from `memory.py`, `dataclass`/`field` retained (still needed for `MemoryResult`), `_pending_writes.clear()` removed from the fixture. `ruff` would have caught any stale import, and the PR reports `make check` clean.

- **`USER_MESSAGE_MARKER` prepend ordering** (`claude.py:417`) - prepend-first lands closest because each subsequent prepend pushes it down toward the user text. `test_delimiter_is_closest_prefix_to_user_text` correctly asserts all three context blocks have a smaller index than the marker, and the whitespace-only `between` check is tight.

- **`_MAX_USER_CHARS` migration** (`memory_extraction.py:57`) - value unchanged at 2000, `_build_extraction_payload` is the sole consumer, reference updated from `memory._MAX_USER_CHARS` to the module-local. `test_long_user_text_truncated` imports from the new location.

- **Legacy `user_raw` row handling** - falls through `_SOURCE_WEIGHTS.get(source, _UNKNOWN_SOURCE_WEIGHT)` to the unknown-source default and through `_SOURCE_SHORT.get(source, "legacy")` to `"legacy"`. Rankable but not preferentially boosted. Correct retrieval-time behavior for rows written by earlier code.

- **Integration test sync/async contract** - `add_structured` is sync, not awaited, and the docstring on `test_format_context_integration` pins this explicitly with a "what breaks if this changes" callout.

- **`test_format_context_extracted_only`** - `tag_re = re.compile(r'\((?:\d{4}-\d{2}-\d{2}, )?fact\)')` is the tighter form recommended in round 1. Correct placement inside `TestFormatContext` class (4-space indent, replaces the removed `TestDropPending` block in the same indentation context).

- **`memory_extraction.py` comment** - "below" is accurate: `_MAX_USER_CHARS` is at line 57, `_build_extraction_payload` is at line 356 in the same file.

No bugs, no security issues, no convention violations.

Issue #362, opened 2026-04-22
Evaluate the semantic memory system


dcellison:
Umbrella issue for measuring whether the semantic memory system (#306, now closed) produces measurable improvements over running without memory.

## Motivation

With #306 complete, Kai has a working semantic memory pipeline: Haiku-based fact extraction into a Qdrant vector store, semantic retrieval, context injection, and a `/memory` browsing command. What is missing is an answer to the question: does any of this measurably improve Kai's behavior compared to running without memory?

A memory system that works structurally but does not move the needle on response quality, factual recall, or user friction is not worth the complexity it adds. Conversely, a measured win is the only thing that distinguishes "we shipped a feature" from "we shipped an improvement." This epic exists to produce that measurement.

## Evaluation approach

Three layers, each targeting a different silent-failure mode in the pipeline. Any single-layer evaluation averages over failure classes and produces a number without an actionable diagnosis.

### Layer 1: Retrieval-only

Tests whether `MemoryStore.search()` returns the correct fact for a given query. Pure retrieval-quality metric: precision@K, recall@K, mean reciprocal rank. Requires a labeled probe set of `(question, expected-fact-id)` pairs drawn from facts already in an operator's store. Isolates the vector-search layer from any downstream generation effects.

This is the cheapest layer to run and the only one that cleanly separates "retrieval miss" from "generator ignored the retrieved fact."

### Layer 2: End-to-end behavioral A/B

Tests whether the full pipeline (retrieve plus inject plus generate) produces responses that better reflect known facts. Run a fixed probe set twice: once with `MEMORY_ENABLED=true`, once with `MEMORY_ENABLED=false`. Score paired responses with an LLM judge against a rubric that references the ground-truth facts. Report win, tie, and loss rates.

Medium cost. Catches the case where retrieval returns the right fact but the generator does not use it, or vice versa.

### Layer 3: Longitudinal friction analysis

Observational, not experimental. Grep an operator's chat history for memory-relevant frictions: repeated-fact frustrations ("I already told you"), preference corrections Kai should have known, and Kai asking for information previously given. Bucket by week.

Natural bucket boundaries are the public milestones that changed what memory did for everyone running the project:

- Before #333 merged (2026-04-17): no semantic memory; facts only in `MEMORY.md`.
- Between #333 and #361 (2026-04-17 to 2026-04-22): semantic memory live, Track 1 raw-utterance ingestion active, no current-message delimiter.
[...]

PR #364, opened 2026-04-22
feat(memory): structured logging for retrieval

### LINKED ISSUE #363: Structured logging for memory retrieval
Part of #362 (memory evaluation epic). Prerequisite for all three evaluation layers.

## What

Emit a structured log line per retrieval call in `src/kai/memory.py`'s `format_context` describing the retrieval result: `chat_id`, query length, per-hit (`source`, `score`, truncated snippet), and `latency_ms`.

## Why

Retrieval is currently silent. There is no way for an operator or an eval harness to see what the memory system is returning for a given query. The epic's Layer 1 (retrieval-only metrics) and Layer 2 (end-to-end A/B attribution) both depend on this visibility. Without it, a failed retrieval and a correctly-retrieved-but-ignored fact look identical from logs.

## Scope

- One structured log emit per retrieval call.
- Fields: `chat_id`, `query_len`, per-hit (`source`, `score`, truncated snippet), `latency_ms`.
- Exact schema, PII redaction policy for queries and snippets, and log level will be pinned in an implementation spec before code work begins.

## Out of scope

- Retrieval algorithm changes.
- New eval tooling (sibling issues under #362).
- Log aggregation or dashboarding.

## Acceptance

- One structured log line per retrieval call.
- One test verifying log format.
- No behavior change in what `format_context` returns.

dcellison:
## Summary

Adds a single structured log line per call to `format_context` in `src/kai/memory.py`. The line carries a stable `memory.recall` prefix followed by a compact JSON payload, designed so a downstream retrieval eval harness can parse retrieval results for precision/recall scoring without re-running search.

Pure observability change: `format_context` returns the same string it did before. Existing `TestFormatContext` tests pass unmodified.

## Schema

Every emitted line carries the same JSON keys (uniform shape) so parsers don't need to branch on field presence. Sentinel values (`0` / `0.0` / `[]`) cover fields whose real values aren't yet known at a given short-circuit. The single non-uniform field is `reason`: present only on short-circuit lines, omitted on success.

Emit sites: 6 total (5 short-circuit returns + 1 success), each carrying its own `reason` value (`disabled`, `empty_query`, `no_results`, `all_below_floor`, `budget_exhausted`).

`hits` reflects every floor-survivor in final post-sort order even when the budget walk cuts the prompt short. `hits[:lines_used]` are the entries the agent saw; `hits[lines_used:]` survived the floor but were dropped by `budget_tokens`. The eval harness needs that distinction to score precision honestly.

## PII posture

The `query` field and per-hit `snippet` field are logged in their truncated form (80 chars), unhashed and unredacted. The target log file is operator-local and already contains the full conversation under other paths (`CLAUDE_DEBUG=true`, history storage), so this introduces no new category of sensitive data. Pattern-based redaction is fragile and creates a false sense of safety; truncation is the only meaningful guardrail. Recorded inline in `_emit_recall_log` so a future reviewer doesn't re-litigate the decision.

## Tests

Three new test methods under `TestRecallLogging`:

1. `test_memory_recall_log_success_has_all_fields` — success path emits exactly one line with every uniform field, no `reason` key, and the post-sort order in `hits` matches the source-weighting rule.
2. `test_memory_recall_log_uniform_shape_on_short_circuit` — parametrized across all five short-circuit reasons; asserts uniform schema, `returned_empty=true`, and the expected `reason` string. Catches regressions where two paths accidentally collapse to the same reason value.
3. `test_memory_recall_log_snippet_and_query_truncated_and_sanitized` — exercises 80-char cap and `\n`/`\r` sanitization on both the query field and per-hit snippets.

`latency_ms >= 0` is the only timing assertion; no real timeouts in tests.

fixes #363

## Test plan

- [x] `make test` passes (2044 tests)
- [ ] CI green
- [ ] Smoke-test post-merge: send a memory-adjacent message in production and confirm one `memory.recall` line appears per turn in `/var/lib/kai/logs/kai.log`

### DIFF
--- a/src/kai/memory.py
+++ b/src/kai/memory.py
@@ -265,8 +265,12 @@ def _truncate(s: str, n: int) -> str:
 
     Used by the memory.recall logging path to render both the query
     text and per-hit snippets as a single line of human-readable JSON.
-    Newline rewriting must happen BEFORE truncation so that a CR/LF
-    landing exactly at position n - 1 cannot leak past the cap.
+    Sanitize-first is the canonical idiom: it makes "no control
+    characters in the output" an invariant of this function regardless
+    of any future change to the truncation step. The length cap itself
+    holds with either ordering, since `\n` and `\r` are single chars
+    replaced 1-for-1 with spaces; the choice is structural, not
+    length-correctness.
 
     No ellipsis: the truncation length is a known constant, and the
     eval harness consuming these log lines treats the snippet as a
@@ -527,6 +531,21 @@ async def format_context(
     # ThreadPoolExecutor to avoid blocking the event loop.
     fetch_limit = max(_config.memory_search_limit, _SEARCH_OVERFETCH)
     payload["fetch_limit"] = fetch_limit
+
+    # Read the floor from config at every call (not at module import) so a
+    # `MEMORY_SEARCH_FLOOR` change applied via service restart takes effect
+    # consistently here AND in the `/memory search` UI. _config is non-None
+    # inside this branch because is_enabled() returned True above.
+    #
+    # Captured BEFORE the search call so post-search short-circuit log
+    # lines (no_results, all_below_floor) carry the real floor value in
+    # their payload rather than the 0.0 sentinel that
+    # _base_recall_payload sets, matching how budget_tokens and
+    # fetch_limit are populated up-front. The actual filter step using
+    # `floor` happens later, after the search returns results to filter.
+    floor = _config.memory_search_floor
+    payload["floor"] = floor
+
     loop = asyncio.get_running_loop()
     # latency_ms scopes ONLY the run_in_executor search call: the
     # ranking and formatting that follow are deterministic, microsecond
@@ -544,14 +563,9 @@ async def format_context(
     # Quality gate: drop low-relevance noise before any ranking adjustment.
     # Weighting happens AFTER this filter so a downweighted legacy row
     # cannot survive on raw score, and a boosted extracted fact cannot
-    # be rescued below threshold.
-    #
-    # Read the floor from config at every call (not at module import) so a
-    # `MEMORY_SEARCH_FLOOR` change applied via service restart takes effect
-    # consistently here AND in the `/memory search` UI. _config is non-None
-    # inside this branch because is_enabled() returned True above.
-    floor = _config.memory_search_floor
-    payload["floor"] = floor
+    # be rescued below threshold. `floor` was read from _config above
+    # so the all_below_floor short-circuit log line carries the real
+    # threshold rather than a sentinel.
     results = [r for r in results if r.score >= floor]
     payload["hits_after_floor"] = len(results)
     if not results:
--- a/src/kai/memory.py
+++ b/src/kai/memory.py
@@ -291,7 +291,7 @@ _RECALL_REASON_ALL_BELOW_FLOOR = "all_below_floor"
 _RECALL_REASON_BUDGET_EXHAUSTED = "budget_exhausted"
 
 
-def _base_recall_payload(user_id: str, query: str) -> dict:
+def _base_recall_payload(user_id: str, query: str) -> dict[str, object]:
     """
     Build a memory.recall payload with all uniform-shape fields set.
 
@@ -325,7 +325,7 @@ def _base_recall_payload(user_id: str, query: str) -> dict:
     }
 
 
-def _emit_recall_log(payload: dict) -> None:
+def _emit_recall_log(payload: dict[str, object]) -> None:
     """
     Write a single memory.recall log line as compact JSON.
 
--- a/src/kai/memory.py
+++ b/src/kai/memory.py
@@ -22,8 +22,10 @@ via pyproject.toml's [memory] extra).
 from __future__ import annotations
 
 import asyncio
+import json
 import logging
 import os
+import time
 from dataclasses import dataclass, field
 
 from kai.config import DATA_DIR, Config
@@ -247,6 +249,102 @@ def _estimate_tokens(text: str) -> int:
     return len(text) // 4
 
 
+# Maximum characters of query and per-hit snippet to embed in the
+# memory.recall log line. 80 is enough to fingerprint a hit for an
+# eval harness (which treats the snippet as an identifier, not as
+# full retrieved text) while keeping the log line within a single
+# terminal width and avoiding the ingestion of multi-paragraph user
+# messages verbatim. Single source of truth so query and snippet
+# stay in lockstep.
+_RECALL_TRUNC = 80
+
+
+def _truncate(s: str, n: int) -> str:
+    """
+    Sanitize newlines/CRs to single spaces, then truncate to n chars.
+
+    Used by the memory.recall logging path to render both the query
+    text and per-hit snippets as a single line of human-readable JSON.
+    Newline rewriting must happen BEFORE truncation so that a CR/LF
+    landing exactly at position n - 1 cannot leak past the cap.
+
+    No ellipsis: the truncation length is a known constant, and the
+    eval harness consuming these log lines treats the snippet as a
+    fingerprint rather than as full text. Appending "..." would mislead
+    a human reader into thinking it's the actual stored text.
+    """
+    return s.replace("\n", " ").replace("\r", " ")[:n]
+
+
+# Reasons surfaced in the `reason` field of a memory.recall log line.
+# Each maps 1:1 to a short-circuit return site in format_context.
+# Centralized here so test_memory.py's parametrized short-circuit
+# test can import the same set without re-typing magic strings, and
+# so a typo in any single emit site is caught by the type checker
+# rather than silently producing a unique third reason value.
+_RECALL_REASON_DISABLED = "disabled"
+_RECALL_REASON_EMPTY_QUERY = "empty_query"
+_RECALL_REASON_NO_RESULTS = "no_results"
+_RECALL_REASON_ALL_BELOW_FLOOR = "all_below_floor"
+_RECALL_REASON_BUDGET_EXHAUSTED = "budget_exhausted"
+
+
+def _base_recall_payload(user_id: str, query: str) -> dict:
+    """
+    Build a memory.recall payload with all uniform-shape fields set.
+
+    The schema is uniform across every emit site: every key listed
+    here is present on every memory.recall log line, with sentinel
+    values (0 / 0.0 / []) for fields whose real values are not yet
+    known at the short-circuit point. This is a deliberate choice so
+    downstream parsers (the retrieval eval harness and operator
+    grep/jq one-liners) can treat every line under one schema rather
+    than branching on `if "fetch_limit" in record`.
+
+    `returned_empty` defaults to True; the success path flips it to
+    False just before emit. `reason` is the one non-uniform field:
+    present only on short-circuit lines, omitted on success — so it's
+    not in the base payload. Real query and user_id are always
+    populated; they are never sentineled.
+    """
+    return {
+        "user_id": user_id,
+        "query_len": len(query),
+        "query": _truncate(query, _RECALL_TRUNC),
+        "fetch_limit": 0,
+        "hits_raw": 0,
+        "hits_after_floor": 0,
+        "floor": 0.0,
+        "latency_ms": 0,
+        "returned_empty": True,
+        "lines_used": 0,
+        "budget_tokens": 0,
+        "hits": [],
+    }
+
+
+def _emit_recall_log(payload: dict) -> None:
+    """
+    Write a single memory.recall log line as compact JSON.
+
+    Compact separators (no spaces) are required: the eval harness
+    parses these lines via simple grep + json.loads, so multi-line
+    output would break extraction. The "memory.recall " prefix is
+    the stable greppable tag.
+
+    PII posture (recorded here so a future reviewer does not
+    re-litigate it): the query and per-hit snippets are logged in
+    their truncated form (80 chars), unhashed and unredacted. Target
+    log file is operator-local and already contains the full
+    conversation under other paths (CLAUDE_DEBUG=true, history
+    storage). Pattern-based redaction is fragile and creates a false
+    sense of safety; truncation is the only meaningful protection
+    here. Operators wanting stricter handling should rotate logs
+    aggressively or shift the log level for this module.
+    """
+    log.info("memory.recall %s", json.dumps(payload, separators=(",", ":")))
+
+
 # ── Public API ──────────────────────────────────────────────────────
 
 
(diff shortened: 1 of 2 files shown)


dcellison:
## Review by Kai

## PR Review: feat(memory) - structured logging for retrieval

Overall the PR is well-implemented. Emit-once guarantee holds across all 6 paths, sentinel values are consistent, `_truncate` correctly sanitizes before truncating (not after), and the `hits` snapshot before the budget walk is the right design for the eval harness. No critical issues.

---

### Warning

**W1 - Test doesn't exercise source-weighting rank inversion** (`tests/test_memory.py`, success test)

The mock uses extracted `score=0.9`, legacy `score=0.7`. Both raw order (0.9 > 0.7) and adj order (1.08 > 0.42) already put extracted first, so the sort assertion passes even if `_source_weight` is disabled or returns 1.0 for both sources. To actually test that source weighting can override raw score, the legacy hit needs a higher raw score than the extracted hit (e.g., legacy 0.95, extracted 0.90 - adj would be 0.57 vs 1.08).

**W2 - Spec cross-references removed as collateral edits** (`src/kai/memory.py`)

Three existing spec references were dropped while reformatting nearby lines: `spec §5.3` (weighting-after-filter rule), `spec 310 §7.5` (one-knob-two-paths decision), and `spec §5.4` (source tag format). This is out of scope for a pure observability change and may matter to future maintainers tracing spec compliance.

---

### Suggestion

**S1 - `import json as _json` inside helper function** (`tests/test_memory.py`, `_parse_recall_log`)

The import belongs at module level, not inside a helper function. The alias adds no clarity.

**S2 - `r.message` vs `r.getMessage()`** (`tests/test_memory.py`, `_parse_recall_log`)

`LogRecord.message` is only set as a side effect of `Formatter.format()`; `r.getMessage()` is the stable documented API. pytest caplog does populate `message` in practice so this works, but `r.getMessage()` is more explicit and doesn't rely on formatter side effects.

dcellison:
## Review by Kai

The file isn't in the local tree, so I'll work from the diff directly.

---

## Review: feat(memory) - structured logging for retrieval

Prior items W1, S1, S2 are addressed in this revision. W1's fix (swapping raw scores to 0.95/0.90 so only source weighting can flip the order) is correct. S1 (module-level `import json`) and S2 (`getMessage()`) are both done.

No new critical or warning-level issues. Two suggestions:

---

### Suggestion

**S1 - Misleading comment in success test** (`tests/test_memory.py`, `test_memory_recall_log_success_has_all_fields`, line near the end of the test body)

The final comment reads:

> "so the extracted hit must come first in the hits array even though Mem0 returned them with the higher raw score **on the extracted entry** by coincidence"

This is backwards. In the mock, legacy has the higher raw score (0.95) and extracted has the lower one (0.90). The comment should say "even though Mem0 returned legacy with the higher raw score." As written it contradicts the preceding explanation and would confuse anyone verifying the test's purpose.

**S2 - Redundant local `import logging` inside test methods** (`tests/test_memory.py`, all three `async def` test bodies)

`logging` is almost certainly already imported at module level in this file (it's used in `caplog.at_level`). Re-importing it locally adds noise. Minor tidying.

---

### Everything else is clean

- Emit-once guarantee: confirmed across all 6 paths - payload is built once, mutated in-place, emitted at the return site.
- `latency_ms` sentinel (0) is correct only for `disabled`/`empty_query` paths; the measured value is set before `no_results`/`all_below_floor`/`budget_exhausted` emit.
- `hits` snapshot before the budget walk is the right design; `hits[:lines_used]` / `hits[lines_used:]` split is sound.
- `_truncate` sanitizes before truncating, not after - correct.
- `json.dumps(payload, separators=(",", ":"))` keeps lines single-line for grep/jq parsers - correct.
- `budget_exhausted` test case (budget=10, header ~17 tokens) will correctly never add a line and trigger the short-circuit.
- PII posture is documented in `_emit_recall_log` inline; no new data category introduced.

dcellison:
## Review by Kai

Looking at the diff after two prior review cycles. The items from both prior reviews are fully addressed. Three minor findings remain.

---

## Findings

### Suggestion

**S1 - `_source_weight(r)` called twice per result** (`src/kai/memory.py`, around line 563 and the list comprehension below it)

`_source_weight(r)` is computed once in the `sorted()` key and again in the `payload["hits"]` list comprehension. If the weight function is cheap and pure this is harmless inefficiency, but it's also easy to avoid by zipping the pre-computed weights:

```python
weighted = [(r, _source_weight(r)) for r in results]
weighted.sort(key=lambda t: t[0].score * t[1], reverse=True)
payload["hits"] = [
    {"source": ..., "score": round(r.score, 4), "adj": round(r.score * w, 4), "snippet": ...}
    for r, w in weighted
]
```

Not a bug given `_source_weight` is deterministic, but the duplication is the kind of thing that bites if the function ever gains a side effect or logging of its own.

---

**S2 - `budget_exhausted` test comment is inaccurate** (`tests/test_memory.py`, setup branch for `budget_exhausted`)

The comment says `"huge text overflows any small budget"` but the header alone is ~70 chars (~17 tokens via `_estimate_tokens`), which already exceeds `budget=10`. The for loop never executes, so `"x" * 10000` has no effect on the outcome. The test is correct, but any memory text (even `"hi"`) would trigger the same path. The comment should say the header alone exhausts the budget, not the memory text.

---

**S3 - `import kai.memory as mem_mod` / `from kai.memory import format_context` inside every test body** (`tests/test_memory.py`, all three async methods in `TestRecallLogging`)

These imports repeat in each method body where a module-level import would do. Not a correctness issue, but it's at odds with the module-level `import json` and `import logging` added in this same PR.

---

## Everything else is clean

- Emit-once guarantee holds across all 6 paths; no double-emit path visible.
- `_truncate` sanitizes before slicing - correct.
- Sentinel values are consistent; `returned_empty=True` default is correct.
- `payload["hits"]` snapshot precedes the budget walk - the `hits[:lines_used]` / `hits[lines_used:]` split is sound.
- `reason` absent on success, present on every short-circuit - schema is uniform where it should be.
- `getMessage()` (not `.message`) in `_parse_recall_log` - correct stable API.
- Prior S1 (misleading success-test comment) and S2 (local `import logging`) from the second review are both resolved.
- No injection vectors; `json.dumps` handles escaping; no new PII category introduced.

dcellison:
## Review by Kai

Looking at the diff after three prior review cycles. Items S3 from review 3 (module-level imports in test methods) was seen by the author and not addressed; I will not re-raise it.

---

## Review: feat(memory) - structured logging for retrieval

The implementation is sound. Emit-once guarantee holds across all 6 paths, sentinel values are consistent, `_truncate` sanitizes before slicing, `hits` snapshot precedes the budget walk, and the `weighted` list is materialized by `sorted()` so the second iteration in `results = [r for r, _ in weighted]` is safe.

One new finding.

---

### Suggestion

**S1 - `_RECALL_REASON_*` comment claims test import that never happens** (`src/kai/memory.py`, comment block above `_RECALL_REASON_DISABLED`)

The comment reads:

> Centralized here so test_memory.py's parametrized short-circuit test can import the same set without re-typing magic strings

The parametrized test uses literal strings throughout (`"disabled"`, `"empty_query"`, etc.) - both in the `@pytest.mark.parametrize` decorator and in the `assert payload["reason"] == reason` assertion. The constants are never imported or referenced in the tests. The comment's stated rationale is false, which makes the comment actively misleading rather than just inaccurate.

The real benefit of these constants is preventing a typo at any single emit site from producing a silently unique reason value - which is worth keeping. The comment should just say that, and drop the claim about test import.

---

### Everything else is clean

- `budget_tokens`, `fetch_limit`, `hits_raw`, `latency_ms`, and `floor` are populated at the correct points in the control flow; each short-circuit emits only the fields that are known at that point, with sentinels for the rest.
- `_source_weight(r)` called once per result via the `weighted` tuples - prior S1 fully resolved.
- `budget_exhausted` test comment accurately describes the header-exhausts-budget mechanism - prior S2 fully resolved.
- `getMessage()` used in `_parse_recall_log` - correct stable API.
- `json.dumps(..., separators=(",", ":"))` keeps each line single-line for grep/jq - correct.
- No injection vectors; `json.dumps` handles escaping; no new PII category introduced.

dcellison:
## Review by Kai

After four rounds of review, this PR is in good shape. I'll focus only on findings not previously raised.

---

## Review: feat(memory) - structured logging for retrieval

All items from prior reviews have been addressed (S1 from review 4 - the misleading `_RECALL_REASON_*` comment about test import - is corrected in this diff). No critical or warning-level issues.

---

### Suggestion

**S1 - Bare `dict` return type on `_base_recall_payload`** (`src/kai/memory.py`, line ~295)

```python
def _base_recall_payload(user_id: str, query: str) -> dict:
```

The project uses Pyright strict mode. An unparameterized `dict` annotation resolves to `dict[Unknown, Unknown]` under strict, which defeats type narrowing for any consumer that tries to use the returned value with typed expectations. `dict[str, object]` would be the minimal correct annotation and would also serve as lightweight documentation that keys are always strings.

---

### Everything else is clean

- Emit-once guarantee holds across all 6 paths; no double-emit path.
- `_source_weight(r)` computed once per row via the `weighted` tuples; `payload["hits"]` snapshot precedes the budget walk - both prior `_source_weight` double-call and snapshot-timing concerns are fully resolved.
- `_truncate` sanitizes before slicing, not after - correct.
- `latency_ms` sentinel (0) is correct for `disabled`/`empty_query` paths where the search call is never reached; real value is measured for the remaining four paths.
- `_RECALL_REASON_*` comment no longer claims test import - the misleading text from review 4's S1 is gone.
- `budget_tokens=0` sentinel for `empty_query` is consistent with the "not yet computed at this point in the control flow" contract.
- `getMessage()` in `_parse_recall_log` - correct stable API.
- `json.dumps(..., separators=(",", ":"))` keeps each line single-line - correct.
- `hits[:lines_used]` / `hits[lines_used:]` split is sound for eval harness precision/recall scoring.
- No injection vectors; `json.dumps` handles escaping; no new PII category introduced.

dcellison:
## Review by Kai

Looking at the diff after five prior review cycles. The S1 from round 5 (`dict[str, object]` return type on `_base_recall_payload`) is addressed. Two new findings, both suggestions.

---

## Review: feat(memory) - structured logging for retrieval

### Suggestion

**S1 - `_truncate` docstring's "leak past the cap" claim is inaccurate** (`src/kai/memory.py`, `_truncate` docstring)

> Newline rewriting must happen BEFORE truncation so that a CR/LF landing exactly at position n - 1 cannot leak past the cap.

This is not correct. `\n` and `\r` are single characters; replacing them with single spaces leaves the string length unchanged. Whether you sanitize-then-slice or slice-then-sanitize, the output is always at most `n` characters - the cap is respected either way. "Leak past the cap" has no coherent meaning here.

The real reason the ordering matters is simpler: sanitize-then-slice guarantees no control characters appear anywhere in the output, regardless of position. Slice-then-sanitize also works for length, but "sanitize first" is still the right idiom and deserves a correct explanation.

---

**S2 - `floor` stays at sentinel 0.0 for the `no_results` path even though it is computable** (`src/kai/memory.py`, around line 526 in the diff)

`budget_tokens` and `fetch_limit` are both populated before the `no_results` check fires (they are computed from `_config` which is guaranteed non-None at that point). `floor = _config.memory_search_floor` is also available at that point - the code just computes it later. As a result, `no_results` log lines carry `"floor": 0.0` (sentinel) while `"budget_tokens"` and `"fetch_limit"` carry real values. An eval harness operator looking at a `no_results` log line cannot tell what the floor threshold was in effect during that run.

Either populate `floor` before the search call (consistent with `budget_tokens`/`fetch_limit`), or document in the PR description that the `no_results` sentinel for `floor` is intentional.

---

### Everything else is clean

- Emit-once guarantee holds across all 6 paths; payload is local to each call, no cross-call mutation risk.
- `_source_weight(r)` computed once per row via `weighted` tuples - double-call from round 3 is fully resolved.
- `payload["hits"]` snapshot precedes the budget walk; `hits[:lines_used]` / `hits[lines_used:]` split is sound.
- `_truncate` sanitizes before slicing (correct character-set guarantee even if the length argument above about ordering is wrong in the docstring).
- `budget_tokens=0` sentinel for `disabled`/`empty_query` paths is consistent with "not computed at this point."
- `latency_ms` sentinel (0) is correct for `disabled`/`empty_query` paths; real measured value for the three post-search short-circuits.
- `dict[str, object]` on `_base_recall_payload` - round 5 S1 is resolved.
- `getMessage()` in `_parse_recall_log` - correct stable API.
- `json.dumps(..., separators=(",", ":"))` keeps each line single-line for grep/jq - correct.
- `_RECALL_REASON_*` comment no longer claims test import - round 4 S1 is resolved.
- Module-level `import json` and `import logging` in test file - rounds 2-3 suggestions resolved.
- No injection vectors; `json.dumps` handles escaping; no new PII category introduced.

dcellison:
## Review by Kai

Looking at this diff after six prior review rounds. Both S1 and S2 from round 6 are resolved: the `_truncate` docstring now accurately explains that sanitize-first is a structural choice rather than a length-correctness requirement, and `floor` is populated from `_config` before the search call so `no_results` short-circuit lines carry the real threshold.

No new critical or warning-level issues.

---

### Suggestion

**S1 - `_emit_recall_log` eagerly serializes JSON regardless of log level** (`src/kai/memory.py`, `_emit_recall_log`)

```python
log.info("memory.recall %s", json.dumps(payload, separators=(",", ":")))
```

Python evaluates `json.dumps(payload, ...)` before calling `log.info`, so the serialization cost is paid on every `format_context` call even when the `kai.memory` logger is set to WARNING or above. The logger's lazy `%s` interpolation only defers the string interpolation step, not the argument evaluation. For a small fixed-schema dict this is microseconds, so it is not a bug, but it is worth knowing about if operators ever crank up log verbosity to suppress these lines (e.g. for PII reasons): the serialization still runs. A guard like `if log.isEnabledFor(logging.INFO):` would make the suppression path truly zero-cost.

---

### Everything else is clean

- Round 6 S1 (`_truncate` docstring "leak past the cap" claim) fully resolved; the new wording is accurate.
- Round 6 S2 (`no_results` floor sentinel) fully resolved; `floor` is now read and assigned to `payload["floor"]` before the `run_in_executor` call, matching `budget_tokens` and `fetch_limit` placement.
- Emit-once guarantee holds across all 6 paths; payload is local to each call.
- `_source_weight` computed once per row via `weighted` tuples; no double-call.
- `payload["hits"]` snapshot precedes the budget walk; `hits[:lines_used]` / `hits[lines_used:]` split is sound.
- `getMessage()` in `_parse_recall_log` - correct stable API.
- `dict[str, object]` return type on `_base_recall_payload` - round 5 concern resolved.
- `_RECALL_REASON_*` comment no longer claims test import - round 4 concern resolved.
- `budget_tokens=0` sentinel for `disabled`/`empty_query` paths is consistent with "not computed at this point in the control flow."
- `latency_ms` sentinel (0) correct for `disabled`/`empty_query`; real measured value for the three post-search paths.
- `json.dumps(..., separators=(",", ":"))` keeps each line single-line for grep/jq.
- No injection vectors; `json.dumps` handles escaping; no new PII category introduced.

dcellison:
Thanks for the catch on the eager-evaluation point - the technical observation is correct (`json.dumps` runs before `log.info` regardless of the lazy `%s`). Declining the suggested guard, though, for three reasons:

**1. The suppression scenario contradicts the feature's purpose.**

This logging exists to feed the retrieval eval harness in #362. The contract is "every `format_context` call emits exactly one `memory.recall` line." Operators raising the `kai.memory` logger above INFO to suppress these lines would defeat the eval harness; the feature was built assuming INFO-level capture is on. Optimizing the suppression path is optimizing a configuration we built the feature to require against.

**2. The cost is genuinely below the noise floor.**

`format_context` runs once per Telegram turn, not in a hot loop. `json.dumps` on the ~12-key fixed-schema payload is on the order of 5-20 microseconds. The embedding search that just ran in the same function takes 50-200 milliseconds - five orders of magnitude more. Even at 1000 turns/day (well above realistic single-user volume), the total serialization cost is ~20 ms/day. There is no workload where this is observable.

**3. The guard adds defensive code that protects nothing in practice.**

`if log.isEnabledFor(level):` is a worthwhile pattern when the deferred formatting is genuinely expensive: rendering a stack trace, pickling a model, walking a deep tree. As a reflex applied to small fixed-schema dicts it becomes cargo-culted: every future reader has to interpret it as "what edge case is this guarding?" and the answer ("operators turning off the logger we require for the eval harness") doesn't justify the cognitive cost.

If a future change makes the payload meaningfully expensive to serialize (e.g. embedding the raw search results for richer logs), the guard becomes worth adding at that point. As of this PR, it's not.

PR #366, opened 2026-04-22
feat(eval): Layer 1 retrieval evaluation harness

### LINKED ISSUE #365: Layer 1 probe set and retrieval-eval harness
Sub-issue under #362. Builds the Layer 1 retrieval-only evaluation described in the parent epic.

## Summary

Given a probe set of labeled `(question, expected_fact_id)` pairs, run each probe through the memory system and measure whether vector search surfaces the expected fact. Emit precision@K, recall@K, and mean reciprocal rank. Support a sweep mode that re-runs the probe set across a grid of retrieval parameters and surfaces the Pareto frontier for tuning.

Prerequisite #363 (structured retrieval logging) is merged; the harness consumes the `memory.recall` log lines emitted there rather than re-running retrieval in a parallel code path.

## Why this layer first

Layer 1 is the cheapest of the three evaluation layers and the only one that cleanly separates "retrieval miss" from "generator ignored the retrieved fact." It runs entirely offline, with no LLM-judge calls and no live traffic. If retrieval itself cannot surface the right fact for an obvious probe, the higher layers measure noise downstream of a broken base.

Layer 1 also produces the first objective tuning signal for `memory_search_floor`, the source-weight table in `_SOURCE_WEIGHTS`, and the `_SEARCH_OVERFETCH` constant, all currently set by informed guess rather than measurement.

## Scope

- **Probe set**: a versioned collection of `(question, expected_fact_id)` pairs. Coverage target: at least one probe per extracted-fact tag present in the store. Authored so the set measures retrieval quality rather than re-confirming what retrieval already happens to get right.
- **Harness**: for each probe, runs retrieval via `format_context`, parses the corresponding `memory.recall` log line, and scores hits against the labeled answer. Outputs precision@K, recall@K, mean reciprocal rank.
- **Config sweep**: re-runs the probe set across a grid of retrieval parameters (floor, source weights, overfetch) and reports the Pareto frontier.
- **Baseline score**: committed reference metrics for current defaults so future changes have a regression anchor.

## Out of scope

- Layer 2 (end-to-end behavioral A/B with `MEMORY_ENABLED` toggle). Separate sub-issue under #362.
- Layer 3 (longitudinal friction analysis). Separate sub-issue.
- Changes to `format_context` or any retrieval internals beyond what the harness needs to consume.
- Fixes for whatever the evaluation surfaces. Per the epic's scope boundary, iterations on retrieval quality go under a future epic.

## Acceptance

- Harness runs from the command line against a real memory store and emits precision@K, recall@K, mean reciprocal rank.
- Config sweep mode operational, with Pareto-frontier output.
- Baseline score file committed to the repo as a regression anchor.
- Probe set covers at least one extracted-fact tag per bucket present in the operator's store.
- Null results are acceptable per the parent epic's scope.

Parent: #362. Depends on: #363 (merged).

dcellison:
fixes #365

## What

Adds a Layer 1 retrieval-eval harness at `src/kai/eval/retrieval.py`, runnable as `python -m kai.eval.retrieval`. Given a JSONL probe set of `(question, expected_fact_id)` pairs, it runs each probe through the live `format_context` pipeline, parses the structured `memory.recall` log line, and reports precision@K, recall@K, MRR, `fraction_in_prompt`, and p50/p95 latency.

Two modes:

- **Single config**: scores against whatever `_config` / `_SOURCE_WEIGHTS` / `_SEARCH_OVERFETCH` the live memory module currently carries; an `--output` flag writes a baseline JSON file matching the committed `*.example.json` schema.
- **Sweep** (`--sweep`): iterates a `(floor, extracted_weight, overfetch)` grid by mutating module-level state per iteration with `try/finally` restoration, then prints the Pareto frontier on `(precision@5, latency_p50)` followed by the full table sorted best-quality-first.

## Why

Without this, "did that retrieval tweak help?" is judged eyeball-only. The harness gives a reproducible answer per probe set, and the baseline JSON makes regression detection a `jq` diff away. Per-tag breakdowns surface category-level quality (e.g. preferences vs. infra) so a tweak that helps overall but hurts one category is visible.

## Drift handling

Probes whose expected fact has been deleted or filtered out by `get_by_id` (ownership mismatch, source mismatch, or never-existed) are bucketed as drift and excluded from BOTH the numerator and the denominator. `N_scored = N_probes - N_drift`. Mixing drift with retrieval misses would conflate "retrieval failed" with "probe set went stale" - different operator actions.

## Prerequisite (commit 1 of 2)

Adds an `id` field to each per-hit dict in the `memory.recall` payload, sourced from `MemoryResult.id`. The harness's ranking logic depends on it; otherwise scoring would have to fall back to fragile snippet-substring matching against the 80-char truncated text. One existing test asserts the per-hit key set; it's updated in the same commit.

## Files

- `src/kai/eval/__init__.py` - package init (room for future Layer 2/3 here)
- `src/kai/eval/retrieval.py` - harness + CLI (~1,100 lines)
- `tests/test_eval_retrieval.py` - 16 tests across 6 classes
- `home/evals/probes.example.jsonl` - schema demo, three synthetic probes
- `home/evals/retrieval-baseline.example.json` - baseline file shape
- `.gitignore` - real probe set + baseline gitignored, examples tracked
- `src/kai/memory.py` + `tests/test_memory.py` - prerequisite `id` field

## Why probes are gitignored

Real probe questions are conversation excerpts, and `expected_fact_id` values are Mem0 row IDs specific to one operator's store. Both are operator data; only the synthetic `*.example.*` fixtures ship in the repo.

## Test plan

- [ ] `make check` (ruff lint + format)
- [ ] `make test` (full suite, including new `tests/test_eval_retrieval.py`)
- [ ] `python -m kai.eval.retrieval --probes home/evals/probes.example.jsonl --user-id <chat_id>` smoke run against a real store
- [ ] `python -m kai.eval.retrieval --probes ... --user-id ... --sweep` smoke run on a small grid

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,6 +43,13 @@ home/.claude/*
 !home/config
 home/config/*
 !home/config/goose-config.yaml
+# Evaluation harness: committed examples show the schema; real
+# baselines and probe sets contain per-operator data (conversation
+# fragments, IDs specific to that store) and must stay gitignored.
+!home/evals
+home/evals/*
+!home/evals/retrieval-baseline.example.json
+!home/evals/probes.example.jsonl
 
 # OS / IDE
 .DS_Store
--- a/home/evals/probes.example.jsonl
+++ b/home/evals/probes.example.jsonl
@@ -1,0 +1,11 @@
+# Example probe set for the Layer 1 retrieval eval harness.
+# Lines starting with `#` are comments and skipped by the loader.
+# Real probes live in home/evals/probes.jsonl (gitignored) so each
+# operator's actual conversation excerpts stay out of the repo.
+
+# probes added 2026-04-22 covering tag=preferences
+{"question": "what timezone am I in?", "expected_fact_id": "00000000-0000-0000-0000-000000000001", "source_turn_ts": "2026-04-01T10:15:00", "notes": "from session covering travel scheduling"}
+{"question": "which editor do I use?", "expected_fact_id": "00000000-0000-0000-0000-000000000002", "source_turn_ts": "2026-04-03T19:42:00", "notes": "asked while debugging a config tweak"}
+
+# probes added 2026-04-22 covering tag=infra
+{"question": "where does the bot run?", "expected_fact_id": "00000000-0000-0000-0000-000000000003", "source_turn_ts": "2026-04-08T08:00:00", "notes": "from a deployment-status check"}
--- a/home/evals/retrieval-baseline.example.json
+++ b/home/evals/retrieval-baseline.example.json
@@ -1,0 +1,39 @@
+{
+  "version": 1,
+  "generated_at": "2026-04-22T15:00:00Z",
+  "probe_set_hash": "sha256:0000000000000000000000000000000000000000000000000000000000000000",
+  "probe_count": 47,
+  "drift_count": 2,
+  "config": {
+    "floor": 0.3,
+    "extracted_weight": 1.2,
+    "legacy_weight": 0.6,
+    "overfetch": 20
+  },
+  "metrics": {
+    "precision_at_1": 0.68,
+    "precision_at_3": 0.81,
+    "precision_at_5": 0.83,
+    "precision_at_lines_used": 0.77,
+    "mrr": 0.74,
+    "fraction_in_prompt": 0.77,
+    "latency_p50_ms": 43,
+    "latency_p95_ms": 127
+  },
+  "by_tag": {
+    "preferences": {
+      "precision_at_k": {"1": 0.71, "3": 0.84, "5": 0.86},
+      "recall_at_k": {"1": 0.71, "3": 0.84, "5": 0.86},
+      "precision_at_lines_used": 0.79,
+      "mrr": 0.78,
+      "fraction_in_prompt": 0.79
+    },
+    "infra": {
+      "precision_at_k": {"1": 0.62, "3": 0.75, "5": 0.78},
+      "recall_at_k": {"1": 0.62, "3": 0.75, "5": 0.78},
+      "precision_at_lines_used": 0.72,
+      "mrr": 0.69,
+      "fraction_in_prompt": 0.72
+    }
+  }
+}
--- a/home/evals/retrieval-baseline.example.json
+++ b/home/evals/retrieval-baseline.example.json
@@ -24,14 +24,22 @@
   },
   "by_tag": {
     "preferences": {
-      "precision_at_k": {"1": 0.71, "3": 0.84, "5": 0.86},
-      "recall_at_k": {"1": 0.71, "3": 0.84, "5": 0.86},
+      "precision_at_1": 0.71,
+      "precision_at_3": 0.84,
+      "precision_at_5": 0.86,
+      "recall_at_1": 0.71,
+      "recall_at_3": 0.84,
+      "recall_at_5": 0.86,
       "mrr": 0.78,
       "fraction_in_prompt": 0.79
     },
     "infra": {
-      "precision_at_k": {"1": 0.62, "3": 0.75, "5": 0.78},
-      "recall_at_k": {"1": 0.62, "3": 0.75, "5": 0.78},
+      "precision_at_1": 0.62,
+      "precision_at_3": 0.75,
+      "precision_at_5": 0.78,
+      "recall_at_1": 0.62,
+      "recall_at_3": 0.75,
+      "recall_at_5": 0.78,
       "mrr": 0.69,
       "fraction_in_prompt": 0.72
     }
--- a/home/evals/retrieval-baseline.example.json
+++ b/home/evals/retrieval-baseline.example.json
@@ -14,7 +14,9 @@
     "precision_at_1": 0.68,
     "precision_at_3": 0.81,
     "precision_at_5": 0.83,
-    "precision_at_lines_used": 0.77,
+    "recall_at_1": 0.68,
+    "recall_at_3": 0.81,
+    "recall_at_5": 0.83,
     "mrr": 0.74,
     "fraction_in_prompt": 0.77,
     "latency_p50_ms": 43,
@@ -24,14 +26,12 @@
     "preferences": {
       "precision_at_k": {"1": 0.71, "3": 0.84, "5": 0.86},
       "recall_at_k": {"1": 0.71, "3": 0.84, "5": 0.86},
-      "precision_at_lines_used": 0.79,
       "mrr": 0.78,
       "fraction_in_prompt": 0.79
     },
     "infra": {
       "precision_at_k": {"1": 0.62, "3": 0.75, "5": 0.78},
       "recall_at_k": {"1": 0.62, "3": 0.75, "5": 0.78},
-      "precision_at_lines_used": 0.72,
       "mrr": 0.69,
       "fraction_in_prompt": 0.72
     }
--- a/src/kai/eval/__init__.py
+++ b/src/kai/eval/__init__.py
@@ -1,0 +1,9 @@
+"""
+Evaluation package for the Kai memory pipeline.
+
+Currently houses Layer 1 only (`retrieval`): precision / recall / MRR
+on a probe set scored against the live `format_context` log line.
+Future layers (end-to-end A/B, longitudinal friction) will land as
+sibling modules so they share probe loaders, output formatting, and
+the same package home.
+"""
--- a/src/kai/eval/retrieval.py
+++ b/src/kai/eval/retrieval.py
@@ -406,8 +406,13 @@ def score_probes(results: list[ProbeResult]) -> dict[str, Any]:
     }
 
 
-def _percentile(values: list[float], pct: float) -> float:
-    """Return the requested percentile from values (0-100 scale).
+def _percentile(values: list[float], pct: int) -> float:
+    """Return the requested percentile from values (1-99 integer scale).
+
+    `pct` is `int` not `float` so a fractional percentile (e.g. 50.9)
+    cannot be silently truncated to the 50th by `int(pct) - 1`. The
+    only callers today are `_percentile(latencies, 50)` and `(_, 95)`,
+    so narrowing the contract costs nothing and removes the footgun.
 
     Uses statistics.quantiles for the standard linear-interpolation
     estimate. Returns 0.0 on empty input rather than raising so the
@@ -419,13 +424,12 @@ def _percentile(values: list[float], pct: float) -> float:
         return 0.0
     if len(values) == 1:
         return float(values[0])
-    # n=100 partitions plus boundary fix-up gives us linear-interp
-    # percentile lookup at any integer percentage. quantiles returns
-    # the 99 cut points for n=100, so index `pct - 1` is the
-    # requested percentile (1-indexed math collapses to the simple
-    # subtraction).
+    # n=100 partitions gives 99 cut points (quantiles returns n-1
+    # cuts for n partitions), so index `pct - 1` is the requested
+    # 1-indexed percentile. clamp() defends against pct == 0 or
+    # pct >= 100, which would index out of the cuts list.
     cuts = statistics.quantiles(values, n=100)
-    idx = int(pct) - 1
+    idx = pct - 1
     idx = max(0, min(idx, len(cuts) - 1))
     return float(cuts[idx])
 
--- a/src/kai/eval/retrieval.py
+++ b/src/kai/eval/retrieval.py
@@ -184,7 +184,6 @@ class Metrics:
     n_drift: int
     precision_at_k: dict[int, float]
     recall_at_k: dict[int, float]
-    precision_at_lines_used: float
     mrr: float
     fraction_in_prompt: float
     latency_p50_ms: float
@@ -346,21 +345,17 @@ def score_probes(results: list[ProbeResult]) -> dict[str, Any]:
     """Compute aggregate metrics over a list of ProbeResult.
 
     Pure function: takes scored per-probe outcomes (rank, lines_used,
-    in_prompt) and returns the metrics dict. Test #1 exercises this
-    directly with hand-constructed inputs, including a mix of scored-
-    hit, scored-miss, and drift probes, asserting that drift probes
-    are excluded from BOTH numerator and denominator. The caller is
-    responsible for filtering out drift probes before invoking this
-    function; that contract is enforced by the type signature
-    (drift probes don't produce a ProbeResult at all).
+    in_prompt) and returns the metrics dict. The harness's drift-
+    detection step filters drifted probes out before they reach this
+    function; the type signature does NOT enforce that contract (any
+    list[ProbeResult] type-checks), so the caller is responsible.
 
     For the single-answer case (each probe has exactly one expected
     fact), recall@K equals precision@K. They are reported separately
     so a future multi-answer probe format can diverge without
-    reshaping the output. fraction_in_prompt differs from precision
-    @lines_used: in_prompt is computed per probe against that probe's
-    actual lines_used (varies with budget exhaustion), whereas
-    precision@K uses a fixed K across all probes.
+    reshaping the output. fraction_in_prompt is the per-probe variant:
+    each probe is checked against its own lines_used (varies with
+    budget exhaustion), whereas precision@K uses a fixed K.
     """
     n = len(results)
     if n == 0:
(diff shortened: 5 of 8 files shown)


dcellison:
## Review by Kai

I have the full diff in context and can review it directly.

---

## PR #366 Review: feat(eval) Layer 1 retrieval evaluation harness

Overall this is a well-structured and well-commented harness. A few real issues are worth fixing before merge.

---

### Warning

**`precision_at_lines_used` and `fraction_in_prompt` are numerically identical**
`retrieval.py`, `score_probes()` function.

Both fields compute the exact same expression:
```python
in_prompt_count = sum(1 for r in results if r.in_prompt)
...
"precision_at_lines_used": in_prompt_count / n,
"fraction_in_prompt":      in_prompt_count / n,   # same thing
```
`ProbeResult.in_prompt` is already computed per-probe as `rank is not None and rank <= lines_used` in `_run_one_probe`, so both fields draw from the same pre-computed boolean. The docstring's claim that they differ ("fraction_in_prompt differs from precision@lines_used") is incorrect. The baseline JSON, stdout render, and `_metrics_to_dict` all emit both, so operators will see two identically-valued fields in every output. One name should be dropped, or `precision_at_lines_used` should be computed inline as `sum(1 for r in results if r.rank is not None and r.rank <= r.lines_used) / n` if the intent was always per-probe K independent of `in_prompt`.

---

**`detect_drift` is called once per grid point during a sweep, not once total**
`retrieval.py`, `run_sweep()`.

`run_sweep` calls `evaluate(probes, user_id)` for each grid point; `evaluate` always calls `detect_drift`, which issues one `get_by_id` call per probe. For the default 72-point grid and a 47-probe set this is 3,384 `get_by_id` calls instead of 47. Drift is stable during a sweep (the store does not change in minutes). `detect_drift` should be hoisted before the loop:
```python
scored, drifted, tags_by_id = detect_drift(probes, user_id)
for i, override in enumerate(grid, start=1):
    ...
    metrics = await _evaluate_scored(scored, tags_by_id, ...)
```

---

**`recall_at_k` absent from baseline JSON top-level `metrics` block**
`retrieval.py`, `_build_baseline_json()`; `home/evals/retrieval-baseline.example.json`.

The PR description states the harness "reports precision@K, recall@K, MRR", and `_metrics_to_dict` (used for per-tag and sweep rows) does include `recall_at_k`. But `_build_baseline_json` constructs the top-level `metrics` block by hand and omits `recall_at_k` entirely. The `by_tag` sub-objects (populated via `_metrics_to_dict`) include it, so a reader of the baseline file sees `recall_at_k` only in per-tag breakdowns and nowhere at the top level. The example fixture matches the code but both are inconsistent with the stated interface.

---

### Suggestion

**`_render_single_metrics`: comment describes probe count that is never rendered**
`retrieval.py`, around line 870:
```python
# Per-tag probe count: count of probes whose expected fact carries this tag...
lines.append(f"  {tag}: p@5={p5:.3f} in_prompt={fip:.3f}")
```
The comment promises a count in the output; the format string does not include one. Either add `n={len(by_tag_bucket)}` to the render line or remove the misleading comment.

---

**`_render_sweep_table` uses bare dict index on `precision_at_k`**
```python
f"{m.precision_at_k[1]:>6.3f} {m.precision_at_k[3]:>6.3f} {m.precision_at_k[5]:>6.3f}  "
```
Raises `KeyError` if `_K_VALUES` is ever extended or reordered. Use `.get(k, 0.0)` to match the defensive pattern in `pareto_frontier`.

---

**Misleading docstring in `score_probes`**
"that contract is enforced by the type signature (drift probes don't produce a ProbeResult at all)" - the type `list[ProbeResult]` accepts any `ProbeResult`; the contract is convention, not type enforcement. Minor, but callers reading this may over-trust it.

---

### Clean

- Memory module `id` field addition and test update are correct; the prerequisite commit is minimal and honest.
- `_RecallLogCapture`, handler attach/detach, and the `try/finally` sweep restoration are all sound.
- Gitignore rules, probe file comment extension, and `probe_set_hash` sort-invariance are all handled correctly.
- Security posture is fine: no shell execution, no injection vectors, PII stays gitignored.
- Test coverage is solid across the six classes; the sweep-restoration test (class 3) is the most important and is well-constructed.

dcellison:
## Review by Kai

All six issues from the prior review have been addressed. Here's what I verified against the diff:

1. **`precision_at_lines_used` duplicate** - removed; `score_probes()` returns only `fraction_in_prompt`.
2. **`detect_drift` per grid point** - fixed; `detect_drift` is now called once before the loop in `run_sweep()`, and `_score_against_store()` is called per iteration with the pre-computed results.
3. **`recall_at_k` missing from baseline top-level** - fixed; `_build_baseline_json()` now includes `recall_at_1/3/5` alongside `precision_at_1/3/5`.
4. **Misleading `_render_single_metrics` comment** - gone; the per-tag loop just appends the format string with no comment promising a probe count.
5. **Bare dict index in `_render_sweep_table`** - fixed; all three reads now use `.get(1, 0.0)` / `.get(3, 0.0)` / `.get(5, 0.0)`.
6. **Misleading `score_probes` docstring** - corrected; now explicitly says "the type signature does NOT enforce that contract."

---

**One new finding:**

**Suggestion** - `_build_baseline_json()` vs `_metrics_to_dict()` schema inconsistency (`retrieval.py`, `_build_baseline_json`)

The top-level `metrics` block uses flat keys (`"precision_at_1"`, `"recall_at_3"`, etc.), but the `by_tag` sub-objects (delegated to `_metrics_to_dict`) use nested dicts (`"precision_at_k": {"1": v, "3": v, "5": v}`). The example fixture confirms this:

```json
"metrics": { "precision_at_1": 0.68, ... }   // flat
"by_tag":  { "preferences": { "precision_at_k": {"1": 0.71} } }  // nested
```

A jq script diffing two baselines needs different paths for the same metric depending on scope: `.metrics.precision_at_5` vs `.by_tag.preferences.precision_at_k["5"]`. It is not a bug, but it will trip up anyone writing a diff script and is worth flattening to match - or the docstring for `_build_baseline_json` should call it out explicitly.

---

Otherwise the code is clean. The sweep restoration test, drift-denominator contract, log capture attach/detach, and `run_sweep` try/finally semantics all look correct.

dcellison:
## Review by Kai

The prior two rounds of review caught the material issues and the fixes look correct. Here is what the current diff shows:

**Clean items (confirmed fixed):**
- `detect_drift` is hoisted before the sweep loop in `run_sweep` - one call per sweep, not per grid point.
- `recall_at_k` is present in the top-level `metrics` block of `_build_baseline_json`.
- `precision_at_lines_used` duplicate removed; only `fraction_in_prompt` remains.
- `_render_sweep_table` uses `.get(k, 0.0)` throughout.
- `_flatten_score_block` unifies both `metrics` and `by_tag` to flat keys, resolving the schema inconsistency raised in review 2. The example fixture matches.
- Sweep restoration test and try/finally semantics are sound.
- `id` field threaded through correctly; test 4 round-trips it end-to-end.

**One new finding:**

**Suggestion - `_percentile` silently truncates non-integer `pct` values** (`retrieval.py`, `_percentile`)

```python
idx = int(pct) - 1
```

`pct` is typed as `float` and only called with `50` and `95` today, so this is inert. But the function signature accepts any float. `pct=50.9` would return the 50th percentile, not the 51st. A caller reading the signature and passing a fractional percentile gets a silently wrong answer. Either narrow the parameter to `int`, or use `round(pct)` instead of `int(pct)`. Low priority given current call sites, but worth tightening before this function is called from a new context.

---

Everything else is clean. The scoring math, drift denominator handling, log capture attach/detach, module-state restoration, probe file parser, and test coverage across all six classes are all correct. The security posture is fine: no shell execution, no injection surface, PII stays gitignored.

dcellison:
## Review by Kai

The file isn't on this machine; working from the diff directly.

---

## PR #366 Review (Round 4)

The one open suggestion from review 3 - narrowing `_percentile`'s `pct` parameter from `float` to `int` - has been applied. The parameter is now typed `int`, `idx = pct - 1` replaces `int(pct) - 1`, and the docstring explains why. That closes all prior findings.

---

### New finding

**Suggestion - sweep output JSON omits `legacy_weight` from per-config `config` block** (`retrieval.py`, `_run_cli`, sweep output branch)

The single-config baseline records `legacy_weight` in the `config` envelope via `_PRODUCTION_LEGACY_WEIGHT`. The sweep output's per-row config block does not:

```python
# sweep mode (line ~1090)
"config": {
    "floor": cfg.floor,
    "extracted_weight": cfg.extracted_weight,
    "overfetch": cfg.overfetch,
    # legacy_weight absent
},

# single-config baseline (_build_baseline_json)
"config": {
    "floor": ..., "extracted_weight": ...,
    "legacy_weight": _PRODUCTION_LEGACY_WEIGHT,   # present
    "overfetch": ...,
},
```

The two output shapes can't be parsed uniformly by a downstream script. Since `legacy_weight` is fixed during a sweep, adding `"legacy_weight": _PRODUCTION_LEGACY_WEIGHT` to the sweep config block costs nothing and makes the schemas consistent.

---

### Clean

- `_percentile` narrowed to `int` correctly; `idx = pct - 1` with `max/min` clamp is sound.
- `_flatten_score_block` unifies both `metrics` and `by_tag` to flat keys; example fixture matches.
- `detect_drift` hoisted before the sweep loop; `_score_against_store` called per grid point.
- `recall_at_k` present in the top-level `metrics` block of `_build_baseline_json`.
- `_render_sweep_table` uses `.get(k, 0.0)` throughout.
- Sweep restoration `try/finally` and the test (class 3) are correct.
- `id` field threaded from `MemoryResult.id` through the recall payload; `test_memory.py` assertion updated.
- Security posture unchanged: no shell execution, no injection surface, PII stays gitignored.

One minor suggestion; nothing blocking.

dcellison:
## Review by Kai

The file isn't on this machine - working from the diff directly, which is sufficient.

---

## PR #366 Review (Round 5)

The Round 4 suggestion - adding `legacy_weight` to the sweep per-row config block - is addressed in this diff. The sweep JSON output now includes `"legacy_weight": _PRODUCTION_LEGACY_WEIGHT` in every row's config block, matching the single-config baseline shape. That closes all prior findings.

---

### New finding

**Suggestion - sweep JSON output lacks a top-level `drift_count`** (`retrieval.py`, `_run_cli`, sweep branch)

The single-config baseline (`_build_baseline_json`) emits `"drift_count"` at the top level alongside `"probe_count"`. The sweep JSON output has `"probe_count": len(probes)` but no `"drift_count"`. Since drift is hoisted before the loop (good), the count is available: it's `len(probes) - len(scored)` from the `detect_drift` return. Every per-row `metrics` dict carries `n_drift` via `_metrics_to_dict`, but all 72 rows will have the same value - it belongs once at the envelope level.

A downstream script comparing a single-config baseline against a sweep output needs different paths for the same piece of information (`.drift_count` vs `.sweep[0].metrics.n_drift`). Adding `"drift_count": len(drifted)` to the sweep envelope (where `drifted` is already in scope from the `detect_drift` call) costs one line and makes the two output types parse uniformly.

---

### Clean

- `legacy_weight` now present in the sweep per-row config block - schema is consistent with single-config baseline.
- `_flatten_score_block` correctly unifies both `metrics` and `by_tag` to flat keys; `_build_baseline_json` and `_metrics_to_dict` use it and the nested form respectively, with intent documented in the docstring.
- `_percentile` narrowed to `int`, `idx = pct - 1`, clamp is sound.
- `detect_drift` hoisted before the sweep loop; `_score_against_store` per grid point.
- `run_sweep` `try/finally` and the restoration test (class 3) are correct.
- `id` field threaded from `MemoryResult.id` through the recall log payload; `test_memory.py` assertion updated.
- Exit codes 0/1/2 are distinct and consistently used.
- Security posture unchanged: no shell execution, no injection surface, PII gitignored.

One minor suggestion; nothing blocking.

dcellison:
## Review by Kai

## PR #366 Review (Round 6)

The round 5 suggestion - adding `drift_count` to the sweep envelope - is addressed. The implementation uses `next((m.n_drift for _, m in sweep_results), 0)` which correctly retrieves the pre-hoisted drift count from the first row's Metrics (all rows carry the same `n_drift` since drift is computed once before the loop).

---

### New finding

**Suggestion - `run_sweep` docstring describes a per-iteration `try/finally` that does not exist** (`retrieval.py`, `run_sweep` docstring)

The docstring reads:

> "Restoration in `try/finally` per iteration defends against a probe raising mid-sweep; the OUTER try/finally restores once at the very end..."

This implies two levels of try/finally - one per grid iteration and one enclosing the loop. The actual code has only one outer `try/finally` wrapping the entire loop. The behavior is correct (any exception from any iteration propagates to the outer finally and state is restored), but the docstring describes a different design. The phrase "per iteration" and "OUTER try/finally" are both stale. Removing the "per iteration" sentence and the "OUTER" qualifier would make the docstring match the code.

---

### Clean

- `drift_count` now present at the sweep envelope level alongside `probe_count`; matches the single-config baseline shape.
- All six issues from rounds 1-5 remain fixed.
- `legacy_weight` in sweep per-row config block is consistent with single-config baseline.
- `_flatten_score_block` unifies baseline and `by_tag` shapes; example fixture matches.
- `_percentile` narrowed to `int`; clamp is sound.
- `detect_drift` hoisted before the sweep loop; `_score_against_store` called per grid point.
- Sweep `try/finally` restoration and test class 3 correctly validate state after a mid-sweep exception.
- `id` field threaded from `MemoryResult.id` through the recall payload; `test_memory.py` assertion updated.
- Security posture unchanged: no shell execution, no injection surface, PII gitignored.

One minor suggestion; nothing blocking.

dcellison:
## Review by Kai

This PR is in good shape after six rounds. Two minor new findings.

---

**Suggestion - Sweep `--output` comment promises a frontier sub-array that does not exist** (`retrieval.py`, `_run_cli`, sweep branch, ~line 1100)

The comment above the payload construction reads:

```python
# Sweep output mode: write a JSON file containing every
# config + metrics tuple plus the frontier subset, so a
# downstream consumer can re-render either view without
# re-running the sweep.
```

The actual payload has no `"frontier"` key - only `"sweep"` (all configs). A script that reads the comment and expects `payload["frontier"]` gets a `KeyError`. Either add it:

```python
"frontier": [
    {"config": dataclasses.asdict(cfg), "metrics": _metrics_to_dict(m, include_by_tag=False)}
    for cfg, m in pareto_frontier(sweep_results)
],
```

or drop "plus the frontier subset" from the comment. The frontier is re-derivable from the sweep rows so omitting it is defensible - the comment just needs to match the code.

---

**Suggestion - `--output` help text is only accurate in non-sweep mode** (`retrieval.py`, `_build_parser`)

```python
parser.add_argument(
    "--output",
    type=Path,
    help="Write a baseline JSON file matching the *.example.json schema.",
)
```

In sweep mode, `--output` writes a sweep-specific envelope (`version`, `probe_count`, `drift_count`, `sweep: [...]`) that does not match `retrieval-baseline.example.json`. Noting "single-config only" or splitting the description by mode would prevent confusion when an operator passes both `--sweep` and `--output` and then tries to validate the output against the example schema.

---

**Clean** - everything from rounds 1-6 is confirmed addressed: `run_sweep` docstring no longer mentions "per iteration try/finally" or "OUTER"; `drift_count` is at the sweep envelope; `legacy_weight` is in per-row config blocks; `_percentile` is narrowed to `int`; `_flatten_score_block` unifies both output shapes; `_render_sweep_table` uses `.get(k, 0.0)` throughout. The `id` field threads correctly through `memory.py` and the test assertion in `test_memory.py` is updated. No security or correctness issues.

dcellison:
## Review by Kai

## PR #366 Review (Round 8)

Both round 7 findings are addressed in this diff.

**Comment on sweep `--output`** (`_run_cli`, sweep branch): The comment no longer mentions "plus the frontier subset" - it now explains why the frontier is *not* serialized and how a consumer re-derives it. Matches the actual payload.

**`--output` help text** (`_build_parser`): Now explicitly documents both behaviors - single-config baseline vs. sweep envelope - with the explicit callout that the sweep shape does NOT match the baseline example schema. Accurate.

---

I looked hard for new issues and found nothing worth raising:

- `_flatten_score_block` receives integer-keyed `precision_at_k`/`recall_at_k` dicts from `score_probes`, and correctly uses `.get(1, 0.0)` with integer keys throughout. Consistent with the example fixture.
- `drift_count` in the sweep envelope via `next((m.n_drift for _, m in sweep_results), 0)` is correct - all rows carry the same `n_drift` since drift is hoisted pre-loop, and the fallback to 0 on an empty grid is sound.
- `--extracted-weight` -> `args.extracted_weight` argparse conversion is standard and correct.
- `_percentile` with `pct: int`, `idx = pct - 1`, and the clamp is sound.
- The `try/finally` sweep restoration, `_attach_capture`/`_detach_capture` symmetry, and the six test classes all look correct.

The PR is clean. No new findings.

PR #368, opened 2026-04-22
feat(memory): extractor-side consolidation via intent (#367)

### LINKED ISSUE #367: Extractor pipeline accumulates partial-overlap facts; no update logic
## Problem

The extractor (Haiku + `add_structured(infer=False)`) has no update or consolidation logic. Dedup is a top-1 semantic search at write time with a 0.9 similarity threshold, which catches near-verbatim repetition but misses paraphrase-level redundancy.

Multiple facts that overlap in topic but assert different (potentially contradictory) information can accumulate in the store.

## Evidence

Smoke run against an active production memory store today found three separate facts covering "where Kai data lives":

- `Kai's DATA_DIR is /var/lib/kai/.`
- `Kai's code is installed in /opt/kai/, data is stored in /var/lib/kai/, secrets and users.yaml are in /etc/kai/...`
- `Kai deliberately stores data outside its code installation tree to enable code updates without losing user data, history, or memory.`

Pairwise cosine similarity: 0.55, 0.61, 0.67. All well below the 0.9 threshold. The dedup mechanism is doing what it was designed to do; it wasn't designed to catch this case.

## Risk

Today's three siblings happen to agree. The general failure mode: a value changes (a path, a tunable, a config), a new conversation produces a fact reflecting the new value, but the old fact stays in the store because the paraphrased pair doesn't hit the threshold. Retrieval then returns both stale and current facts at high confidence, and the downstream LLM has no signal about which is current.

This is asymmetric risk: the longer the store accumulates, the more likely contradictions become.

## Architectural constraint

Mem0 has built-in update detection via `infer=True`. That path was deliberately removed in #361 because it was destroying real user input. Re-enabling it is not on the table; the fix has to live on top of the current Track 2 pipeline.

## Proposed direction

Extractor-side consolidation: pass the top-N existing facts for the user into the Haiku extractor's context, and have it output one of three intents per proposed fact: `new`, `update_of <existing_id>`, or `skip_redundant <existing_id>`. Storage honors `update_of` by replacing the old fact with the new.

Open design questions for the spec:

- Top-N strategy (semantic top-N, recent-N, all)
- Atomic replace vs delete-then-add at the Mem0 layer
- One-time consolidation pass over existing facts vs forward-only fix

dcellison:
Closes #367.

## Summary

Adds a Haiku-extractor consolidation system so paraphrase-equivalent facts can update or replace stale entries instead of accumulating as near-duplicates that bypass the existing 0.9 cosine dedup gate.

The extractor now emits one of three intents per fact:
- `new` - store as a fresh memory (existing behavior, still gated by the top-1 semantic dedup)
- `update_of <existing_id>` - delete the named prior memory then store the new one (Mem0 has no atomic replace, so it's delete-then-add)
- `skip_redundant <existing_id>` - drop the new fact entirely; the prior memory already covers it

Top-k semantic candidates from prior memories (default 8) are rendered into an `EXISTING FACTS` block in the prompt. The validator enforces that `update_of` / `skip_redundant` ids come from that candidate set; **hallucinated ids cause the fact to be dropped entirely** (with a structured log line carrying `intent="hallucinated_id"` and `original_intent` set to the rejected intent), so the rate is observable. Dropping rather than falling back to `new` is deliberate: storing an `update_of` payload whose anchor cannot be verified would risk creating orphaned duplicates.

The candidate-fetch step uses Mem0's vector search keyed on the assistant text. The user-side semaphore that already serializes extraction calls also serializes this new fetch, so concurrency is unchanged.

## Behavior changes

- Bumps `_EXTRACTION_PROMPT_VERSION` from `"1"` to `"2"` so any tooling that gates re-extraction on prompt-version sees a clean break.
- Adds two structured log sentinels:
  - `memory.consolidate.candidates` - per-extraction, lists candidate ids and similarities
  - `memory.consolidate.intent` - per-fact, captures the intent the extractor chose plus any validator overrides (`hallucinated_id` for rule-3 rejections, etc.)
- Storage-layer outcomes for the `new` intent are split into `dropped_duplicate` (healthy dedup) and `dropped_backend` (storage disabled or sick Mem0) so dashboard alerts on backend failure don't fire on benign deduplication.
- The `memory.extract` summary line now also reports `replaced=N skipped=N` counts.

## Kill switch

`MEMORY_CONSOLIDATION_CANDIDATES_N=0` skips the candidate-fetch step entirely; the extractor never sees an `EXISTING FACTS` block, the validator runs in the same path it did before, and behavior is byte-identical to pre-consolidation extraction. This is the documented rollback path; no redeploy needed.

Default value is `8`. Wizard prompts for it inside the extraction-enabled branch; non-default values are persisted to `install.conf` and re-emitted on round-trip.

## Storage layer

`_store_facts` now returns `(stored, replaced, skipped)` and branches on intent:
- `new`: existing path, `_is_duplicate` check at threshold 0.9 still applies
- `update_of`: delete the prior memory by id, then `add_structured(infer=False)` the new one
- `skip_redundant`: log + return without touching storage

The delete-then-add pattern is the only option Mem0 v2.0.0 exposes - there's no atomic replace primitive. Documented in the storage helper.

## Test plan

- [x] `pytest tests/test_memory_extraction.py` - 119 tests pass (64 existing migrated + 55 new across 11 test classes)
- [x] `pytest tests/test_config.py -k consolidation` - 5 new env-var tests pass
- [x] `pytest tests/test_install.py -k memory` - 22 wizard-flow tests pass (memory_block fixtures updated for new prompt position)
- [x] `pytest` - full suite, 2120 passed
- [x] `make check` - ruff check + format clean
- [ ] Manual smoke after merge: confirm `memory.consolidate.candidates` and `memory.consolidate.intent` log lines appear during a real extraction; confirm `MEMORY_CONSOLIDATION_CANDIDATES_N=0` produces a `candidates=0` log and no `EXISTING FACTS` block in the rendered payload.

## Notes for reviewer

- Forward-only: existing prompt-v1 memories stay; the version bump means new extractions are tagged `prompt_version=2`. No one-time migration pass is included.
- `_capped_assistant` lifts the assistant-text cap out of `_build_extraction_payload` so the candidate-fetch query and the Haiku prompt see byte-identical input. The cap behavior itself didn't change.
- The validator's batch-uniqueness rule (one `update_of` per existing_id within a single extraction) is enforced before storage runs - prevents Mem0 from getting two delete-then-add ops against the same id in one call.
- Validator rule 4 blocks both `skip_redundant` AND `update_of` on `confirmed_action` facts; defense-in-depth against the model disobeying the prompt instruction "always new for confirmations" (an `update_of` confirmation with a valid quote would otherwise pass all five rules and silently replace the prior confirmation).
- `_render_candidate_line` runs `_strip_role_labels` on `cand.text` so a fact in the store whose content contains injected USER:/ASSISTANT: markers cannot become a second-order injection vector through the EXISTING FACTS block.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -169,6 +169,13 @@ MEMORY_ENABLED=false
 #MEMORY_EXTRACTION_BUDGET_USD=0.01
 # Subprocess timeout (seconds). Haiku typically finishes in 2-4s.
 #MEMORY_EXTRACTION_TIMEOUT_S=10
+# Number of paraphrase-equivalent prior facts shown to the extractor
+# as candidates for consolidation (intent: update_of/skip_redundant).
+# Top-k semantic search at the assistant text. Set to 0 to disable
+# the candidate-fetch step entirely (rolls back to pre-consolidation
+# behavior - extractor only emits intent="new"). The 0 path is the
+# documented kill switch.
+#MEMORY_CONSOLIDATION_CANDIDATES_N=8
 
 # ── File retention (optional) ────────────────────────────────
 # Delete uploaded files (photos, documents, voice notes) older than this many
--- a/src/kai/config.py
+++ b/src/kai/config.py
@@ -458,6 +458,22 @@ class Config:
     # an executor thread on a hung subprocess.
     memory_extraction_timeout_s: int = 10
 
+    # Number of existing facts surfaced to the extractor per call as
+    # consolidation candidates (intent: update_of / skip_redundant).
+    # Selected by semantic similarity to the assistant payload, capped
+    # at this value. Set to 0 to disable consolidation entirely: the
+    # candidate fetch is skipped, the EXISTING FACTS data block is
+    # omitted from the Haiku payload, and the extractor falls back to
+    # always emitting intent="new" (anchored by the CONSOLIDATION
+    # prompt section, which is retained even when the data block is
+    # empty). Storage then uses the existing _is_duplicate semantics.
+    # Tradeoff: each candidate adds ~50-100 chars to the Haiku payload
+    # and contributes to per-call cache-creation tokens. Default 8 is
+    # 2x the per-call fact cap (5), which gives the model enough to
+    # find 1-2 update targets per proposed fact without doubling the
+    # per-call cost.
+    memory_consolidation_candidates_n: int = 8
+
     # Minimum Mem0 similarity score for a memory to be returned by
     # search-driven paths: both `format_context` (context injection
     # at session start) and the `/memory search` UI surface in
@@ -1415,6 +1431,16 @@ def load_config() -> Config:
             raise SystemExit("MEMORY_EXTRACTION_TIMEOUT_S must be a positive integer")
     except ValueError:
         raise SystemExit("MEMORY_EXTRACTION_TIMEOUT_S must be an integer") from None
+    # Consolidation candidate count. Zero is a valid kill-switch value
+    # (consolidation disabled, extractor falls back to all-`new`); only
+    # negatives are rejected. Same try/except pattern as the other
+    # memory_* numeric vars.
+    try:
+        memory_consolidation_candidates_n = int(os.environ.get("MEMORY_CONSOLIDATION_CANDIDATES_N", "8"))
+        if memory_consolidation_candidates_n < 0:
+            raise SystemExit("MEMORY_CONSOLIDATION_CANDIDATES_N must be a non-negative integer")
+    except ValueError:
+        raise SystemExit("MEMORY_CONSOLIDATION_CANDIDATES_N must be an integer") from None
 
     # Search relevance floor. Float in [0.0, 1.0]; default 0.3 matches
     # Mem0's built-in default and the prior hard-coded constant. Same
@@ -1588,5 +1614,6 @@ def load_config() -> Config:
         memory_extraction_model=memory_extraction_model,
         memory_extraction_budget_usd=memory_extraction_budget_usd,
         memory_extraction_timeout_s=memory_extraction_timeout_s,
+        memory_consolidation_candidates_n=memory_consolidation_candidates_n,
         memory_search_floor=memory_search_floor,
     )
--- a/src/kai/install.py
+++ b/src/kai/install.py
@@ -214,6 +214,19 @@ def _validate_positive_int(value: str) -> bool:
         return False
 
 
+def _validate_non_negative_int(value: str) -> bool:
+    """Check that a string is a non-negative integer (zero allowed).
+
+    Used for fields where 0 is a meaningful kill-switch value, not an
+    invalid input. Distinct from `_validate_positive_int` so the
+    operator-facing error message can reflect the actual constraint.
+    """
+    try:
+        return int(value) >= 0
+    except ValueError:
+        return False
+
+
 def _validate_chat_id(value: str) -> bool:
     """Check that a string is a valid Telegram chat ID (any non-zero integer)."""
     try:
@@ -714,6 +727,7 @@ def _cmd_config() -> None:
     memory_extraction_enabled = False
     memory_extraction_budget_usd = "0.01"
     memory_extraction_timeout_s = "10"
+    memory_consolidation_candidates_n = "8"
     memory_token_budget = "2000"
     memory_search_limit = "10"
     if memory_enabled:
@@ -750,6 +764,21 @@ def _cmd_config() -> None:
                     if _validate_positive_int(memory_extraction_timeout_s):
                         break
                     print("  Must be a positive integer.")
+                # Consolidation candidate count. Zero is a valid
+                # kill-switch value (extractor falls back to all-`new`
+                # output and the EXISTING FACTS data block is omitted),
+                # so use the non-negative validator instead of the
+                # positive-only one. Sits inside the extraction guard
+                # because consolidation is only meaningful when the
+                # Haiku extractor runs.
+                while True:
+                    memory_consolidation_candidates_n = _prompt(
+                        "Consolidation candidate count (0 disables consolidation)",
+                        existing_env.get("MEMORY_CONSOLIDATION_CANDIDATES_N", "8"),
+                    )
+                    if _validate_non_negative_int(memory_consolidation_candidates_n):
+                        break
+                    print("  Must be a non-negative integer (0 to disable consolidation).")
         while True:
             memory_token_budget = _prompt(
                 "Memory context token budget per turn",
@@ -885,6 +914,13 @@ def _cmd_config() -> None:
             # the dataclass default.
             if int(memory_extraction_timeout_s) != 10:
                 env["MEMORY_EXTRACTION_TIMEOUT_S"] = memory_extraction_timeout_s
+            # Consolidation candidate count. Same gate as the other
+            # extraction tunables because the field is only consulted
+            # by the Haiku extraction path; numeric compare so "08" or
+            # "8 " do not produce a spurious entry equal to the
+            # dataclass default.
+            if int(memory_consolidation_candidates_n) != 8:
+                env["MEMORY_CONSOLIDATION_CANDIDATES_N"] = memory_consolidation_candidates_n
         if int(memory_token_budget) != 2000:
             env["MEMORY_TOKEN_BUDGET"] = memory_token_budget
         # Search limit applies to retrieval (read path), not extraction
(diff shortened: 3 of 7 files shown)


dcellison:
## Review by Kai

Good - `_is_duplicate` calls `memory.is_enabled()` first (line 439), so the test's monkeypatch strategy is sound: patching `is_enabled` to `pytest.fail` correctly guards against `_is_duplicate` being called in the `update_of` branch.

I now have enough context for a complete review.

---

## Review: feat(memory): extractor-side consolidation via intent (#368)

The PR is well-structured and the test coverage is thorough. No critical bugs. A few items worth fixing before merge:

---

### Warning

**`_store_facts`: `dropped` outcome is overloaded (`memory_extraction.py`)**

Two semantically distinct `new`-intent outcomes both emit `"outcome":"dropped"`:
1. `_is_duplicate` returns True - dedup working correctly, expected at volume
2. `add_structured` returns None - silent backend failure, operator should alert

A dashboard alert on `"outcome":"dropped"` cannot distinguish healthy dedup from a sick Mem0 backend. Consider `"dropped_duplicate"` vs `"dropped_backend"` to make the signal actionable without adding fields.

---

**Defensive gap: `memory.search` returning `None` instead of raising (`memory_extraction.py`, `extract_and_store`)**

```python
try:
    candidates = await loop.run_in_executor(
        None, lambda: memory.search(assistant_capped, user_id=user_id, limit=n_candidates),
    )
except Exception:
    candidates = []
```

The `except Exception` catches a search call that *raises*. If `memory.search` returns `None` silently (e.g. a future mock stub or edge in Mem0), execution falls through to `{c.id for c in candidates}` and raises `TypeError` - which lands in the outer `except Exception: return 0` block, silently dropping all extraction for that turn. A one-liner guard (`candidates = candidates or []`) after the executor call removes the ambiguity and matches how `_is_duplicate` defends against search failures.

---

### Suggestion

**`update_of` bypass of `_is_duplicate` can create near-duplicate pairs**

The bypass is intentional and documented. But if the replaced fact's new content is a near-verbatim match of an existing fact that was outside the top-8 candidate window, the replace lands without the 0.9-threshold dedup check, leaving two near-identical facts in the store. The case is rare at N=8 (2x the per-call fact cap), but noting it here so the operational runbook for consolidation-related duplicate spikes covers this path. A comment at the skip site in `_store_facts` pointing back to the candidate-window design would be enough.

**`_validate_non_negative_int` docstring is 3 lines (`install.py`, line ~216)**

Existing helpers in the same file (`_validate_positive_int`, `_validate_chat_id`) use the same multi-line style, so this is internally consistent - but verify against the project's commenting style guide if it caps docstrings at one line.

---

### Confirmed clean

- New-env-var checklist fully followed (dataclass field, `load_config` parsing, wizard prompt, `.env.example`, `_CONFIG_ENV_VARS` in tests).
- `_validate_facts` rule ordering (consolidation rules before confirmation-quote rules) is correct; the first applicable rule fires.
- Rule-5 pre-pass correctly counts only non-`new` intents and evaluates `citation_counts` after rule 3, so hallucinated-id pairs don't poison the uniqueness count for real-candidate facts.
- `_is_duplicate`'s early `memory.is_enabled()` call makes `test_update_of_happy_path`'s monkeypatch-to-fail strategy correct.
- Kill-switch (`n=0`) path: candidate fetch skipped, no `EXISTING FACTS` block in payload, validator's empty `candidate_ids` set drops any non-`new` intents via rule 3 - all verified by the test suite.

dcellison:
## Review by Kai

## Review: feat(memory): extractor-side consolidation via intent (#368)

Both prior warnings are properly addressed: `dropped_duplicate` / `dropped_backend` are now distinct, and `candidates = candidates or []` guards against a `None` return from `memory.search`. No critical bugs found. Two items worth noting:

---

### Suggestion

**Validator/prompt mismatch: `update_of` on `confirmed_action` facts is not blocked (`memory_extraction.py`, `_validate_facts` rule 4)**

The system prompt says "A confirmed_action fact is always 'new'" (meaning Haiku should never emit `update_of` for a confirmation), but rule 4 only rejects `skip_redundant + confirmed_action`. If Haiku disobeys the prompt and emits `update_of` for a `confirmed_action` fact with a valid quote, it passes all five rules, deletes the prior confirmation, and stores the new one in its place.

The confirmation-quote rules do still run (so an `update_of` confirmation without a valid quote gets caught), but the consolidation intent itself is not blocked. Given the explicit prompt instruction, adding `intent in ("update_of", "skip_redundant")` to rule 4 would close the gap:

```python
if intent in ("update_of", "skip_redundant") and "confirmed_action" in tags:
```

**Risk**: Low - Haiku is unlikely to disobey the explicit instruction, and even if it does, the outcome (replacing an old confirmation with a new one) is often harmless.

---

### Note

**`cand.text` in the EXISTING FACTS block is not role-label-stripped (`memory_extraction.py`, `_render_candidate_line`)**

`_strip_role_labels` is applied to `user_text` and `assistant_text` but not to candidate content. A stored fact whose text contains injected role-label tokens (e.g., via an adversarially crafted earlier exchange that got extracted into the store) would be rendered into the `EXISTING FACTS` block without sanitization - a second-order prompt injection vector through the store. The attack chain is two steps (compromise extraction, then exploit retrieval), so the risk is low in practice, but it's a structural gap worth tracking as the store grows.

---

### Confirmed clean

- Both prior warnings resolved (`dropped_duplicate` / `dropped_backend` split; `candidates or []` guard).
- New env-var checklist fully followed in all six steps.
- Rule ordering in `_validate_facts` is correct: consolidation rules fire before confirmation-quote rules, rule 3 is the sole structured-log emitter.
- Pre-pass citation-count counts only `update_of` / `skip_redundant` intents, so a stray `existing_id` on a `new` fact does not poison the uniqueness count.
- `_capped_assistant` lifted to function scope correctly; payload builder no longer truncates assistant text, and `test_long_assistant_text_passthrough_when_pre_capped` pins that contract.
- Kill-switch path (`n=0`): candidate fetch skipped, empty `candidate_ids` causes rule 3 to drop any non-`new` intents, `candidates.log` still emits with `n_candidates=0`. Structurally sound.
- `test_update_of_happy_path` monkeypatch-to-fail on `memory.is_enabled` correctly guards that `_is_duplicate` is bypassed in the `update_of` branch.
- `_store_facts` `(stored, replaced, skipped)` semantics match the docstring: `add_failed_after_delete` increments neither counter, `delete_failed_added_anyway` increments both `stored` and `replaced`.

dcellison:
## Review by Kai

## Review: feat(memory): extractor-side consolidation via intent (#368)

This is a third-pass review. Both prior warnings are resolved. One documentation discrepancy noted; no bugs or security issues found.

---

### Note

**PR description misstates hallucinated-id behavior (`memory_extraction.py`, rule 3 in `_validate_facts`)**

The description says:
> "Hallucinated ids fall back to `new` with a structured log line so the rate is observable."

The code does the opposite - it drops the fact entirely (`continue` after `_emit_intent_log`), confirmed by `test_rule3_hallucinated_update_of_emits_log` asserting `== []`. "Fall back to `new`" would mean storing the content as a fresh fact (running it through `_is_duplicate` and `add_structured`); that is not what happens.

The drop behavior is defensible - storing an `update_of` whose target cannot be anchored could create orphaned duplicates - but the description would mislead anyone writing an ops runbook around the `hallucinated_id` outcome. Worth a one-line fix in the description or the docstring on rule 3 before the PR closes.

---

### Confirmed clean

- Both prior warnings fully addressed: `dropped_duplicate` / `dropped_backend` split in place; `candidates or []` guard present; rule 4 now blocks `update_of` on `confirmed_action` as well as `skip_redundant`.
- `_FACT_SCHEMA` and `_EXTRACTION_SYSTEM_PROMPT` are imported in the test file and presumably used in the 51 new test classes not visible in the truncated diff.
- `_capped_assistant` correctly lifted out of `_build_extraction_payload`; `test_long_assistant_text_passthrough_when_pre_capped` pins the contract.
- Kill-switch path (`n=0`): candidate fetch skipped, empty `candidate_ids` causes rule 3 to drop any non-`new` intents, `candidates.log` still emits with `n_candidates=0`.
- New env-var checklist: all six steps followed (dataclass field, `load_config` parsing, wizard prompt inside extraction guard, `.env.example`, `_CONFIG_ENV_VARS` in tests, `env.pop` on the disabled path).
- `_render_candidate_line` runs `_strip_role_labels` on `cand.text` - second-order injection path through the store is structurally defended.
- `_store_facts` `(stored, replaced, skipped)` semantics match the docstring: `add_failed_after_delete` increments neither, `delete_failed_added_anyway` increments both `stored` and `replaced`.
- Rule ordering is correct: consolidation rules 1-5 fire before confirmation-quote rules; first applicable rule wins; rule 3 is the sole structured-log emitter.

PR #370, opened 2026-04-23
feat(eval): Layer 2 behavioral A/B harness for memory injection

### LINKED ISSUE #369: Layer 2: end-to-end behavioral A/B harness
## Layer 2: end-to-end behavioral A/B harness

Sub-issue of #362. Builds on the retrieval-only evaluation completed in #365 (results: https://github.com/dcellison/kai/issues/362#issuecomment-4303851435).

### What this layer answers

Layer 1 confirmed that retrieval surfaces stored facts at MRR 0.806 on a 26-probe set. What it does NOT answer: when memory does inject relevant facts into the prompt, does the generated response actually use them better than a memory-off response would? And inversely, when memory injects irrelevant facts (false positives, paraphrase clusters), does it make responses worse?

This is the only layer that catches:
- Generator ignoring retrieved facts (retrieval works, generation squanders it)
- Generator hallucinating facts that contradict retrieved truth
- Memory injecting noise that distracts from a question the generator could have answered cleanly without it
- The "memory makes no measurable difference" null result

### Design

Run a probe set twice for each question:
- **Arm A:** prompt assembled with memory block injected (current production behavior).
- **Arm B:** prompt assembled with the memory block omitted.

Both arms call `claude --print` with identical model, temperature, and system prompt. Responses are paired, anonymized (random A/B order per probe), and scored by a separate LLM judge call against a rubric tied to the gold-fact text from the probe.

Judge returns: `A_wins` / `B_wins` / `tie` / `both_wrong`. Aggregate to win-rate, tie-rate, loss-rate.

### Probe schema extension

Reuse the Layer 1 probes file but add an optional `ground_truth_text` field per probe. The judge reads the gold fact text directly from the store via `expected_fact_id` for probes that omit `ground_truth_text`, so the existing 26 probes are usable without rewriting.

### Scope choices (worth pushing back on if wrong)

1. **Skip the bot wrapper.** The harness builds prompts directly via `format_context` and calls `claude --print`. No Telegram pipeline, no message router, no session logic. The hypothesis under test is about retrieve + inject + generate; the bot wrapper is well-tested elsewhere and adds latency, cost, and a confound. The eval is measuring memory's effect on generation, not the bot's effect on memory.

2. **Single-shot per arm.** No multi-sample variance reduction in v1. Generation is non-deterministic but at temperature 0 (or low) the pairwise comparison signal is strong enough to read directly from single samples on a 26-probe set. If win-rates land near 50/50, a v2 with N=3 samples per arm would be the natural next iteration.

3. **Haiku as judge.** Cheapest path. Risks: judge bias toward verbosity, falling for confident wrong answers. Mitigations: anonymized A/B pairing (judge does not know which arm produced which response), explicit rubric anchored on gold-fact text. If judge reliability looks weak in early runs, swap to Sonnet for the judge call (the generator stays whatever model the user normally runs).

4. **Out of scope:** longitudinal effects (Layer 3 covers that), multi-turn conversations (one probe = one user message + one response), and conversational style judgments (judge scores fact accuracy and relevance only).

### Cost estimate

Per run, on the existing 26 probes: 26 generations + 26 generations + 26 judge calls = 78 Claude calls. At Haiku rates roughly $1-2 per run; at Sonnet rates roughly $5-10. Cheap enough to run weekly during iteration; not cheap enough for CI.

### Deliverables

- Module `kai.eval.behavioral` with CLI: `python -m kai.eval.behavioral --probes <path> --user-id <id>`
- A/B prompt builder that toggles memory block in/out
- Anonymized judge call with structured output (`A_wins` / `B_wins` / `tie` / `both_wrong`)
- Aggregate output: win-rate, tie-rate, loss-rate, per-probe breakdown, plus a few qualitative examples (best memory help, worst memory hurt, clearest tie)
- Output JSON schema versioned for comparison across runs
- Tests: judge-call contract test (mocked), prompt-assembly diff test (memory-on vs memory-off prompts differ exactly in the memory block)

### Success criterion for the sub-issue

The harness runs end-to-end and produces a win/tie/loss number with a per-probe breakdown that an operator can read without invoking the implementer. The number itself is a measurement, not a target -- a 50/50 win-rate or a memory-loses result is a valid outcome and gets reported honestly per the epic's null-results-acceptable clause.

### Related

- #362 (parent epic)
- #363 (retrieval logging, prerequisite, merged)
- #365 (Layer 1 harness, merged)

dcellison:
## Summary

- New eval harness at `src/kai/eval/behavioral.py` (CLI: `python -m kai.eval.behavioral`)
- For each probe, runs the question through the model twice (memory-on and memory-off arms, randomized A/B order); LLM judge scores the two responses against the gold-fact text
- Reports win/loss/tie/both_wrong rates plus per-tag breakdown, qualitative examples, and per-probe rows in a versioned JSON envelope

## Why

Layer 1 (retrieval eval) measures whether the right facts are surfaced. It does not answer the operator-facing question: when retrieval works, does memory injection actually make the response better? When retrieval surfaces noise, does it make the response worse? This harness answers that.

## Design notes

- **A/B isolation.** Both arms go through the same `build_arm_prompt` helper, which calls the production `prepend_to_prompt` join; the only branch is the `format_context` call. A byte-level regression test asserts the two arms differ in EXACTLY one place (the memory block).
- **Subprocess sandboxing.** Both judge and generator run via `claude --print` with the same flag vector the extractor uses: explicit neutral `cwd` (no CLAUDE.md auto-discovery), env allow-list, `--tools ""`, `--no-session-persistence`, kill-on-timeout. Generator passes `--system-prompt ""` (literally empty, not omitted) so the CLI default cannot confound the measurement.
- **Anonymization.** The arm letter that carries memory is randomized per probe, seeded from `sha256(probe_set_hash + user_id)` for reproducibility within a probe-set + user pair. The judge never sees arm semantics.
- **Failure buckets.** `judge_error`, `generation_error`, and `drift` are deliberately separate so a noisy run is diagnosable.

## Test plan

- [x] `make check` passes (ruff check + format)
- [x] `make test` passes (2164 tests including 44 new)
- [x] New tests cover A/B isolation, judge prompt rendering, judge output parsing (valid + 7 malformed shapes), anonymization determinism, the 8-cell rollup truth table, subprocess timeout bucketing, drift exclusion from rate denominators, explicit subprocess `cwd`, and the literal empty system prompt
- [x] All subprocess calls in tests are mocked; no real `claude` binary or network calls
- [ ] Operator runs the harness against the live store on the existing 26-probe set and posts results to the parent epic

## Out of scope

- CI integration (per-run cost is too high for per-PR runs; this is operator-driven)
- Multi-sample variance reduction (defer if v1 win-rates land near 50%)
- Multi-turn conversations (one probe = one user message + one response)

fixes #369

### DIFF
--- a/src/kai/eval/behavioral.py
+++ b/src/kai/eval/behavioral.py
@@ -1,0 +1,1569 @@
+"""
+Layer 2 end-to-end behavioral A/B evaluation harness.
+
+Reachable as `python -m kai.eval.behavioral`.
+
+Layer 1 (`kai.eval.retrieval`) measures whether the retrieval step
+surfaces the right facts. It does NOT answer the question that matters
+to the operator: when retrieval works, does memory injection actually
+make the model's response BETTER than a memory-off response would be?
+And when retrieval surfaces irrelevant facts (false positives, paraphrase
+clusters), does it make the response WORSE?
+
+This module answers that. For each probe, it runs the question through
+the model TWICE: once with the memory block prepended (arm A or B,
+chosen randomly), once with the memory block omitted. The two responses
+are paired and scored by a separate LLM-judge call against the gold-fact
+text. The judge picks A_wins / B_wins / tie / both_wrong; the harness
+de-anonymizes that into memory_wins / memory_loses / tie / both_wrong
+and aggregates win-rate, loss-rate, tie-rate, both_wrong-rate.
+
+Design rationale (the parts that are not obvious from the code):
+
+- A/B isolation. The hypothesis "memory helps" only holds if the two
+  arms differ in EXACTLY one place: presence vs absence of the memory
+  block. Any other delta (whitespace, system prompt, history, workspace
+  CLAUDE.md, message ordering) is a confound. The harness builds both
+  arms through the SAME `build_arm_prompt` helper which calls the same
+  production `prepend_to_prompt` join, and the only branch in the code
+  is the `format_context` call itself. A regression test (test 1) asserts
+  byte-level equality against `prepend_to_prompt(question, memory_block)`
+  so a future refactor of the production helper cannot quietly diverge
+  the two arms.
+
+- Subprocess sandboxing. Both judge and generator run via `claude --print`
+  subprocesses, lifted from the extractor's pattern in
+  `kai.memory_extraction._run_extractor`. The flag vector, env allow-list,
+  neutral cwd (no CLAUDE.md auto-discovery), and kill-on-timeout shape
+  are all reused as-is. Generator deliberately passes `--system-prompt ""`
+  (literally empty, not omitted) so the CLI's default system prompt does
+  not confound the measurement. Some default context still flows from
+  the CLI no matter what (~340 input tokens at CLI 2.1.118), but it is
+  identical between arms within one run, so it cancels for A/B purposes.
+  Cross-run drift in that residual is detectable via the
+  `claude_cli_version` field captured in the output JSON.
+
+- Anonymization. The per-probe arm assignment (memory-on -> A vs B) is
+  randomized so the judge cannot infer arm semantics from position.
+  Random is seeded deterministically from `sha256(probe_set_hash + user_id)`
+  by default, so a rerun against the same store produces identical arm
+  assignments — that property matters when debugging "why did probe 17
+  swing from win to loss." Operators can override with `--seed`.
+
+- Drift bucket. Probes whose `expected_fact_id` does not resolve via
+  `memory.get_by_id` are bucketed as `drift` (fact deleted, ownership
+  mismatch, source not "extracted", typo at probe authoring) and
+  excluded from win/loss/tie denominators. Same discipline as Layer 1.
+  Reusing `kai.eval.retrieval.detect_drift` keeps the two layers in
+  agreement about which probes are scorable on a given snapshot.
+
+- Failure buckets. `judge_error` and `generation_error` are deliberately
+  separate so a noisy run is diagnosable: a 30% generation_error rate
+  means the gen subprocess is broken (rate-limited, timed out, model
+  unavailable); a 30% judge_error rate means the judge subprocess is
+  broken (bad budget cap, schema regression). Mixing them would hide
+  which side needs operator attention.
+
+PII posture: probe questions and gold-fact text remain in the gitignored
+probes file plus the operator-named output JSON. The committed test
+fixtures use synthetic data only.
+
+Cost: per run, on the existing 26-probe set, ~26 generations + 26
+generations + 26 judge calls = 78 Claude calls. At Sonnet generator +
+Haiku judge that lands around $5-10 per run; cheap enough for weekly
+operator-driven iteration, too expensive for per-PR CI.
+"""
+
+from __future__ import annotations
+
+import argparse
+import asyncio
+import hashlib
+import json
+import logging
+import os
+import random
+import statistics
+import subprocess
+import sys
+import time
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any
+
+# Import from the Layer 1 sibling module so the two layers stay aligned
+# on probe schema, drift detection, and probe-set-hash semantics. If
+# Layer 1's drift bucketing changes, Layer 2 follows automatically.
+from kai.eval.retrieval import (
+    Probe,
+    detect_drift,
+    load_probes,
+    probe_set_hash,
+)
+
+log = logging.getLogger(__name__)
+
+
+# ── Constants ───────────────────────────────────────────────────────
+
+
+# Schema version written into the output JSON. Bump when the output
+# shape changes in a way that downstream consumers (jq filters,
+# comparison scripts) would misinterpret. The probe_set_hash is the
+# orthogonal "is this comparable to that other run" check; the
+# version is the "can my parser still read this" check.
+_OUTPUT_SCHEMA_VERSION = 1
+
+
+# Default judge model. Haiku is the cheapest path that has held up
+# in practice for short structured-output classification tasks. If the
+# judge starts producing biased verdicts (verbosity preference, falling
+# for confident wrong answers), swap to Sonnet via `--judge-model`.
+_DEFAULT_JUDGE_MODEL = "claude-haiku-4-5-20251001"
+
+
+# Default generation model. The generator is the system-under-test:
+# the operator typically wants this to match whatever the bot runs in
+# production. Sonnet is the production default; override via `--gen-model`.
+_DEFAULT_GEN_MODEL = "sonnet"
+
+
+# Per-call cost ceilings. The judge call is small (~3-4k input tokens,
+# tens of output tokens), so 0.05 USD is generous. The generator can
+# emit longer responses for multi-fact questions; 0.10 USD covers a
+# 1-2k output-token reply at Sonnet rates.
+_DEFAULT_JUDGE_BUDGET_USD = 0.05
+_DEFAULT_GEN_BUDGET_USD = 0.10
+
+
+# Subprocess timeout ceilings. Match the extractor's 60s default; a
+# generator producing more text might run longer, so it gets 120s.
+# These are wall-clock kill caps, not soft hints — exceeding either
+# kills the subprocess and buckets the probe as judge_error or
+# generation_error respectively.
+_DEFAULT_JUDGE_TIMEOUT_S = 60
+_DEFAULT_GEN_TIMEOUT_S = 120
+
+
+# Default concurrency for the per-probe Semaphore. Each probe slot
+# runs both arms sequentially plus the judge call (3 subprocesses
+# total), so 4 in flight means up to 12 simultaneous claude processes.
+# Higher concurrency hits rate limits on Anthropic's API and burns
+# laptop fans without speeding up the wall-clock meaningfully.
+_DEFAULT_MAX_CONCURRENCY = 4
+
+
+# JSON Schema passed to `claude --print --json-schema` for the judge
+# call. The four-value `choice` enum maps directly to the harness's
+# outcome buckets (after de-anonymization). `additionalProperties=false`
+# closes the schema so the judge cannot smuggle extra fields the
+# harness would silently ignore. The judge's `reasoning` field is kept
+# permissive (just a string) — the prompt caps its length verbally but
+# the schema does not enforce, because schema-side max-length on string
+# fields is not portably supported across CLI versions.
+_JUDGE_SCHEMA: dict[str, Any] = {
+    "type": "object",
+    "properties": {
+        "choice": {
+            "type": "string",
+            "enum": ["A_wins", "B_wins", "tie", "both_wrong"],
+        },
+        "reasoning": {"type": "string"},
+    },
+    "required": ["choice", "reasoning"],
+    "additionalProperties": False,
+}
+
+
+# The judge prompt rubric, embedded as the system prompt. Per-probe
+# substitutions (question, ground_truth_text, response_a, response_b)
+# are interpolated into this template at call time and fed via stdin.
+# Keeping the rubric in the system slot rather than stdin lets the
+# Claude CLI's prompt-caching reuse it across the dozens of judge calls
+# in one run.
+#
+# CRITICAL: the rubric's "Do NOT score on" block is load-bearing. LLM
+# judges drift toward verbosity / confidence preference by default;
+# spelling out what NOT to weight pushes back. The narrowing sentence
+# on `both_wrong` ("Use 'both_wrong' only when NEITHER response uses
+# the GROUND-TRUTH FACT") prevents the judge from over-firing on
+# responses that use the fact imperfectly. Past internal review caught
+# both of these as biases worth fixing at prompt time, not in code.
+_JUDGE_SYSTEM_PROMPT = """You are an impartial evaluator comparing two assistant responses to the same user question. Your goal is to identify which response more accurately uses the GROUND-TRUTH FACT to answer the question.
+
+Score on these criteria, in order:
+1. Factual accuracy: which response is more consistent with the GROUND-TRUTH FACT? A response that contradicts the ground truth loses regardless of style.
+2. Relevance: which response actually addresses the user's question rather than tangentially related material?
+
+Do NOT score on:
+- Verbosity (longer is not better)
+- Confidence or hedging
+- Formatting, markdown, bullet points
+- Conversational tone
+
+Use "tie" when both responses are roughly equivalent in accuracy and relevance. Use "both_wrong" only when NEITHER response uses the GROUND-TRUTH FACT (a response that uses the fact but adds tangential noise is still better than one that ignores it).
+
+Output ONLY a single JSON object on one line, no prose:
+{"choice": "A_wins"|"B_wins"|"tie"|"both_wrong", "reasoning": "<one sentence, max 30 words>"}"""
+
+
+# Per-probe stdin template. The system slot above is the rubric (cached
+# across calls); this is the only thing that varies per probe. Keeping
+# them split means the judge sees identical rubric text for every probe,
+# which is the only way the prompt-caching path actually amortizes.
+_JUDGE_USER_TEMPLATE = """USER QUESTION:
+{question}
+
+GROUND-TRUTH FACT (what the operator has previously told the assistant):
+{ground_truth_text}
+
+RESPONSE A:
+{response_a}
+
+RESPONSE B:
+{response_b}"""
+
+
+# Bucket names used in the output JSON's `outcomes` block. Defined as
+# constants so the test that asserts the sum-check (`sum(outcomes) +
+# drift == probe_count`) and the aggregation logic below stay aligned.
+_OUTCOME_BUCKETS: tuple[str, ...] = (
+    "memory_wins",
+    "memory_loses",
+    "tie",
+    "both_wrong",
+    "judge_error",
+    "generation_error",
+)
+
+
+# Mapping from raw judge `choice` string to per-probe `memory_outcome`.
+# The mapping depends on which arm letter was assigned to memory-on
+# for that probe; see `_rollup_outcome` for the full truth table.
+# Defined inline rather than as a module-level dict because it depends
+# on the per-probe arm letter; no point pre-computing.
+
+
+# ── Data classes ────────────────────────────────────────────────────
+
+
+@dataclass(frozen=True)
+class BehavioralConfig:
+    """Per-run configuration, materialized from CLI args.
+
+    Frozen so a probe slot cannot mutate it under another slot.
+    Carries everything needed to drive both subprocesses plus the
+    seed/concurrency knobs that govern run-level reproducibility.
+    """
+
+    judge_model: str
+    judge_budget_usd: float
+    judge_timeout_s: int
+    gen_model: str
+    gen_budget_usd: float
+    gen_timeout_s: int
+    seed: str
+    max_concurrency: int
+
+
+@dataclass
+class ProbeOutcome:
+    """Result of one probe's full A/B + judge cycle.
+
+    `memory_arm_letter` is the post-hoc record of which arm carried
+    the memory block — needed at aggregation time to translate the
+    judge's A/B verdict back into memory-on/memory-off terms. The raw
+    `judge_choice` is preserved alongside the rolled-up `memory_outcome`
+    so a future debugger can see what the judge actually said before
+    the de-anonymization step.
+
+    `responses` and `latency_ms`/`cost_usd` are dicts keyed by arm
+    letter ("A", "B") so the per-probe row in the output JSON can
+    surface both arms symmetrically.
+    """
+
+    probe: Probe
+    tags: tuple[str, ...]
+    memory_arm_letter: str  # "A" or "B"; the arm that received the memory block
+    responses: dict[str, str]  # {"A": "...", "B": "..."}
+    judge_choice: str  # one of A_wins|B_wins|tie|both_wrong|judge_error|generation_error|drift
+    judge_reasoning: str
+    memory_outcome: str  # rolled-up bucket name (one of _OUTCOME_BUCKETS or "drift")
+    latency_ms: dict[str, float]  # {"A": ..., "B": ..., "judge": ...}
+    cost_usd: dict[str, float | None]  # {"A": None, "B": None, "judge": ...}
+
+
+# ── Probe loading ──────────────────────────────────────────────────
+
+
+def load_behavioral_probes(path: Path) -> tuple[list[Probe], dict[str, str]]:
+    """Load probes plus optional per-probe ground_truth_text overrides.
+
+    Reuses Layer 1's `load_probes` for the base schema (question,
+    expected_fact_id, source_turn_ts, notes) and then re-parses the
+    same file to pick up the optional `ground_truth_text` field,
+    returned as a dict keyed by `expected_fact_id`. Probes without
+    the field will have their gold text resolved at scoring time via
+    `memory.get_by_id`.
+
+    Two-pass design: Layer 1's loader is the source of truth for the
+    base probe schema (validation, comment skipping, line numbering),
+    and re-parsing here is cheap (probes files are small JSONL). The
+    alternative — extending `load_probes` with an optional return —
+    would couple Layer 1 to a Layer 2-only field. Independent layers
+    over a shared loader is the cleaner separation.
+    """
+    probes = load_probes(path)
+    overrides: dict[str, str] = {}
+    raw = path.read_text(encoding="utf-8").splitlines()
+    for line in raw:
+        stripped = line.lstrip()
+        # Mirror Layer 1's comment/blank skip exactly so we touch the
+        # same set of lines. A divergence here would silently mean
+        # Layer 1 sees probes Layer 2 doesn't (or vice versa).
+        if not stripped or stripped.startswith("#"):
+            continue
+        try:
+            obj = json.loads(stripped)
+        except json.JSONDecodeError:
+            # Already validated by load_probes above; if a line passed
+            # there, json.loads here cannot fail. The except guards the
+            # comment-line edge case where a probe file has a malformed
+            # commented-out probe — ignored as a comment, not raised.
+            continue
+        gt = obj.get("ground_truth_text")
+        # Only record string overrides; non-strings (None, numbers)
+        # are treated as absence rather than as a coercion target. An
+        # operator who wrote `"ground_truth_text": null` clearly meant
+        # "fall back to the store"; honor that rather than silently
+        # stringifying.
+        if isinstance(gt, str) and gt.strip():
+            expected = obj.get("expected_fact_id")
+            if isinstance(expected, str):
+                overrides[expected] = gt
+    return probes, overrides
+
+
+# ── Subprocess construction (judge + generator) ────────────────────
+
+
+def _build_judge_cmd(config: BehavioralConfig) -> list[str]:
+    """Construct the argv vector for one judge subprocess invocation.
+
+    Lifted from `kai.memory_extraction._run_extractor` (the extractor's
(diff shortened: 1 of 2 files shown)


dcellison:
## Review by Kai

## PR Review: feat(eval) Layer 2 behavioral A/B harness

Overall the harness is well-designed and the tests are thorough. Three warnings and a few suggestions.

---

### Warning: format string injection on model-generated text

**`behavioral.py`, `_render_judge_user_payload` (~line 317)**

```python
return _JUDGE_USER_TEMPLATE.format(
    question=question,
    ground_truth_text=ground_truth_text,
    response_a=response_a,
    response_b=response_b,
)
```

`response_a` and `response_b` are model-generated text fed straight into `str.format()`. If either response contains a literal `{response_b}` or any `{named_key}` that matches a kwarg, it gets silently substituted. If it contains `{unknown_key}`, this raises `KeyError` and crashes the probe. The comment dismisses this as "an edge case no real probe will hit" - but model output is untrusted by definition and an adversarial memory could trigger it. Replace with string concatenation or f-string building to eliminate the risk entirely.

---

### Warning: RNG probe index is within scored list, not probe file

**`behavioral.py`, `_run_all_probes` (~line 745)**

```python
tasks = [_run_under_semaphore(p, i) for i, p in enumerate(probes)]
```

`probes` here is `scored_probes` (the drift-filtered subset). The per-probe RNG seed hashes `(config.seed, probe_index)` where `probe_index` is the position in the scored subset, not in the original file. If one earlier probe drifts between two runs, every subsequent probe gets a different `probe_index` and therefore a different arm assignment. The "reproducibility across reruns of the same probe set + user pair" guarantee in `_resolve_seed`'s docstring is silently broken whenever the drift set changes. The fix is to pass the probe's original file position (or `expected_fact_id`) as the seed input instead.

---

### Warning: unhandled OSError on output file write

**`behavioral.py`, `_run_cli` (~line 1530)**

```python
if args.output:
    args.output.write_text(json.dumps(output, indent=2), encoding="utf-8")
```

This runs after all 78 subprocess calls complete. An unwritable path or a full disk raises an unhandled `OSError`, crashing the process and losing the entire run's results. Wrap in `try/except OSError` and return exit code 1 with the summary already printed to stdout.

---

### Suggestion: `--max-budget-usd` flag unverified per conventions

**`behavioral.py`, `_build_judge_cmd` and `_build_gen_cmd`**

The CLAUDE.md rule is "verify CLI flags exist (via --help or docs) before adding them to subprocess commands." Both command builders pass `--max-budget-usd`. The PR says these flags are "reused" from the extractor, but the diff doesn't show evidence of a `claude --help` verification in this PR. If the flag doesn't exist in the installed CLI version, subprocesses silently fail and bucket every probe as a generation/judge error with no diagnostic message. Worth a comment citing the verification step.

---

### Suggestion: empty gold-fact not detected as drift

**`behavioral.py`, `_run_under_semaphore` (~line 758)**

```python
if not gt:
    gt = probe.notes
```

If both `fact.text` and `probe.notes` are empty strings, the judge receives a blank `GROUND-TRUTH FACT` field and produces an unreliable verdict that lands in a real outcome bucket (not `drift`). The comment acknowledges this is "genuinely unrecoverable" but doesn't surface it to the operator. At minimum, log a warning so the operator can see which probes had empty gold fields.

---

### Suggestion: `_capture_claude_cli_version` is synchronous in an async context

**`behavioral.py`, ~line 1000**

`subprocess.run(...)` blocks the event loop. Since this is called from the async `_run_cli` before the gather starts, it does not affect probe execution, but it's stylistically inconsistent and would block if the CLI hangs. Low priority but worth noting.

---

The logic for drift exclusion from denominators, the A/B isolation guarantee, the truth table in `_rollup_outcome`, and the subprocess sandboxing design are all correct. Test coverage for the eight-cell rollup, malformed judge output, and the cwd/system-prompt locks is solid.

dcellison:
## Review by Kai

## PR Review: feat(eval) Layer 2 behavioral A/B harness - follow-up

All three warnings from the prior review are cleanly addressed: format string injection fixed with concatenation (verified by the new `test_curly_braces_in_response_are_passed_through_literally` test), RNG seed now uses `probe.expected_fact_id` instead of positional index, and the OSError on file write is wrapped with exit code 1. The prior suggestions on empty gold-fact and `--max-budget-usd` verification are also handled. The remaining findings are new.

---

### Suggestion: `asyncio.get_event_loop()` should be `asyncio.get_running_loop()`

**`behavioral.py`, `_run_subprocess` (~line 340)**

```python
loop = asyncio.get_event_loop()
start = loop.time()
```

`asyncio.get_event_loop()` emits a `DeprecationWarning` in Python 3.10+ when called from a coroutine (it's deprecated in favor of `get_running_loop()`). Since this function is always called from within an async context, `asyncio.get_running_loop()` is the correct and safe replacement. The semantics are identical here; `get_running_loop()` just makes the requirement explicit and avoids the deprecation path.

---

### Suggestion: `_resolve_ground_truth` is `async def` with no `await`

**`behavioral.py`, `_resolve_ground_truth`**

The function is declared `async` but every call inside it is synchronous — `_mem.get_by_id` is called without `await`. At 26 probes with p50=9ms per call (as the docstring documents), this blocks the event loop for ~234ms before any probes launch. The prior review flagged the identical pattern in `_capture_claude_cli_version`; this function has the same shape. It does not affect correctness (it runs before `_run_all_probes` starts) but is worth making consistent with that earlier note, either by removing the `async` qualifier or by confirming `get_by_id` has an async variant worth using.

---

### Suggestion: `probe_id` does not correspond to source-file line number

**`behavioral.py`, `_run_cli` and `build_output_json`**

```python
all_outcomes = scored_outcomes + drift_outcomes
# ...
"per_probe": [_outcome_to_per_probe_dict(o, probe_id=i + 1) for i, o in enumerate(outcomes)],
```

`probe_id` is assigned by position in `all_outcomes` (scored probes first, drifted appended). If 3 probes drift, their `probe_id` values in the output JSON are N-2, N-1, N regardless of where they appear in the source file. An operator cross-referencing the output against the `.jsonl` file by `probe_id` will be off for any run where the drift set is non-empty or changes between runs. The `expected_fact_id` field is already present in each row and is the stable per-probe identity; using it as the primary cross-run key (and treating `probe_id` as informational only) would make this less surprising. Low-priority given the docstring acknowledges the concat ordering, but the field name implies source position.

---

The A/B isolation guarantee, the `expected_fact_id`-seeded RNG, the rollup truth table, subprocess sandboxing, and test coverage for all eight rollup cells are correct. The harness is ready for operator use.

dcellison:
## Review by Kai

This is a third-round review. All three warnings and the two suggestions from the prior two rounds have been addressed. The findings below are new.

---

## Warning: `except TimeoutError:` does not catch `asyncio.TimeoutError` on Python < 3.11

**`behavioral.py`, `_run_subprocess` (~line 440)**

```python
except TimeoutError:
    proc.kill()
    await proc.wait()
```

On Python 3.10 and earlier, `asyncio.wait_for` raises `asyncio.TimeoutError`, which inherits from `concurrent.futures.TimeoutError` - not from the builtin `TimeoutError` (an `OSError` subclass). The `except TimeoutError:` block never fires. As a result, `proc.kill()` and `proc.wait()` are skipped: the subprocess keeps running (or accumulates as a zombie), and the exception propagates to `_run_under_semaphore`'s catch-all `except Exception:`, which produces a `generation_error` with no subprocess cleanup. On Python 3.11+ the two classes are unified and this is safe.

If Python 3.11+ is the minimum, add a comment asserting that. If it is not, change to `except asyncio.TimeoutError:` to be unambiguous on all versions.

---

## Suggestion: `scored_count` includes error probes, not just verdicted probes

**`behavioral.py`, `build_output_json`**

```python
"scored_count": sum(counts.values()),
```

`counts` is the output of `_aggregate_outcomes` across all six buckets, so `scored_count` = (probe_count - drift_count), which includes `judge_error` and `generation_error` probes. An operator reading the output JSON might reasonably interpret `scored_count` as "probes that received a meaningful verdict" (win/loss/tie/both_wrong), but a run with 10 generation errors would still report `scored_count=26`. A comment on the field, or a rename to `attempted_count`, would prevent misinterpretation when comparing runs.

---

## Suggestion: exception log in `_run_under_semaphore` logs scored-subset index, not source position

**`behavioral.py`, `_run_under_semaphore` (~line 810)**

```python
log.warning("probe %d raised: %s", probe_index, e)
```

`probe_index` is the loop index over `scored_probes` (the drift-filtered subset), not the source-file line number. When a probe raises unexpectedly, the diagnostic log is harder to act on than it could be. `probe.expected_fact_id` is the stable cross-run identifier and is already in scope; using it here would make the warning immediately cross-referenceable against the probes file.

---

The A/B isolation guarantee, rollup truth table, OSError guard, RNG seeding, source-position `probe_id` fix, and the concatenation-based judge payload are all correct. Test coverage across malformed judge shapes, drift exclusion, and subprocess sandboxing is solid.

dcellison:
## Review by Kai

This is a fourth-round review. All findings from the prior three rounds have been addressed cleanly. The findings below are new.

---

## Warning: concurrent subprocess count overstated by 3x in comments

**`behavioral.py`, `_DEFAULT_MAX_CONCURRENCY` constant (~line 138) and `_run_all_probes` docstring (~line 800)**

```python
# _DEFAULT_MAX_CONCURRENCY = 4
# "so 4 in flight means up to 12 simultaneous claude processes"

# docstring:
# "so a Semaphore capacity of N translates to up to 3N concurrent claude subprocesses"
```

Both claims are wrong. The Semaphore is acquired before `_run_one_probe` runs gen-A, and the three subprocess calls inside each probe slot are `await`ed sequentially - gen-A completes before gen-B starts, gen-B completes before the judge starts. The event loop cannot schedule a second subprocess in the same slot while the first is in-flight. Therefore, N Semaphore slots = at most N concurrent Claude subprocesses, not 3N.

The practical risk: an operator reading "12 simultaneous processes" who wants to stay under a concurrency limit will set `--max-concurrency` to one-third of what they could safely use, getting 3x worse throughput for no reason. The default comment and the docstring should both say "up to N concurrent subprocesses."

---

## Suggestion: `probe_index` parameter in `_run_under_semaphore` is never used

**`behavioral.py`, `_run_under_semaphore` (~line 790) and `_run_all_probes` gather call (~line 870)**

```python
async def _run_under_semaphore(probe: Probe, probe_index: int) -> ProbeOutcome:
    # ... probe_index never referenced in the body
```

The parameter was meaningful in the earlier version where `probe_index` seeded the RNG. Since that round's fix, the RNG uses `probe.expected_fact_id` and the exception log uses `probe.expected_fact_id`. The parameter is accepted (so `enumerate` continues to work) but never read. Replace with `_` to document the intentional non-use, or change the gather line to `[_run_under_semaphore(p) for p in probes]` and drop the parameter.

---

## Suggestion: `_render_summary` latency stats include `judge_error` probe latencies

**`behavioral.py`, `_render_summary` (~line 1500)**

```python
judge_latencies = [
    p["latency_ms"]["judge"]
    for p in output.get("per_probe", [])
    if p["judge_choice"] not in {"drift", "generation_error"}
]
```

`judge_error` probes are not excluded. When the judge times out, its latency is the full `_DEFAULT_JUDGE_TIMEOUT_S` (60s). A run with 3 judge timeouts will report a p50 and max that reflect the ceiling, not the happy-path latency. This misleads an operator trying to tune the timeout or compare judge performance between model versions. Exclude `judge_error` from this filter, or compute it in a separate line so both views are surfaced.

---

Everything else - the A/B isolation guarantee, `expected_fact_id`-seeded RNG, `asyncio.get_running_loop()`, synchronous `_resolve_ground_truth`, `source_position_by_fact_id` for `probe_id`, `attempted_count` rename with comment, `expected_fact_id` in exception logs, and the `asyncio.TimeoutError` / Python 3.11+ note - is correct. Test coverage for the eight rollup cells, malformed judge shapes, drift exclusion, and subprocess sandboxing is solid.

dcellison:
## Review by Kai

This is a fourth-round review. All prior findings have been addressed. Three new findings below.

---

## Suggestion: `--max-concurrency` help text contradicts the corrected comments

**`behavioral.py`, `_build_parser` (~line 1620)**

```python
help=(
    f"Maximum probes in flight; each probe runs 3 subprocesses "
    f"sequentially (default: {_DEFAULT_MAX_CONCURRENCY})."
),
```

Round 4 corrected `_DEFAULT_MAX_CONCURRENCY` and `_run_all_probes` to say "N not 3N concurrent subprocesses." The argparse help string - the operator-facing surface - still leads with "each probe runs 3 subprocesses," which invites the same 3x-underuse misreading the prior warning described. Operators who read `--help` rather than the source will set `--max-concurrency` to one-third of what they could safely use. Suggest: "Maximum probes in flight simultaneously; subprocesses within a probe run sequentially (default: 4)."

---

## Suggestion: `_parse_judge_stdout` and `_extract_judge_cost_usd` double-parse `stdout`

**`behavioral.py`, `_run_one_judge` (~line 680)**

```python
parsed = _parse_judge_stdout(stdout)
cost = _extract_judge_cost_usd(stdout)
```

Both functions call `json.loads(stdout)` independently. On a 26-probe run this is 26 redundant parses - not a correctness issue and not a performance concern at this scale, but `_extract_judge_cost_usd` could accept the already-parsed `envelope: dict` as a parameter (with `_parse_judge_stdout` returning it on success) to eliminate the duplication. Low priority.

---

## Note: test file truncated - `captured.update` in Test 9 appears incomplete

**`tests/test_eval_behavioral.py`, `TestSubprocessExplicitCwd` (~line 875)**

```python
async def fake_exec(*cmd, **kwargs):
    captured.update
```

`captured.update` is a bare attribute reference with no call or arguments - it has no side effect. If this is the complete function body, the test captures nothing and any assertions on `captured` below would pass vacuously. The diff was truncated before the rest of `fake_exec` is visible, so I cannot confirm whether the next line completes the expression. Worth verifying that `captured.update(kwargs)` (or equivalent) is actually present in the unshown portion.

---

Everything else - the A/B isolation guarantee, `asyncio.get_running_loop()`, synchronous `_resolve_ground_truth`, `source_position_by_fact_id` for stable `probe_id`, `attempted_count` rename with comment, `expected_fact_id` in exception logs, `asyncio.TimeoutError` / Python 3.11+ comment, corrected concurrency constant and docstring, removed `probe_index` parameter, and `judge_error` exclusion from latency stats - is correct. The harness is ready for operator use subject to the note on the truncated test.

dcellison:
## Review by Kai

This is a fifth-round review. All five findings from round 5 appear addressed: the `--max-concurrency` help text is corrected, `_decode_judge_envelope` is called once and shared between `_extract_judge_choice` and `_extract_judge_cost_usd`, the `probe_index` parameter is gone from `_run_under_semaphore`, `judge_error` latencies are excluded from `_render_summary`, and the concurrency constant and docstring are correct. Three new findings below.

---

### Note: Test 9 (`captured.update`) still unverifiable

The diff is truncated at line 911 of the test file, which is the same boundary as the prior round. `TestSubprocessExplicitCwd` (test 9) and `TestGeneratorEmptySystemPrompt` (test 10) are not visible. The prior round's concern - that `captured.update` (bare attribute reference, no call) might silently make assertions vacuous - cannot be confirmed or cleared without seeing the full file. This note carries over; worth verifying out-of-band.

---

### Suggestion: `_make_outcome` test helper sets wrong `judge_choice` for non-win/lose outcomes

**`tests/test_eval_behavioral.py`, `_make_outcome` (~line 107)**

```python
judge_choice="A_wins" if memory_outcome == "memory_wins" else "B_wins",
```

When `memory_outcome` is `"tie"`, `"both_wrong"`, `"judge_error"`, or `"generation_error"`, the helper encodes `judge_choice = "B_wins"` even though the `memory_outcome` field says something else. The current aggregation tests only read `memory_outcome`, so no test passes vacuously today. But a future test that reads `judge_choice` alongside `memory_outcome` (e.g. testing `_outcome_to_per_probe_dict` serialization) would get an inconsistent fixture without any visible signal. A more explicit mapping (`judge_choice=memory_outcome` for passthrough states, or at least matching `"tie"` -> `"tie"`) would make the helper trustworthy for any read path.

---

### Suggestion: `_run_subprocess` discards `stderr` on non-zero exit

**`behavioral.py`, `_run_one_arm` and `_run_one_judge` (~lines 570, 620)**

```python
rc, stdout, _stderr, elapsed_ms = await _run_subprocess(...)
if rc != 0:
    return None, elapsed_ms, None
```

`_stderr` is captured but thrown away. When a subprocess exits non-zero (rate limit, unknown flag, OOM kill), the stderr content is the primary diagnostic signal. An operator debugging a 30% `generation_error` run has no way to see what the subprocess actually said. A `log.debug("subprocess stderr: %s", _stderr[:500].decode("utf-8", errors="replace"))` (or warning-level on rc != 0) on the non-timeout failure path would make these buckets actionable without adding stdout noise to clean runs.

---

The production logic - A/B isolation, `expected_fact_id`-seeded RNG, OSError guard, `asyncio.get_running_loop()`, `TimeoutError`/Python 3.13 pin, `source_position_by_fact_id` for stable `probe_id`, `attempted_count` with comment, `judge_error` exclusion from latency stats, and corrected concurrency semantics throughout - is all correct. The harness is ready for operator use pending the note on the truncated test.

dcellison:
## Review by Kai

## Round 6 Review: feat(eval) Layer 2 behavioral A/B harness

All findings from rounds 1-5 are addressed in the current diff. Three prior-round suggestions confirmed fixed: `_make_outcome` now uses a proper `judge_choice_by_outcome` dict mapping (covering tie/both_wrong/error passthrough), `_run_subprocess` logs stderr on non-zero exit, and `_render_summary` correctly excludes `judge_error` latencies from the latency stats filter. Nothing new to raise on those.

---

### Carryover note (still unverifiable)

`tests/test_eval_behavioral.py`, `TestSubprocessExplicitCwd` and `TestGeneratorEmptySystemPrompt` remain invisible - the diff is truncated at line 922 mid-assertion (`assert ra`). Since `make test` passes with 44 new tests, the actual file is complete, but the truncation means the cwd-lock and empty-system-prompt-lock tests cannot be confirmed correct from the diff alone. Worth a local file read out-of-band.

---

### Suggestion: `proc.returncode or 0` is opaque about intent

**`behavioral.py`, `_run_subprocess` (~line 480)**

```python
rc = proc.returncode or 0
```

After `communicate()` returns, `returncode` is always set (process has terminated), so the `or 0` guard is defending against a state that cannot occur on the success path. The expression is correct - `0 or 0 == 0`, any non-zero value passes through - but a reader unfamiliar with asyncio subprocess semantics has to think twice about whether this handles a negative exit code correctly. `proc.returncode if proc.returncode is not None else 0` makes the intent unambiguous and satisfies the type checker more explicitly. Very low priority.

---

### Note: `_resolve_ground_truth` docstring cross-reference is slightly misleading

**`behavioral.py`, `_resolve_ground_truth` docstring**

> Matches the synchronous shape of `_capture_claude_cli_version`.

The cross-reference is accurate - both functions are synchronous and called before the gather. But naming a sibling function in a docstring to explain a design choice couples two unrelated functions; if `_capture_claude_cli_version` is ever moved or renamed, the reference rots silently. The note about "synchronous because get_by_id is synchronous; no awaits; blocks the event loop for ~234ms" is the load-bearing sentence. Dropping the cross-reference wouldn't weaken it. Cosmetic only.

---

The production logic is correct throughout: A/B isolation, `expected_fact_id`-seeded RNG, OSError guard, `asyncio.get_running_loop()`, `TimeoutError`/Python 3.13 pin, `source_position_by_fact_id` for stable `probe_id`, `attempted_count` with comment, `judge_error` exclusion from latency stats, corrected concurrency semantics, single envelope decode per judge call, and subprocess sandboxing are all sound. The harness is ready for operator use pending verification of tests 9 and 10.

dcellison:
## Review by Kai

This is a seventh-round review. All six prior rounds' findings are addressed. Three new findings below.

---

### Warning: `--max-concurrency 0` deadlocks the entire run

**`behavioral.py`, `_run_all_probes` and `_build_parser`**

```python
sem = asyncio.Semaphore(config.max_concurrency)
# ...
async with sem:   # blocks forever if counter starts at 0
```

`argparse` accepts `--max-concurrency 0` without error (`type=int` performs no range check). `asyncio.Semaphore(0)` starts with an internal counter of zero; every `acquire()` call blocks indefinitely waiting for a `release()` that never comes because no task can ever hold the semaphore. The harness hangs silently - no output, no timeout, no error message - consuming the operator's time with no diagnostic.

(`-1` at least crashes with `ValueError: Semaphore initial value must be >= 0`, which is recoverable. Zero is the worse case.)

Fix: add a `min_value` guard in `_build_parser` (a custom `type=` that raises `argparse.ArgumentTypeError` for `< 1`) or a `sys.exit(1)` check at the top of `_run_cli` before the semaphore is created.

---

### Suggestion: `_arm_letter_for_memory` docstring states a false bias

**`behavioral.py`, `_arm_letter_for_memory` (~line 590)**

```python
# random.choice is biased toward the first element when the seed is small, but the
# bias is well below 1% for our 16-hex-digit (64-bit) seeds and over 26 probes
```

This claim is not correct. Python's `random.choice` on a 2-element sequence uses the Mersenne Twister and is exactly 50/50 regardless of seed magnitude; there is no bias toward the first element. The code itself is correct - the comment is the problem. A future reader investigating a skewed win-rate could waste time chasing a phantom ~1% bias that does not exist. The sentence should be removed or replaced with the accurate statement that the coin is fair.

---

### Carryover note: Tests 9 and 10 remain unverifiable

`tests/test_eval_behavioral.py` is truncated at line 922 mid-sentence (same boundary as rounds 5 and 6). `TestSubprocessExplicitCwd` and `TestGeneratorEmptySystemPrompt` are not visible in the diff. `make test` passes with 44 new tests, so both tests exist and run - but the correctness of the cwd-lock assertion and the literal-empty-string `--system-prompt` assertion cannot be confirmed from the diff alone.

---

Everything else is sound: A/B isolation guarantee, `expected_fact_id`-seeded RNG, `asyncio.get_running_loop()`, `asyncio.TimeoutError`/Python 3.13 pin, `source_position_by_fact_id` for stable `probe_id`, `attempted_count` with comment, `judge_error` exclusion from latency stats, corrected concurrency semantics throughout, single envelope decode per judge call, `judge_choice_by_outcome` mapping in test helper, stderr logging on non-zero subprocess exit, and subprocess sandboxing are all correct.

dcellison:
## Review by Kai

This is an eighth-round review. All seven prior rounds' findings are addressed in the current diff. The `--max-concurrency 0` deadlock is fixed via `_positive_int`, the false bias claim in `_arm_letter_for_memory` is corrected, and the round 6 `proc.returncode or 0` note is resolved. Two new findings below.

---

### Carryover note: truncation boundary has shifted

**`tests/test_eval_behavioral.py`**, diff cutoff at line 950

Prior rounds flagged Tests 9 and 10 (`TestSubprocessExplicitCwd`, `TestGeneratorEmptySystemPrompt`) as unverifiable. Those remain invisible. Additionally, the current diff cuts off mid-body inside `test_one_arm_failure_also_buckets_as_generation_error` (Test 7, second method) - the `asyncio.run(...)` call is present but the assertions that follow it are not visible. Since `make test` passes with 44 new tests the file is complete, but three tests cannot be confirmed correct from the diff alone.

---

### Suggestion: `_resolve_seed` guard is `if cli_seed:` not `if cli_seed is not None:`

**`behavioral.py`, `_resolve_seed`**

```python
if cli_seed:
    return cli_seed
```

An empty string (`--seed ""`) evaluates as falsy and falls through to the hash-based default, silently ignoring the operator's explicit override. `if cli_seed is not None:` is the precise guard that matches argparse's own `default=None` contract and would honor an empty string as a deliberate choice. Very low priority - `--seed ""` is an unusual invocation - but the current form is technically inconsistent with the parameter's contract.

---

Everything else is sound: `_positive_int` for `--max-concurrency`, `asyncio.get_running_loop()`, `asyncio.TimeoutError`/Python 3.13 pin, `expected_fact_id`-seeded RNG, `source_position_by_fact_id` for stable `probe_id`, `attempted_count` with comment, `judge_error` exclusion from latency stats, corrected concurrency semantics throughout, single envelope decode per judge call, `judge_choice_by_outcome` mapping in test helper, stderr logging on non-zero subprocess exit, and OSError guard on output write are all correct. The harness is ready for operator use pending out-of-band verification of the three truncated tests.

dcellison:
## Review by Kai

## Round 9 Review: feat(eval) Layer 2 behavioral A/B harness

All findings from rounds 1-8 are addressed in the current diff. No new warnings or critical issues found. One carryover note and one minor suggestion.

---

### Carryover note: three tests still unverifiable from the diff

`tests/test_eval_behavioral.py` is truncated at line 961, inside `test_one_arm_failure_also_buckets_as_generation_error` (Test 7, method 2) mid-parameter-list (`fake_run_subprocess(*, cmd, stdin_payload, ti`). Tests 9 (`TestSubprocessExplicitCwd`) and 10 (`TestGeneratorEmptySystemPrompt`) are entirely invisible. `make test` passing with 44 new tests confirms the file is complete, but the cwd-lock, the empty-system-prompt-lock, and the second method of the one-arm-failure test cannot be confirmed correct from the diff alone. Worth a local file read out-of-band.

---

### Suggestion: `_pick_qualitative_examples` uses `elif` instead of independent `if` branches

**`behavioral.py`, `_pick_qualitative_examples`**

```python
if o.memory_outcome == "memory_wins" and not examples["best_help"]:
    examples["best_help"] = _example_dict(o)
elif o.memory_outcome == "memory_loses" and not examples["worst_hurt"]:
    ...
elif o.memory_outcome == "tie" and not examples["clearest_tie"]:
    ...
```

The `elif` chain is correct today because `memory_outcome` is a single value that cannot simultaneously be `memory_wins` and `memory_loses`. But the `elif` framing requires a reader to reason about mutual exclusivity to verify it. Three independent `if` branches would make the intent obvious without any behavioral change. Very low priority.

---

Everything else is sound: A/B isolation guarantee, `expected_fact_id`-seeded RNG, `asyncio.get_running_loop()`, `asyncio.TimeoutError`/Python 3.13 pin, `source_position_by_fact_id` for stable `probe_id`, `attempted_count` with comment, `judge_error` exclusion from latency stats, `_positive_int` for `--max-concurrency`, corrected concurrency semantics and help text throughout, single envelope decode per judge call, `judge_choice_by_outcome` mapping in test helper, stderr logging on non-zero exit, `OSError` guard on output write, and `is not None` guard in `_resolve_seed` are all correct. The harness is ready for operator use pending out-of-band verification of the three truncated tests.

dcellison:
## Review by Kai

## Round 10 Review: feat(eval) Layer 2 behavioral A/B harness

All findings from rounds 1-9 are addressed. Round 9's `elif` -> independent `if` branches in `_pick_qualitative_examples` is confirmed fixed. No new warnings or bugs found.

---

### Carryover note: three tests still unverifiable from the diff

`tests/test_eval_behavioral.py` truncates at line 961 mid-expression (`assert outcome.latency_ms["judg`), same boundary as rounds 5-9. `TestSubprocessExplicitCwd` (test 9) and `TestGeneratorEmptySystemPrompt` (test 10) are entirely invisible, and the second assertion of `test_one_arm_failure_also_buckets_as_generation_error` is cut off. `make test` passing with 44 new tests confirms the file is complete, but the cwd-lock and literal-empty-string `--system-prompt` assertions remain unverifiable from the diff alone.

---

Everything else is sound: A/B isolation guarantee, `expected_fact_id`-seeded RNG, `asyncio.get_running_loop()`, `asyncio.TimeoutError` / Python 3.13 pin, `source_position_by_fact_id` for stable `probe_id`, `attempted_count` with comment, `judge_error` exclusion from latency stats, `_positive_int` for `--max-concurrency`, corrected concurrency semantics and help text throughout, single envelope decode per judge call, `judge_choice_by_outcome` mapping in test helper, stderr logging on non-zero subprocess exit, `OSError` guard on output write, `is not None` guard in `_resolve_seed`, and independent `if` branches in `_pick_qualitative_examples` are all correct. The harness is ready for operator use pending out-of-band verification of the three truncated tests.

PR #371, opened 2026-04-23
feat(eval): synthetic Track 1 reconstruction for Layer 2 pollution test

dcellison:
## Summary

Adds an eval-only `--pollution-lines N` flag to the Layer 2 behavioral A/B harness. When set, the memory-on arm's memory context block is appended with N synthetic `- User said: <text>` bullets sampled deterministically per probe from the user's actual chat history at `DATA_DIR/history/<user_id>/`. The memory-off arm is never polluted.

The motivation is the open Layer 2 hypothesis: did removing raw-utterance pollution (the deleted ingestion path) matter more than adding semantic retrieval? Reconstructing Track 1's full machinery to answer that is hundreds of lines plus a real risk surface (one bit-flip of an "eval-only" flag in production reintroduces the original input-loss bug). Synthetic pollution mimics the prompt SHAPE Track 1 produced without resurrecting the production code path, so the comparison can run safely.

## Design invariants (locked by tests)

1. `--pollution-lines 0` (the default) is byte-identical to today's harness behavior. The existing A/B isolation invariant continues to hold, and the established memory-on baseline remains directly comparable.
2. The pollution sampler uses a SEPARATE RNG seeded with `sha256(seed + fact_id + b"pollution")[:16]`. The A/B coin-flip RNG omits the `b"pollution"` salt, so enabling pollution does NOT shift arm assignments; pollution-on vs pollution-off runs against the same `(probes, user_id, seed)` triple stay comparable per probe.
3. The memory-off arm ignores `pollution_lines` even when supplied. The asymmetry IS the experimental contrast.
4. Per-line truncation matches the deleted Track 1 cap (2000 chars), so the reconstructed prompt shape is faithful rather than artificially shorter.

## Run procedure

Snapshot must include `history/` alongside `memory/`:

```
rm -rf /tmp/kai-eval-snapshot && mkdir -p /tmp/kai-eval-snapshot
cp -R <data_dir>/memory /tmp/kai-eval-snapshot/memory
cp -R <data_dir>/history /tmp/kai-eval-snapshot/history
rm -f /tmp/kai-eval-snapshot/memory/qdrant/.lock
```

Three-way compare (clean baseline / moderate pollution / heavy pollution):

```
KAI_DATA_DIR=/tmp/kai-eval-snapshot \
  MEM0_DIR=/tmp/kai-eval-snapshot/mem0-telemetry \
  MEMORY_ENABLED=true \
  .venv/bin/python -m kai.eval.behavioral \
  --probes home/evals/probes.jsonl --user-id <user_id> \
  --pollution-lines 0 --output /tmp/behavioral-clean.json
```

Repeat with `--pollution-lines 10` and `--pollution-lines 20` to walk up the noise dose. The clean run should match the most recent published baseline; deviations are pollution effect.

## Test plan

- [x] Lint clean (`ruff check`, `ruff format --check`)
- [x] Full pytest suite: 2198 tests pass (21 new in `TestPollutionInjection`)
- [x] No em or en dashes in the diff
- [ ] Manual: run `--pollution-lines 0` against a fresh snapshot; verify outcomes match the established Layer 2 baseline within noise
- [ ] Manual: run `--pollution-lines 10` and `--pollution-lines 20`; record headline win/loss/tie/both_wrong shifts vs the clean run

### DIFF
--- a/src/kai/eval/behavioral.py
+++ b/src/kai/eval/behavioral.py
@@ -295,6 +295,17 @@ class BehavioralConfig:
     # arg) so it surfaces in the output JSON automatically alongside
     # seed and max_concurrency, making each saved run self-describing.
     pollution_lines: int = 0
+    # Snapshot of load_config().data_dir captured once at run start.
+    # Threaded through BehavioralConfig (rather than re-loaded inside
+    # _run_all_probes) so the memory snapshot read by init_memory and
+    # the history snapshot read by _load_user_history_messages share a
+    # single source of truth. If KAI_DATA_DIR were ever to change
+    # between two load_config() calls (subprocess wrapper rewriting
+    # the env mid-run, fixture cleanup racing the eval), the two
+    # snapshots could diverge silently. Default points at the dev tree
+    # so test fixtures that build a config without invoking the CLI
+    # still get a sensible path.
+    data_dir: Path = Path(".")
 
 
 @dataclass
@@ -818,6 +829,17 @@ def _load_user_history_messages(user_id: str, data_dir: Path) -> list[str]:
     after a power-loss restart) rather than failing the whole load.
     The bot's own history reader follows the same convention.
     """
+    # Path-traversal guard: user_id flows in from --user-id with no
+    # validation upstream, and we are about to use it as a path
+    # component under data_dir. A value like "../../etc" would escape
+    # the history root entirely. Path(user_id).name strips any
+    # directory parts and any platform-specific separators; equality
+    # against the input means user_id is exactly one segment, no
+    # slashes, no parent refs. Eval-only operator-run code, so the
+    # guard is consistent with _positive_int / _non_negative_int rather
+    # than a security boundary.
+    if Path(user_id).name != user_id or user_id in ("", ".", ".."):
+        raise ValueError(f"invalid user_id for history path: {user_id!r}")
     history_dir = data_dir / "history" / user_id
     if not history_dir.is_dir():
         return []
@@ -873,9 +895,19 @@ def _sample_pollution_for_probe(
     """
     if n <= 0 or not history:
         return []
+    # Domain-separated SHA256 update: each field is followed by a NUL
+    # delimiter so concatenation cannot collide. Without delimiters,
+    # ("ab", "cde", "pollution") and ("abc", "de", "pollution") feed
+    # the same bytes into the hash and would draw identical samples.
+    # NUL is the conventional choice because it cannot appear in the
+    # UTF-8 encoding of any of these fields (base_seed is hex chars,
+    # expected_fact_id is the upstream UUID/string, neither of which
+    # contains a literal 0 byte).
     h = hashlib.sha256()
     h.update(base_seed.encode("utf-8"))
+    h.update(b"\x00")
     h.update(expected_fact_id.encode("utf-8"))
+    h.update(b"\x00")
     h.update(b"pollution")
     rng = random.Random(int(h.hexdigest()[:16], 16))
     return rng.sample(history, min(n, len(history)))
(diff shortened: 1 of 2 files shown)


dcellison:
## Review by Kai

## PR Review: feat(eval) synthetic Track 1 reconstruction

Overall this is a well-designed, well-tested eval feature. The invariants are locked down clearly and the test suite is thorough. A few issues worth addressing:

---

### Warning: Hash construction lacks domain separator

`behavioral.py`, `_sample_pollution_for_probe`:
```python
h.update(base_seed.encode("utf-8"))
h.update(expected_fact_id.encode("utf-8"))
h.update(b"pollution")
```

SHA256's `update()` is purely concatenative, so `("ab", "cde")` and `("abc", "de")` produce identical hashes. Two distinct `(base_seed, fact_id)` pairs with the same string concatenation will draw the same pollution sample. This won't affect the A/B independence guarantee (which only requires the same pair to differ between the two RNGs), but if seeds and fact IDs ever share a character domain, two different probes could get identical pollution sets, silently undermining the per-probe determinism property the tests lock down. Fix: add a delimiter byte between updates, e.g. `h.update(b"\x00")` between the seed and fact_id.

---

### Warning: `load_config()` re-executed inside `_run_all_probes`

`behavioral.py`, inside `_run_all_probes`:
```python
from kai.config import load_config
data_dir = load_config().data_dir
```

Config was already loaded in `_run_cli` to build `BehavioralConfig`, but `BehavioralConfig` doesn't carry `data_dir`, forcing a second `load_config()` call inside the async runner. If `KAI_DATA_DIR` is unset or changes between those two points (unlikely, but possible with subprocess wrappers), the history path would diverge from the memory path. The cleaner fix is adding `data_dir: Path` to `BehavioralConfig` and populating it in `_run_cli`, eliminating the re-load entirely.

---

### Suggestion: `user_id` not sanitized in path construction

`behavioral.py`, `_load_user_history_messages`:
```python
history_dir = data_dir / "history" / user_id
```

`user_id` comes from `--user-id` CLI args with no traversal guard. A value like `../../../etc` would escape the history directory. This is eval-only operator-run code, so it is low severity, but adding `Path(user_id).name == user_id` or a simple `assert "/" not in user_id` at the function entry would be consistent with the input-validation discipline applied elsewhere (`_positive_int`, `_non_negative_int`).

---

### Suggestion: Redundant help text in `--pollution-lines`

`behavioral.py`, `_build_parser`:

> "Defaults to 0 (no injection); set to 0 for the clean A/B baseline."

The second clause restates the first. Minor, but the help string is already long.

---

### Conventions check

- Comments are thorough, as required.
- No em/en dashes found in the diff (the `──` dividers are U+2500 box-drawing characters, not dashes, consistent with existing file style).
- Branch name `feature/eval-pollution-injection` follows the required prefix/kebab convention.
- No new env vars, so the env-var pattern checklist doesn't apply.
- `ruff check` / `ruff format` pass per test plan.

dcellison:
## Review by Kai

## PR Review: feat(eval) synthetic Track 1 reconstruction (round 2)

All three prior-review items were addressed cleanly:
- NUL delimiters added between SHA256 fields
- `data_dir` threaded through `BehavioralConfig`; no second `load_config()` call
- Path-traversal guard on `user_id` added with parametrized tests covering common payloads

One new finding:

---

### Warning: Help text contradiction in `--pollution-lines`

`behavioral.py`, `_build_parser`, the new argument's `help=` string:

```
"The memory-off arm "
"is never polluted, by design; pollution is a property of "
"the memory-off prompt."
```

The last clause says "pollution is a property of the memory-**off** prompt," directly contradicting the sentence before it. It should be "memory-**on** prompt." An operator reading this help text to decide whether to copy the flag into a production invocation would get the wrong mental model about which arm receives the injected lines.

---

Everything else looks clean:

- The RNG independence test correctly mirrors the actual A/B seeding (no NUL delimiters on the A/B side; the test comment acknowledges this and explains why it is intentional).
- `_inject_synthetic_pollution` no-op on empty list correctly protects the byte-level isolation guarantee.
- `_initialize_memory` return-type change from `bool` to `Path | None` is propagated consistently to the caller and the mock in `test_write_failure_returns_exit_1`.
- `_non_negative_int` mirrors `_positive_int` discipline; both the validator and its three test parametrizations are correct.
- `_load_user_history_messages` reads files in `sorted()` order and the multi-file test asserts lexicographic ordering, matching the claimed bot convention.
- No em/en dashes in the diff. The `──` dividers are U+2500 box-drawing, not dashes.
- No new env vars, so the env-var checklist does not apply.

PR #372, opened 2026-04-23
fix(eval): _initialize_memory uses kai.config.DATA_DIR (module constant)

dcellison:
## Summary

Fix-forward for the synthetic-pollution PR that was just merged. `_initialize_memory` returned `config.data_dir`, but `Config` (the dataclass returned by `load_config`) has no `data_dir` field — `DATA_DIR` is a module-level constant in `kai.config`, evaluated at import time from `KAI_DATA_DIR`. Every CLI invocation failed at startup with `eval: init failed: 'Config' object has no attribute 'data_dir'`.

## Why the bug shipped

The behavioral test suite mocks `_initialize_memory` wholesale (`patch.object(behavioral, "_initialize_memory", return_value=tmp_path)`), so the real attribute access never ran under test. The CI suite was green; the broken code only surfaced on the first real eval-campaign launch.

## Fix

Import `DATA_DIR` from `kai.config` instead of reading a non-existent attribute on the loaded config:

```python
from kai.config import DATA_DIR, load_config
...
return DATA_DIR
```

The CLI sets `KAI_DATA_DIR` before the Python process starts, so the import-time evaluation captures the right path.

## Regression tests

Three new tests in `TestInitializeMemoryDataDirContract`:

1. `test_kai_config_exposes_data_dir_constant` - asserts `kai.config.DATA_DIR` exists and is a `Path`. Locks the import contract; if a future config refactor renames or removes the constant, this fails before the broken-CLI symptom.
2. `test_initialize_memory_returns_path_when_memory_enabled` - exercises the live success branch by mocking only the collaborators (`load_config`, `init_memory`, `is_enabled`), letting `_initialize_memory`'s own body run. The previously-shipped bug would have failed this test.
3. `test_initialize_memory_returns_none_when_memory_disabled` - exercises the live disabled branch, locking the `Path | None` return-type contract.

The lesson is broader than this one bug: a function that's mocked end-to-end across the whole test suite has zero unit coverage on its body, so any internal contract change can ship green. The new tests stub only the collaborators, not the function itself.

## Verification

- `make check` clean
- Full pytest: 2209 pass (3 new)
- `python -m kai.eval.behavioral --help` launches without error against the snapshot env
- No em or en dashes in the diff

## Test plan

- [x] Lint clean
- [x] Full pytest 2209/2209
- [x] CLI `--help` smoke test
- [ ] Manual: `--pollution-lines 0` end-to-end against a fresh snapshot reaches `Wrote /tmp/...json` (the symptom that surfaced this bug)

### DIFF
--- a/src/kai/eval/behavioral.py
+++ b/src/kai/eval/behavioral.py
@@ -295,16 +295,16 @@ class BehavioralConfig:
     # arg) so it surfaces in the output JSON automatically alongside
     # seed and max_concurrency, making each saved run self-describing.
     pollution_lines: int = 0
-    # Snapshot of load_config().data_dir captured once at run start.
-    # Threaded through BehavioralConfig (rather than re-loaded inside
+    # Snapshot of kai.config.DATA_DIR captured once at run start.
+    # Threaded through BehavioralConfig (rather than re-imported inside
     # _run_all_probes) so the memory snapshot read by init_memory and
     # the history snapshot read by _load_user_history_messages share a
-    # single source of truth. If KAI_DATA_DIR were ever to change
-    # between two load_config() calls (subprocess wrapper rewriting
-    # the env mid-run, fixture cleanup racing the eval), the two
-    # snapshots could diverge silently. Default points at the dev tree
-    # so test fixtures that build a config without invoking the CLI
-    # still get a sensible path.
+    # single source of truth. DATA_DIR is a module-level constant
+    # evaluated at import time from KAI_DATA_DIR; storing it on config
+    # makes the binding visible at the call sites instead of relying
+    # on every reader to re-import. Default points at the dev tree so
+    # test fixtures that build a config without invoking the CLI still
+    # get a sensible path.
     data_dir: Path = Path(".")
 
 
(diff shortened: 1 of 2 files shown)


dcellison:
## Review by Kai

The fix is correct and the PR is clean. Here are my findings:

**No critical issues.**

---

**Warning (test correctness - verify mentally before merging)**

`test_initialize_memory_returns_path_when_memory_enabled` patches `DATA_DIR` via `monkeypatch.setattr(_cfg, "DATA_DIR", tmp_path)`, but the import inside `_initialize_memory` is `from kai.config import DATA_DIR` - a function-local import that runs at call time. This works correctly: Python resolves `from kai.config import DATA_DIR` by reading `sys.modules['kai.config'].DATA_DIR` at that moment, which is the monkeypatched value. Same logic applies to `patch("kai.memory.init_memory/is_enabled")` - function-local `from kai.memory import ...` grabs the patched attribute from the already-imported module. Not a bug, but the mechanics are subtle enough to deserve a note for future readers. The test is valid.

---

**Suggestion (minor, pre-existing)**

All imports inside `_initialize_memory` live inside the `try` block (`from kai.config import DATA_DIR, load_config` and `from kai.memory import ...`). This is pre-existing style, not introduced here. It's harmless but unusual - if either import fails for a reason unrelated to memory config (e.g., a broken `kai.memory` install), the exception is silently swallowed and `None` is returned, printing only `eval: init failed: ...`. Not worth changing in this PR since it's out of scope, but the catch-all `except Exception` on line 1929 is worth flagging if `_initialize_memory` ever gets a dedicated follow-up.

---

**Convention check:** Comments are thorough and accurate per the project style. No em/en dashes. Fix is tightly scoped. Tests stub only collaborators, not the function itself - exactly the right lesson applied.

The root cause analysis is correct, the fix is minimal and accurate, and the three regression tests are well-constructed. Good to merge.

dcellison:
## Review by Kai

The fix is correct and the PR is clean. Nothing new to raise beyond what the prior review covered.

**Fix correctness:** The change from `return config.data_dir` to `return DATA_DIR` is accurate. `DATA_DIR` is a module-level constant in `kai.config`; the `Config` dataclass has no `data_dir` field. `load_config()` is still called and passed to `init_memory`, so its import remains justified.

**Tests:**

- `test_initialize_memory_returns_none_when_memory_disabled` correctly omits the `DATA_DIR` monkeypatch - `is_enabled()` returns `False` so the function hits `return None` before ever reading `DATA_DIR`. No gap here.
- The enabled-path test properly exercises the function's body rather than mocking the function itself, which is exactly the lesson stated in the PR description.
- `test_kai_config_exposes_data_dir_constant` is a useful import-contract lock.

**Convention check:** Comments are accurate and appropriately detailed, no em/en dashes, fix is tightly scoped with no unrelated changes.

The two pre-existing concerns (subtle monkeypatch mechanics, catch-all `except Exception` swallowing unrelated import failures) were already noted in the prior review and are out of scope here.

Good to merge.

PR #374, opened 2026-04-23
feat(eval): Layer 3 longitudinal friction analysis harness

### LINKED ISSUE #373: Layer 3: longitudinal friction analysis
## Summary

Final sub-issue under #362. Build an observational friction analysis that counts memory-relevant friction events per week in the operator's chat history, bucketed by the public milestones defined in the epic body. Unlike Layer 1 (retrieval-only) and Layer 2 (behavioral A/B), this runs against real historical traffic rather than a contrived probe set, so it is the only layer that can answer "did the production user actually experience fewer memory-related frictions after each change shipped."

## Motivation

The Layer 2 result (see the `Layer 2 evaluation: results` comment on #362) established that synthetic Track 1-shaped pollution at N=0, 10, and 20 produces a flat 92.31% win rate against the current 26-probe set. That is a strong null within a small, author-curated probe set. Layer 3 exists to test whether the same conclusion holds in the noisier world of real chat traffic, where the friction signals are not curated questions but actual moments where memory either helped or failed to help.

The epic body names three natural bucket boundaries that correspond to behaviorally distinct memory configurations:

- Before #333 merged (2026-04-17): no semantic memory; facts only in `MEMORY.md`.
- Between #333 and #361 (2026-04-17 to 2026-04-22): semantic memory live, Track 1 raw-utterance ingestion active, no current-message delimiter.
- After #361 merged (2026-04-22): extracted-only ingestion, structural delimiter in place.

If Layer 2's conclusion (retrieval is load-bearing, Track 1 removal was clean-up) holds in real traffic, Layer 3 should show: a material friction drop at the #333 boundary, and roughly flat friction across the #361 boundary. Any other pattern is a surprise worth investigating.

## Scope

Build a CLI that reads an operator's chat history under `DATA_DIR/history/<user_id>/`, applies a small set of pre-defined friction-detection patterns, buckets events by week and by milestone, normalizes by message volume, and writes a structured JSON report plus a human-readable summary.

### Friction signals (pre-committed)

Four signal families, defined in advance so the analysis is not post-hoc pattern-hunting:

1. **Operator-side frustration markers.** User messages containing phrases like "I already told you", "I said earlier", "as I mentioned", "remember I", "you forgot", "why are you asking again". These are the strongest possible friction signal because the operator explicitly names the failure mode.

2. **Kai asking for information previously given.** Kai messages containing clarification-requests whose answer appears earlier in the same chat history. Approximated with: Kai message ending in a question AND a prior user message containing at least one content-word from Kai's question. Noisy but cheap to compute; the bucket-level trend matters more than per-event precision.

3. **Preference corrections.** User messages containing imperative-style corrections of Kai's behavior: "don't", "stop doing", "not that", "please use", "always use". Corrections that Kai should have known from a prior preference fact are the highest-value subset; we will report both the raw count and the subset that overlaps with known `preference`-tagged facts in the memory store at the time the message was sent.

4. **Repeated-fact statements.** Cases where the user restates a fact they previously provided. Detected as: user message whose content n-gram overlap with an earlier user message exceeds a threshold, where the earlier message contained a fact-shaped assertion (heuristic: "I am", "my", "we use", "I prefer", "I always"). This is the noisiest of the four and we expect it to carry the least weight.

Each signal emits a structured event: `{timestamp, bucket, signal_family, surface_text_snippet, context_message_ids}`. Surface text is redacted before any public reporting (see Privacy below).

### Normalization

Raw counts are meaningless because traffic is not uniform across buckets. Every signal is reported as a rate per 100 user-turns in the bucket, alongside the raw count and the bucket's user-turn total.

### Output shape

One JSON report per run, schema-versioned like Layer 1 and Layer 2 outputs. Fields:

- `buckets`: array of `{label, start, end, user_turns, messages_total}`.
- `events`: array of `{timestamp, bucket_label, signal_family, message_id}`. No surface text in the main array.
- `aggregates`: per-bucket-per-signal-family counts and per-100-turn rates.
- `trend_summary`: human-readable one-paragraph interpretation tied to the three-milestone prediction above.

Plus a redacted markdown summary suitable for pasting into a #362 comment, following the same pattern Layer 1 and Layer 2 established.

### Privacy

Chat history contains private user data by definition. The harness runs locally against the operator's own `DATA_DIR/history/`; it never uploads content anywhere. The default output report contains event counts and timestamps but NOT surface text. A `--include-snippets` flag enables short redacted excerpts for local inspection only, never for public posting.

### Bucket boundary correctness

The three merge dates are public and correspond to squash commits on `main`. The harness will read these as configuration constants, with a comment pointing back to the PR numbers. If the epic's bucket definitions change, one constant edit updates the analysis.

### Confounders acknowledged up-front

The epic body already names codebase improvements as a confounder. Additional confounders worth naming in the report: (a) operator behavior drift (the operator learned to phrase things better over time), (b) traffic composition drift (heavier use of the `/memory` command itself in recent weeks produces messages that mention memory without being friction events), (c) the `USER_MESSAGE_MARKER` introduced by PR #361 changed prompt structure in a way that could affect how Kai asks clarifying questions independent of memory retrieval.

The report will note these inline rather than burying them in a caveats appendix.

### Baseline smoke run

Before running against the full history, the harness should support a `--sample N` mode that runs against the N most recent days. Use this to sanity-check the signal definitions before processing the full dataset. This mirrors the probes-first, full-run-after cadence used in Layer 1 and Layer 2.

## Out of scope

- No retrieval-algorithm changes. If Layer 3 surfaces a miss pattern that suggests a better retrieval approach, that goes in a future epic, not here.
- No extraction-quality changes. Same logic.
- No auto-remediation (e.g., "when Kai asks for info it should know, inject it automatically"). Observation only.
- No cross-user analysis. Single operator's history is the unit.
- No real-time friction detection in the running bot. Offline analysis only.

## Acceptance

- CLI runs against a real `DATA_DIR/history/<user_id>/` tree and produces the JSON report and redacted markdown summary.
- All four signal families are implemented and covered by unit tests against synthetic transcripts.
- Bucket boundaries are correct per the three merge dates in the epic body.
- Normalization math is covered by tests.
- No surface text appears in the default output; opt-in flag gated.
- Dry-run mode (`--sample N days`) works.
- Harness tolerates missing or empty history directories without crashing.
- Report ends with a trend interpretation that explicitly names which of the three predicted patterns the data matched, including "surprise" (none of them).

## Parent

#362

dcellison:
## Summary

- Adds `src/kai/eval/friction.py`: a single-user, three-bucket longitudinal friction-trend evaluator.
- Reads `DATA_DIR/history/<user_id>/`, runs four pre-committed signal families (frustration, kai_asks_back, preference correction with optional known-fact overlap via Mem0, repeated fact-shape statements), normalizes per 100 user-turns, and buckets around two pinned milestone instants.
- Emits a schema-versioned JSON report and an optional redacted markdown summary suitable for paste-into-issue.
- Trend classifier tiles the `{drop, flat, up}^2` grid into nine named outcomes plus `inconclusive` (volume floor / zero-prior-rate) and a defensive `unclassified` catch-all.
- 111 new tests; full suite at 2320 passing.

## Design notes

- **Determinism contract**: byte-stable JSON for identical inputs. Events sort by `(timestamp_utc, message_id)`, `generated_at_utc` derives from a single wall-clock read in `main()`, and `FrictionEvent.metadata` is `MappingProxyType` -> `dict` only at the render boundary.
- **Memory tri-state**: `MemoryAvailability` is `ENABLED | DISABLED | ERROR`; the harness never `sys.exit`s on a memory failure. Stage 3b (known-fact overlap) silently degrades to "no overlap" when memory is unavailable, and the report records the state.
- **Redaction**: surface text is scrubbed (URLs, emails, @handles, phone-shaped digit runs) and capped at 200 chars; the markdown summary always omits surface text regardless of `--include-snippets`.
- **Traversal guard**: `--user-id` is validated by `_validate_user_id` (rejects empty / `.` / `..` and any value where `Path(user_id).name != user_id`) before any disk read; bad input -> exit code 2 with no I/O. The check is a defensive shape guard, not a hard security boundary - a determined operator running their own harness can still reach arbitrary paths via `--data-dir`.
- **Atomic writes**: temp file + `os.replace()` so a partial write never produces a half-written report at the destination.

## Test plan

- [x] `make check` (ruff + format) clean
- [x] `pytest tests/test_eval_friction.py` 111/111 passed in 0.17s
- [x] Full suite: 2320 passed in 17.82s (above the 2200 acceptance floor)
- [ ] Operator: dry-run against a snapshot copy of production history to spot-check the JSON report shape and markdown summary

fixes #373

### DIFF
--- a/src/kai/eval/friction.py
+++ b/src/kai/eval/friction.py
@@ -1,0 +1,1722 @@
+"""
+Layer 3 longitudinal friction analysis.
+
+Reads an operator's chat history from DATA_DIR/history/<user_id>/, runs
+four pre-committed friction-detection signal families over the ordered
+records, buckets events by milestone (PR #333 and PR #361 merge instants),
+normalizes counts by user-turn volume, and emits a schema-versioned JSON
+report plus an optional redacted markdown summary.
+
+The harness is observational: it never writes to the live history or
+memory stores. It complements Layer 1 (retrieval-only A/B; src/kai/eval/
+retrieval.py) and Layer 2 (behavioral A/B; src/kai/eval/behavioral.py)
+with a real-traffic signal that survives the contrived-probe critique.
+
+Designed to run offline against a SNAPSHOT of DATA_DIR. The typical
+invocation copies history/ and memory/ to /tmp/kai-eval-snapshot,
+points KAI_DATA_DIR there, and invokes `python -m kai.eval.friction`.
+
+This file is intentionally self-contained: no production code path
+imports from it. Outbound dependencies are kai.config (for DATA_DIR
+resolution) and kai.memory (for stage 3b known-fact overlap; optional -
+the harness degrades gracefully when memory is unreachable).
+"""
+
+from __future__ import annotations
+
+import argparse
+import asyncio
+import json
+import logging
+import os
+import re
+import sys
+import tempfile
+from collections.abc import Iterator, Mapping
+from dataclasses import dataclass
+from datetime import UTC, datetime, timedelta
+from enum import Enum
+from pathlib import Path
+from types import MappingProxyType
+from typing import Any
+
+log = logging.getLogger(__name__)
+
+
+# ── Schema version and milestone bucket anchors ──────────────────────
+
+# Bumped when the JSON output shape changes incompatibly. Convention
+# matches `_OUTPUT_SCHEMA_VERSION = 1` at src/kai/eval/behavioral.py:115.
+# Downstream comparison scripts pin against this to detect schema drift
+# without parsing the whole report shape.
+_OUTPUT_SCHEMA_VERSION = 1
+
+# Two milestone anchors carved into module-level constants because they
+# define the entire bucket scheme. They are squash-merge commits on
+# `main` and will not move; treat as authoritative.
+#
+# Original git log timestamps came back in Eastern (-04:00):
+#   commit 5796286 -> 2026-04-17T13:54:27-04:00 -> 2026-04-17T17:54:27Z
+#   commit e508662 -> 2026-04-22T08:43:55-04:00 -> 2026-04-22T12:43:55Z
+# UTC normalization is done here so every downstream comparison is
+# timezone-aware against a single canonical instant. Confirm with
+# `git log --format="%H %aI" -1 <hash>` before changing either anchor.
+_BUCKET_ANCHORS = (
+    datetime(2026, 4, 17, 17, 54, 27, tzinfo=UTC),
+    datetime(2026, 4, 22, 12, 43, 55, tzinfo=UTC),
+)
+
+
+# ── Volume floor and classification thresholds ───────────────────────
+
+# Buckets with fewer user_turns than this floor cannot anchor a confident
+# cross-bucket comparison. The classifier emits
+# `matched_pattern="inconclusive"` when ANY of A/B/C falls below the
+# floor. 30 picked on the same back-of-envelope basis as Layer 1's
+# 26-probe minimum: rate estimates with single-digit denominators have
+# wide enough confidence intervals that label flips are noise-driven.
+_USER_TURN_FLOOR = 30
+
+# "material drop": new-bucket rate is at most this fraction of the prior.
+# The boundary at exactly 0.85x belongs to material drop (the flat band's
+# lower edge is open). Widened from v1's 0.66x to 0.85x in v3 so the
+# bands tile [0, +inf) without an unclassifiable gap.
+_MATERIAL_DROP_RATIO = 0.85
+
+# "up" cutoff: new-bucket rate exceeds this fraction of the prior. The
+# boundary at exactly 1.15x belongs to flat. Used by both the `up` shape
+# in the 9-cell grid and by the `regression` row's B->C condition.
+_UP_RATIO = 1.15
+
+# Pre-committed prediction the harness was built to test. Hardcoded at
+# v1 of the harness; if the prediction itself becomes a tunable
+# parameter, a future revision would expose this as a CLI flag, but
+# tunability via CLI invites post-hoc threshold-fishing so the default
+# stance is "compile-time constant".
+_PREDICTED_PATTERN = "prediction-holds"
+
+
+# ── Signal-family pattern sets ───────────────────────────────────────
+
+# Signal 1: operator-side frustration markers. Substring match on the
+# lowercased user message. Pre-committed: changes to this set require a
+# changelog entry AND a test-suite update so a phrase added after
+# seeing data does not constitute post-hoc pattern-hunting.
+_FRUSTRATION_PHRASES = (
+    "i already told you",
+    "i said earlier",
+    "i just said",
+    "as i mentioned",
+    "as i said",
+    "remember i",
+    "remember that i",
+    "you forgot",
+    "you already",
+    "why are you asking",
+    "why do you keep asking",
+    "i told you",
+)
+
+# Signal 2: stopwords excluded from the content-word set. Filters out
+# common conversational filler so the "3 distinct content-words"
+# threshold reflects substantive overlap rather than incidental shared
+# articles and modal verbs. Frozen so a typo in a test cannot mutate it.
+_STOPWORDS = frozenset(
+    {
+        "this",
+        "that",
+        "with",
+        "from",
+        "have",
+        "will",
+        "what",
+        "when",
+        "where",
+        "which",
+        "would",
+        "could",
+        "should",
+        "about",
+        "there",
+        "their",
+        "your",
+        "they",
+        "them",
+        "been",
+        "were",
+        "than",
+        "then",
+        "into",
+        "some",
+        "such",
+        "more",
+        "most",
+        "also",
+        "does",
+    }
+)
+
+# Signal 2 + stage 3b: the "3 distinct content-words" floor is the
+# false-positive cliff in informal observation - generic clarifications
+# rarely share three or more substantive tokens with prior user content.
+_CONTENT_WORD_FLOOR = 3
+
+# Signal 2: lookback window for prior records. Capping at 20 preceding
+# entries (user + assistant interleaved) keeps the per-record cost
+# bounded; long histories would otherwise scale O(N^2) on the
+# question-answer pair count.
+_KAI_ASKS_BACK_WINDOW = 20
+
+# Signal 3: stage 3a regex set. All compiled with re.IGNORECASE because
+# Telegram traffic routinely capitalizes the first letter of a message
+# via mobile autocorrect ("DON'T", "I Prefer"); without the flag those
+# would silently miss. Word-boundary anchors keep matches honest, e.g.
+# "construction" does not match "instead of" via "instead".
+_CORRECTION_PATTERNS = tuple(
+    re.compile(p, re.IGNORECASE)
+    for p in (
+        r"\bdon'?t\b",
+        r"\bstop (doing|using|asking)\b",
+        r"\bnot that\b",
+        r"\bplease (use|don'?t)\b",
+        r"\balways (use|say|start)\b",
+        r"\bnever (use|say|do)\b",
+        r"\binstead of\b",
+        r"\bi prefer\b",
+    )
+)
+
+# Signal 4: fact-shape prefixes. Compiled into a single alternation so
+# the per-record cost stays linear in record length rather than scaling
+# with the prefix-set size. Word-boundary anchors avoid matching the
+# prefix substring inside an unrelated word.
+_FACT_SHAPE_RE = re.compile(
+    r"\b(?:"
+    r"i am|i'm|my|we use|we're using|"
+    r"i prefer|i always|i use|i don't"
+    r")\b",
+    re.IGNORECASE,
+)
+
+# Signal 4: 5-character shingles + Jaccard similarity threshold. Tuned
+# high enough that paraphrases of the same fact register as a repeat
+# without flagging unrelated user messages that happen to share a few
+# word fragments. Strictly > (not >=) so a synthetic test fixture at
+# exactly 0.4 verifies the open boundary.
+_REPEAT_SHINGLE_K = 5
+_REPEAT_JACCARD_FLOOR = 0.4
+
+# Signal 4: cap the per-record lookback to 200 prior user-turns. Same
+# rationale as Signal 2's window: bound worst-case cost on long
+# histories without losing realistic-repeat detection range.
+_REPEAT_LOOKBACK = 200
+
+
+# ── Surface-text redaction patterns (for --include-snippets) ─────────
+
+# Order matters when patterns overlap. URLs first because they may
+# contain `@` (user:pass URL form); email next because it contains `@`;
+# @handles next because they are pure `@…`; phone last because the loose
+# digit-run pattern would otherwise gobble URL/email substrings. Per-
+# event excerpt cap is enforced separately at _SNIPPET_MAX_CHARS.
+_REDACT_URL_RE = re.compile(r"https?://\S+", re.IGNORECASE)
+_REDACT_EMAIL_RE = re.compile(r"\b[\w.+-]+@[\w.-]+\.[A-Za-z]{2,}\b")
+_REDACT_HANDLE_RE = re.compile(r"@[A-Za-z0-9_]{3,}")
+# Phone: 7+ groups of "digit + optional separator", ending in a digit.
+# Wide enough to cover international and US formats; narrow enough to
+# avoid matching short numeric IDs (e.g. issue #333) which would create
+# false redactions in a chat that frequently references issue numbers.
+_REDACT_PHONE_RE = re.compile(r"(?:\+?\d[\s.-]?){7,}\d")
+_REDACT_REPLACEMENTS = (
+    (_REDACT_URL_RE, "<redacted-url>"),
+    (_REDACT_EMAIL_RE, "<redacted-email>"),
+    (_REDACT_HANDLE_RE, "<redacted-handle>"),
+    (_REDACT_PHONE_RE, "<redacted-phone>"),
+)
+
+# Cap on the surface_text the JSON report may carry per event when
+# --include-snippets is set. 200 chars is plenty to identify the matching
+# message without turning the report into a chat-log copy.
+_SNIPPET_MAX_CHARS = 200
+
+
+# ── Tokenization helper ──────────────────────────────────────────────
+
+# Punctuation stripping uses a simple class-based regex (not the broader
+# unicode category) so it is fast and predictable. Apostrophe is kept
+# inside the character class so contractions like "don't" survive as a
+# single token; downstream stopword filtering does not include "don't"
+# so the token simply does not contribute to the content-word set.
+_PUNCT_RE = re.compile(r"[^\w\s']")
+
+
+def _content_words(text: str) -> set[str]:
+    """Return distinct content-words for overlap comparisons.
+
+    A "content-word" is a token of length >= 4 after lowercasing and
+    stripping punctuation (other than apostrophe), excluding the small
+    `_STOPWORDS` set. Used by Signal 2 (kai_asks_back) and Signal 3
+    stage 3b (known-fact overlap), both of which need a substantive-
+    token overlap floor that is not perturbed by common filler.
+
+    Returns a `set` so callers can intersect via `&`; the floor check
+    is `len(a & b) >= _CONTENT_WORD_FLOOR`.
+    """
+    cleaned = _PUNCT_RE.sub(" ", text.lower())
+    return {t for t in cleaned.split() if len(t) >= 4 and t not in _STOPWORDS}
+
+
+# ── Memory-availability tri-state ────────────────────────────────────
+
+
+class MemoryAvailability(Enum):
+    """Tri-state result of attempting to initialize the memory store.
+
+    DIVERGES from `_initialize_memory` at src/kai/eval/behavioral.py:1901
+    in that the friction harness MUST NOT exit on a non-ENABLED result.
+    A friction analysis without stage 3b is still a valid (degraded)
+    report; the run completes with a `memory-unreachable-frustration-
+    only` caveat instead of failing. Layer 2 cannot run without retrieval
+    so it exits; Layer 3 can.
+
+    Members:
+      ENABLED  - init succeeded and is_enabled() returned True.
+      DISABLED - init succeeded but is_enabled() returned False (the
+                 store is configured off via MEMORY_ENABLED=false or
+                 the embedding-dim validation rejected the model).
+      ERROR    - init raised; the captured exception text is reported
+                 in `memory_store_error_message` for debuggability.
+    """
+
+    ENABLED = "enabled"
+    DISABLED = "disabled"
+    ERROR = "error"
+
+
+@dataclass(frozen=True)
+class MemoryInitResult:
+    """Return shape from `_initialize_memory_local`.
+
+    `error_message` is populated only when `availability == ERROR`.
+    Carrying it on the result rather than logging in the helper lets the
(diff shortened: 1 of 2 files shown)


dcellison:
## Review by Kai

I'll review this PR focusing on bugs, security, error handling, and convention compliance.

## Review: `feat(eval)` Layer 3 friction analysis

Overall the code is well-structured and the test coverage is thorough. A few real issues worth addressing:

---

### Warning: `classify_trend` accepts `memory_state` but never uses it

`friction.py`, `classify_trend` signature:
```python
def classify_trend(
    *,
    ...
    memory_state: MemoryAvailability,   # accepted, never referenced
) -> TrendSummary:
```

The body never reads `memory_state`. The docstring correctly explains why (effect is baked into events at detection time), but a dead parameter still misleads callers and would flag under Pyright strict. Either drop it or add a `# noqa` with a comment.

---

### Warning: `_band_for_transition` is defined but never called

The helper is defined with its own zero-prior handling, but `classify_trend` re-implements the identical logic inline:

```python
# in classify_trend (inline duplicate):
if prior == 0 and new == 0:
    ratios[key] = 1.0
    bands[key] = "flat"
elif prior == 0 and new > 0:
    ratios[key] = None
    bands[key] = "up"
    zero_prior_caveats.append(...)
else:
    ratios[key] = round(new / prior, 3)
    bands[key] = _classify_band(prior, new)
```

`_band_for_transition` is dead code. Either call it or remove it; having both creates a maintenance trap if the band boundaries change.

---

### Suggestion: PR description claims `_USER_ID_RE` that doesn't exist

The description states: "`--user-id` is validated against `_USER_ID_RE` before any disk read." No such constant exists in the diff. The actual guard in `_validate_user_id` uses `Path(user_id).name != user_id`, which is structurally sound but doesn't enforce that the value is numeric (Telegram chat IDs are always integers). If the intent was a regex like `r'^\d+$'`, it's missing. At minimum the PR description is inaccurate.

---

### Suggestion: `_atomic_write` - fd leak edge case

```python
fd, temp_str = tempfile.mkstemp(...)
temp_path = Path(temp_str)
try:
    with os.fdopen(fd, "w", encoding="utf-8") as f:  # if THIS raises, fd leaks
        f.write(content)
```

If `os.fdopen` raises (platform-level issue, extremely rare), `fd` is never closed and the `except` block tries `temp_path.unlink()` but the fd stays open. Wrapping the raw `fd` in `os.close(fd)` inside the except would be defensive:

```python
except Exception:
    try:
        os.close(fd)
    except OSError:
        pass
    ...
```

---

### Clean

- `MappingProxyType` -> `dict` conversion in `_build_event_payload` is correct and the test regression for it is present.
- Bucket boundary semantics (`<=` on drop, `<=` on flat's upper bound) match the docstring and band descriptions.
- Determinism contract (sort by `(timestamp_utc, message_id)`, single `datetime.now` call) is correctly implemented.
- `_atomic_write` temp-in-same-dir + `os.replace` is correct.
- Memory tri-state degradation path is clean; the harness never calls `sys.exit` on memory failure.
- The `_validate_user_id` double-call (in `_run_cli` and again in `read_history`) is fine as defense in depth.

**Summary:** No critical bugs. Two warnings worth fixing before merge (dead parameter in `classify_trend`, dead `_band_for_transition` helper), and the description's `_USER_ID_RE` claim should be corrected.

dcellison:
## Review by Kai

Looking at this PR with the context of what the prior review already covered. The three warnings from the prior round (dead `memory_state` parameter, dead `_band_for_transition`, `_atomic_write` fd leak) all appear to be addressed in this diff - the parameter is gone from `classify_trend`, the helper function doesn't appear anywhere in the file, and the fd cleanup was added to the except block.

New findings:

---

**Warning: Bare `list` type annotations violate Pyright strict**

Three call sites use `list` without a type parameter, which generates `reportMissingTypeArgument` under Pyright strict - the convention explicitly cited for this project:

- `friction.py`: `_detect_preference_correction(..., facts: list)`
- `friction.py`: `detect_events(..., facts: list)`
- `friction.py`: `_load_facts_for_user(...) -> list:`

`from typing import Any` is already imported. All three should be `list[Any]`. The `getattr` access pattern in the detector body means nothing further needs to change - the fix is purely at the annotation boundary.

---

**Suggestion: `shingle_cache` in `_detect_repeated_fact` grows without eviction**

`friction.py`, `_detect_repeated_fact`:

```python
shingle_cache: dict[int, set[str]] = {}
# ...
shingle_cache[idx] = _shingles(record.text)
# priors beyond _REPEAT_LOOKBACK are never accessed again but stay in the dict
```

The `_REPEAT_LOOKBACK = 200` cap bounds which priors get compared, but the cache retains entries for the entire history. For a single-user production run this is bounded by the number of fact-shape records in the full history - likely fine in practice - but the bound isn't enforced. An eviction keyed on `user_indices[max(0, pos - _REPEAT_LOOKBACK)]` would make the bound explicit.

---

**Suggestion: `_atomic_write` - double-close on `os.replace` failure is silently swallowed**

When `os.fdopen` and `f.write` both succeed but `os.replace` raises, the `with` block has already closed `fd` via `f.__exit__`, then the except block calls `os.close(fd)` on the already-closed descriptor, which raises `EBADF` caught by `except OSError: pass`. The temp file is then correctly unlinked. The behavior is correct; the comment "in case os.fdopen itself raised" partially explains this but doesn't mention the no-op-double-close case, which a future reader might mistake for a missing early-exit guard.

---

Everything else looks solid: the determinism contract is correctly implemented (single `datetime.now` call, sort by `(timestamp_utc, message_id)`), the `MappingProxyType` -> `dict` conversion in `_build_event_payload` is present and tested, memory tri-state degradation is clean, and the 9-cell grid is exhaustive with a correct `unclassified` catch-all.

dcellison:
## Review by Kai

Looking at this PR with both prior review rounds in mind. Prior issues (dead `memory_state` param, dead `_band_for_transition`, `_atomic_write` fd leak, bare `list` annotations, `shingle_cache` eviction, double-close comment) all appear resolved in the current diff.

New findings only:

---

### Suggestion: `read_history` - `chat_id` cast can raise an uncaught `ValueError`

`friction.py`, inside the `for line_no, line in enumerate(...)` loop:

```python
chat_id=int(row.get("chat_id") or 0),
```

The `json.JSONDecodeError` try/except above this line does not cover `int(...)`. A valid-JSON line with `"chat_id": "corrupted"` passes the parse guard but raises `ValueError` here, which propagates uncaught through the generator and crashes `list(read_history(...))` in `_run_cli`. The docstring explicitly promises "Malformed lines do NOT abort the run," so this is an inconsistency. A safe cast (`int(row.get("chat_id") or 0)` inside a try/except, or a helper) would close the gap. In practice, production JSONL always has integer `chat_id`, so this is low-probability but worth noting against the stated contract.

---

### Suggestion: `test_window_excludes_records_beyond_lookback` - assertion is weaker than the stated intent

`tests/test_eval_friction.py`, `TestKaiAsksBackSignal`:

```python
assert events == [] or all("outside.jsonl" not in mid for ev in events for mid in ev.context_message_ids)
```

The setup constructs noise records with text `"unrelated noise N"`, whose content-word set (`{"unrelated", "noise"}`) has zero overlap with the assistant question's content-word set (`{"database", "deployment", "container", "postgres"}`). So `events == []` is the guaranteed and correct outcome. The disjunctive form passes even if the window check were broken as long as the specific "outside.jsonl" record isn't cited, which masks a future regression in the lookback boundary. Assert `events == []` directly.

---

Everything else looks solid. The determinism contract (single `datetime.now` call, sort by `(timestamp_utc, message_id)`), `MappingProxyType` -> `dict` conversion in `_build_event_payload`, memory tri-state degradation, `_validate_user_id` shape guard, and the 9-cell grid exhaustiveness with the `unclassified` catch-all are all correctly implemented.

dcellison:
## Review by Kai

All three prior review rounds have been addressed: dead `memory_state` parameter removed, `_band_for_transition` dead code removed, `_atomic_write` fd-leak fixed with a comment explaining the double-close case, bare `list` annotations parameterized, `shingle_cache` eviction added, `chat_id` cast wrapped in try/except, and the weak lookback-boundary assertion replaced with a direct `assert events == []`.

One new finding:

---

**Suggestion: `test_window_excludes_records_beyond_lookback` - comment arithmetic is wrong**

`tests/test_eval_friction.py`, `TestKaiAsksBackSignal`:

```python
# The single matching prior is at position 21; the assistant
# question lands at position 42 (21 + 21 fillers between them),
# so the matching prior is exactly one slot OUTSIDE the 20-record
# lookback window.
```

The count is off by two. After the insert, the list is:

- indices 0-20: 21 noise records
- index 21: matching prior (`outside.jsonl`)
- indices 22-42: 21 more noise records
- index 43: assistant question

The assistant lands at index 43 (not 42), and the matching prior is 22 slots back (`43 - 21 = 22`), placing it 2 slots outside the window (window starts at index 23, prior is at 21). The comment says "exactly one slot OUTSIDE" and "position 42" - both are wrong. The assertion `events == []` is correct and the test behavior is right; only the inline explanation is inaccurate.

---

Everything else looks solid: the determinism contract (single `datetime.now` call, stable `(timestamp_utc, message_id)` sort key), `MappingProxyType -> dict` conversion in `_build_event_payload`, memory tri-state degradation, `_validate_user_id` shape guard, and the 9-cell grid exhaustiveness with the `unclassified` catch-all are all correctly implemented and tested.

dcellison:
## Review by Kai

All four prior review rounds appear fully resolved: dead `memory_state` parameter removed, `_band_for_transition` dead code gone, `_atomic_write` fd-leak comment added, bare `list` annotations parameterized, `shingle_cache` eviction in place, `chat_id` cast guarded, lookback-boundary assertion direct, and the comment arithmetic corrected to "index 43 / 2 slots outside."

One new finding in the visible portion of the diff:

---

**Suggestion: `_make_aggregates` - `count` field is systematically wrong for sub-unit rates**

`tests/test_eval_friction.py`, `_make_aggregates` helper (around line 185 in the test file):

```python
out.append(
    BucketAggregate(
        bucket_label=bucket,
        signal_family=family,
        count=int(rate),          # truncates fractional rates
        user_turns_in_bucket=100,
        rate_per_100_user_turns=rate,
    )
)
```

With `user_turns_in_bucket=100`, the count implied by a rate of `R` is exactly `R` events. `int(R)` truncates: `rate=0.5` gives `count=0`, `rate=2.7` gives `count=2`. Any test that exercises the full pipeline through `_make_aggregates` and then asserts on `report["aggregates"][i]["count"]` will see a wrong value for non-integer rates. `classify_trend` reads only `rate_per_100_user_turns` so classifier-focused tests are unaffected, but JSON-output tests would silently pin incorrect counts. A consistent fix is `count=round(rate)` or `count=int(round(rate))`.

---

**Note:** The diff is truncated; the remaining ~1,200 lines of the test file are not visible, so findings there may exist that this review cannot surface.

Everything else visible looks solid: the determinism contract (single `datetime.now` call, `(timestamp_utc, message_id)` sort key), `MappingProxyType -> dict` conversion in `_build_event_payload`, memory tri-state degradation, `_validate_user_id` shape guard, the 9-cell grid with the `unclassified` catch-all, and the shingle-cache eviction bound are all correctly implemented.

dcellison:
## Review by Kai

All five prior rounds of findings are addressed in the current diff. The `memory_state` parameter is gone, `_band_for_transition` is gone, the `_atomic_write` fd comment is present, `list` annotations are parameterized, `shingle_cache` eviction is in place, `chat_id` is guarded, the lookback assertion is direct, the comment arithmetic reads "index 43 / 2 slots outside," and `_make_aggregates` uses `count=round(rate)`.

One new finding:

---

**Suggestion: `_detect_repeated_fact` - shingle_cache eviction comment overstates the memory bound**

`friction.py`, `_detect_repeated_fact`:

```python
for pos, idx in enumerate(user_indices):
    record = records[idx]
    if not _FACT_SHAPE_RE.search(record.text):
        continue          # <- eviction code is below this, unreachable for non-fact-shape records
    shingle_cache[idx] = _shingles(record.text)
    evict_pos = pos - _REPEAT_LOOKBACK - 1
    if evict_pos >= 0:
        shingle_cache.pop(user_indices[evict_pos], None)
```

The eviction fires only when a fact-shape record is processed. For histories where fact-shape records are sparse, `user_indices[evict_pos]` is almost always a non-fact-shape record and therefore not in the cache, making the `pop` a no-op. The cache then grows to O(K) where K is the count of fact-shape records in the full history, not O(`_REPEAT_LOOKBACK`) as the block comment claims ("peak memory is O(_REPEAT_LOOKBACK) entries regardless of total history length"). Correctness is unaffected - the inner comparison loop is bounded by `start = max(0, pos - _REPEAT_LOOKBACK)` and stale cache entries are never accessed. The fix is either to move the eviction before the `continue` check (runs for every user record, evicting the one entry per step), or to update the comment to accurately say "bounded by the number of fact-shape records in the full history, likely small in practice."

---

Everything else looks solid: the determinism contract (single `datetime.now(UTC)` call, `(timestamp_utc, message_id)` sort key), `MappingProxyType -> dict` conversion in `_build_event_payload`, memory tri-state degradation, `_validate_user_id` path-shape guard, and the exhaustive 9-cell grid with the `unclassified` catch-all are all correctly implemented and tested. The diff was truncated; findings in the remaining ~900 lines of the test file are not visible.

dcellison:
## Review by Kai

All prior-round findings (dead `memory_state` param, `_band_for_transition` dead code, `_atomic_write` fd comment, bare `list` annotations, `shingle_cache` eviction, `chat_id` cast guard, direct lookback assertion, comment arithmetic, `_make_aggregates` `round(rate)`, eviction-before-`continue` placement) are resolved in the current diff. One new finding:

---

**Warning: Bare `dict` type annotations violate Pyright strict**

The same `reportMissingTypeArgument` class of issue that Round 2 caught for `list` is present for `dict` in six places, all introduced in this PR:

- `friction.py`, `HistoryRecord.media: dict | None` - should be `dict[str, Any] | None`
- `friction.py`, `_build_event_payload(...) -> dict` - should be `-> dict[str, Any]`
- `friction.py`, `_build_bucket_payload(...) -> list[dict]` - should be `-> list[dict[str, Any]]`
- `friction.py`, `_render_json(report: dict)` - should be `dict[str, Any]`
- `friction.py`, `_render_markdown(report: dict, ...)` - should be `dict[str, Any]`
- `friction.py`, `_build_report(...) -> dict` - should be `-> dict[str, Any]`

`from typing import Any` is already imported. The fix is purely at the annotation boundary; none of the function bodies need to change.

---

**Clean**

- `_detect_repeated_fact` eviction now runs before the `continue` check; the O(_REPEAT_LOOKBACK) comment is accurate.
- `_atomic_write` comment correctly explains the double-close no-op case.
- `classify_trend` zero-prior handling is correct (0/0 → flat, 0/positive → `up` + inconclusive).
- `_classify_band` bands tile `[0, ∞)` without a gap for `prior > 0`.
- `rate_by_cell` key access in `_render_markdown` is safe because `aggregate()` always initializes the full 12-cell matrix.
- `_json_default` Mapping fallback is a correct defensive net; no nested `MappingProxyType` in practice.
- `test_window_excludes_records_beyond_lookback` comment ("index 43 / 2 slots outside") and direct `assert events == []` are both correct.

dcellison:
## Review by Kai

All seven prior review rounds appear fully resolved. I've gone through the current diff carefully for new findings.

---

### Suggestion: `bands` local variable annotation overstates possible values

`friction.py`, `classify_trend`:

```python
bands: dict[str, str | None] = {}
```

The `| None` is never exercised. Every branch of the loop assigns either `"flat"`, `"up"`, or the return of `_classify_band` (one of `"drop"`, `"flat"`, `"up"`). The annotation misleads readers into thinking `None` is a reachable value, and it's what motivates the defensive `trend.bands.get("A_to_B") or ""` guard in `_render_markdown` (which is also a no-op in practice, since both keys are always set). `dict[str, str]` would be the accurate annotation; the or-guard in `_render_markdown` could then become a plain index access with no `or ""`.

---

### Clean

Everything from all seven prior rounds is confirmed resolved:

- No bare `list` or `dict` annotations anywhere in the file.
- `_detect_repeated_fact` shingle eviction fires before the `continue` check; the O(`_REPEAT_LOOKBACK`) bound comment is accurate.
- `_atomic_write` fd comment correctly describes the EBADF no-op case.
- `chat_id` cast wrapped in `try/except (TypeError, ValueError)`.
- `_make_aggregates` uses `count=round(rate)`.
- `test_window_excludes_records_beyond_lookback` uses direct `assert events == []` with accurate "index 43 / 2 slots outside" comment.
- Determinism contract (single `datetime.now(UTC)` call, `(timestamp_utc, message_id)` sort key) is correctly implemented.
- `MappingProxyType -> dict` conversion in `_build_event_payload` is present and tested.
- Memory tri-state degradation is clean; harness never calls `sys.exit` on memory failure.
- `_validate_user_id` path-shape guard is applied before any disk read.
- 9-cell grid is exhaustive with `unclassified` catch-all.
- `rate_by_cell` key access in `_render_markdown` is safe because `aggregate()` always initializes the full 12-cell matrix.

No critical issues or warnings. The single suggestion above is a type-annotation cosmetic; the runtime behavior is correct.

dcellison:
## Review by Kai

All eight prior-round findings are resolved in the current diff. After a full read of the visible portion of both files, I have one new finding and a confirmation of the one open Round 8 suggestion.

---

### Suggestion: `_detect_kai_asks_back` - early exit on `own_words < floor` masks the window invariant from tests

`friction.py`, `_detect_kai_asks_back`:

```python
own_words = _content_words(record.text)
if len(own_words) < _CONTENT_WORD_FLOOR:
    continue
window = records[max(0, i - _KAI_ASKS_BACK_WINDOW) : i]
```

The early-exit is correct and documented, but the `test_window_excludes_records_beyond_lookback` test constructs an assistant question with four content-words (`database`, `deployment`, `container`, `postgres`) that comfortably clears the floor. That's fine. However, the test's noise records use `"unrelated noise N"` - only `"unrelated"` and `"noise"` survive the length-4 filter; both are below the 3-word floor when intersected with the assistant's word set. The test therefore validates correct window boundaries only for the matching prior, not for the noise records. This is correct but a silent dependency: if someone later changes the noise text to include substantive overlap words, `events == []` would pass for the wrong reason (the matching prior is out of window, but a noise record inside the window might also fire). The test is correct as written; this note is purely for future maintainability.

---

### Open (Round 8, author's decision): `bands: dict[str, str | None]`

`friction.py`, `classify_trend` local variable and `TrendSummary.bands` dataclass field. The `| None` is never instantiated; both should be `dict[str, str]`. The author has seen this.

---

### Clean

Everything else is solid:

- Determinism contract: single `datetime.now(UTC)` call in `main()`, `(timestamp_utc, message_id)` sort key in `_build_report`. ✓
- `_detect_repeated_fact` eviction fires before the `continue` check; the O(`_REPEAT_LOOKBACK`) comment is accurate. ✓
- `_atomic_write` comment correctly explains the EBADF no-op double-close case. ✓
- `chat_id` cast guarded in `try/except (TypeError, ValueError)`. ✓
- `_make_aggregates` uses `count=round(rate)`. ✓
- Bare `list[Any]` and `dict[str, Any]` annotations throughout. ✓
- `MappingProxyType -> dict` conversion in `_build_event_payload` present and tested. ✓
- Memory tri-state degradation; harness never calls `sys.exit` on memory failure. ✓
- `_validate_user_id` applied before any disk I/O; double-call is defense in depth. ✓
- 9-cell grid exhaustive; `unclassified` catch-all present. ✓
- `test_window_excludes_records_beyond_lookback` comment ("index 43 / 2 slots outside") is arithmetically correct. ✓
- `_build_caveats` `ratios_undefined` path and zero-prior handling in `classify_trend` are consistent. ✓

The diff is truncated at ~590 of 1580 test lines; findings in the remaining portion are not visible to this review.

PR #379, opened 2026-04-24
feat(config): add CLAUDE_EFFORT_LEVEL for inner Claude --effort flag

### LINKED ISSUE #378: Add EFFORT_LEVEL config var to control inner Claude --effort flag
## Problem

Inner Claude (the `claude` subprocess Kai launches per chat) currently has no operator-controlled way to set the `--effort` CLI flag. Verified via `claude --help`: the flag accepts `low | medium | high | xhigh | max` and controls how many reasoning tokens Claude spends per turn.

Right now every inner-Claude invocation runs at whatever the binary picks as its default effort, with no way to dial it up for harder work or down to save budget. As cost vs. quality tradeoffs evolve we have no knob to turn.

Outer Claude exposes the same control via the operator's global Claude Code settings.json. There is no equivalent for inner Claude.

## Proposed solution

Add `EFFORT_LEVEL` as a global config var following the existing "Adding a New Env Var" pattern documented in `.claude/CLAUDE.md`:

1. New field `effort_level: str = "high"` on the main `Config` dataclass in `src/kai/config.py`, placed near `memory_extraction_model` since it is similar in shape (single global string with allow-list validation, not per-user).
2. Validated parsing in `load_config()` against the allow-list `{low, medium, high, xhigh, max}`. Invalid values raise `SystemExit` with a clear message naming the env var, mirroring the `DEFAULT_MODEL` validation path.
3. Threaded through `src/kai/pool.py` to `src/kai/claude.py.__init__` (both `ClaudeProcess` instantiation sites in `pool.py` must be updated; the constructor gains an `effort_level` kwarg).
4. Appended to the inner-claude `claude_cmd` list as `--effort <value>`. Important: this must go on the CLI, NOT into the `--settings` JSON blob. settings.json silently drops unknown keys, so a typo in a settings-key path produces a no-op; an invalid CLI flag fails loudly at subprocess startup.
5. Wizard prompt added to `_cmd_config()` in `src/kai/install.py` using the existing `_prompt_choice` helper, exposing the five valid values. Written to env only if non-default, matching the prevailing "delta from default" convention so `install.conf` stays minimal.
6. Documented in `.env.example` near `DEFAULT_MODEL` (both shape inner Claude behavior).
7. Added to `_CONFIG_ENV_VARS` in `tests/test_config.py` plus a new `TestEffortLevel` class covering: default applied when unset, valid value parses, invalid value raises `SystemExit`, uppercase normalized via `.lower()`, surrounding whitespace stripped.
8. New test in `tests/test_claude.py` (or wherever `claude_cmd` is currently tested) asserting `--effort <value>` appears as adjacent elements in the constructed command list.

## Default value

Proposed: `high`. Rationale: the operator's existing outer Claude default is `high`, and matching it keeps inner Claude behavior continuous on first deploy. Choosing `medium` as the default would silently downgrade existing reasoning quality, which is a behavior change disguised as a config addition.

## Out of scope

- Per-user `effort_level` via `users.yaml`. The plumbing for per-user values already exists for `model` and `budget`; promoting `effort_level` to per-user later is a parallel change and should be its own issue.
- Per-workspace overrides. Same reasoning. Add when there is a real use case.
- Any change to the `--settings` JSON path Kai already passes to inner Claude. Out of scope here even though that path has a related, separate bug worth a follow-up issue.

## Acceptance criteria

- [ ] `EFFORT_LEVEL` env var is accepted, validated, and reflected in `Config.effort_level`
- [ ] Inner Claude subprocess receives `--effort <value>` in its command line for every invocation
- [ ] Invalid values raise `SystemExit` at config-load time with a message naming the env var
- [ ] `make config` prompts for the value with the five valid choices and persists it to `install.conf` only when non-default
- [ ] `.env.example` documents the variable with allowed values and interaction notes
- [ ] Test coverage in `tests/test_config.py` (parsing, validation, normalization) and `tests/test_claude.py` (flag appears in command)
- [ ] No silent fallback when the env var is unparseable; the operator must see the failure

dcellison:
## Summary

- Adds `CLAUDE_EFFORT_LEVEL` global env var (and matching `claude_effort_level` Config field) that threads through to the inner Claude subprocess as `--effort <value>`. Allowed values are `low|medium|high|xhigh|max`, validated at config load against an allow-list pulled verbatim from `claude --help`.
- Default value is `high`, matching the operator's outer-Claude default. Critical when `CLAUDE_USER` or per-user `os_user` is set: inner Claude runs as a different OS user and does NOT inherit the operator's settings.json effort default, so without an explicit CLI value user-isolated installs would silently fall to whatever the binary picks as its own default.
- Wizard (`make config`) prompts for the value after the autocompact threshold prompt and emits the env var only when non-default, matching the existing "delta from default" pattern in `install.conf`.
- Naming refinement vs. issue: issue body proposed `EFFORT_LEVEL`; the implemented name is `CLAUDE_EFFORT_LEVEL` to match the established `CLAUDE_*` env-var cluster for inner-Claude knobs (`CLAUDE_TIMEOUT_SECONDS`, `CLAUDE_MAX_SESSION_HOURS`, `CLAUDE_MAX_CONTEXT_WINDOW`, `CLAUDE_AUTOCOMPACT_PCT`). Recorded in an issue comment before this PR.

Fixes #378

## Test plan

- [ ] `make test` (full suite, 2382 tests) — local: passes
- [ ] `TestClaudeEffortLevel` (6 new tests in `tests/test_config.py`): defaults, valid value parsing, invalid value rejection, uppercase normalization, whitespace stripping, whitespace-only falls back to default (the last added in round-1 review to cover the `or "high"` fallback branch directly)
- [ ] `test_effort_flag_default_in_cmd` and `test_effort_flag_custom_value_in_cmd` (2 new tests in `tests/test_claude.py`): assert `--effort <value>` appears adjacently in the constructed subprocess command
- [ ] `test_effort_level_default_omits_env_key` and `test_effort_level_non_default_writes_env_key` (2 new tests in `tests/test_install.py` from round-1 review): pin both sides of the install.conf delta-from-default emission branch
- [ ] Existing `tests/test_install.py` wizard sequences (8 sites) updated to include the new prompt; all pass
- [ ] Verify `make config` walk-through prompts for the new value and persists `CLAUDE_EFFORT_LEVEL` to `install.conf` only when non-default
- [ ] Spot-check that a running install with `CLAUDE_EFFORT_LEVEL=xhigh` in env produces `--effort xhigh` in the inner-Claude subprocess command line (e.g. via `ps -ef | grep claude`)

## Out of scope

- Per-user `claude_effort_level` via `users.yaml` — separate follow-up issue if/when there is a use case
- Per-workspace overrides via `WorkspaceConfig` — same reasoning
- The unrelated `--settings` JSON `maxContextWindow` no-op bug in `claude.py` — separate follow-up issue
- The Haiku memory-extraction subprocess — its effort is controlled implicitly by the Haiku model choice and per-extraction budget envelope

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -44,6 +44,13 @@ TELEGRAM_BOT_TOKEN=your-token-from-botfather
 # Per-user defaults are set via 'max_budget' in users.yaml.
 #BUDGET_CEILING=10.0
 
+# Effort level for inner Claude reasoning (low|medium|high|xhigh|max).
+# Higher = more thinking tokens per turn = better answers + higher
+# token spend. Interacts with BUDGET_CEILING: high effort spends tokens
+# faster, so pair with a higher BUDGET_CEILING for long sessions.
+# Truly global; applies to every chat's inner Claude subprocess.
+#CLAUDE_EFFORT_LEVEL=high
+
 # Maximum session age in hours before recycling (0 = no limit).
 # Prevents unbounded memory growth in the inner Claude process.
 # Recommended: 4-8 hours for memory-constrained machines.
--- a/src/kai/claude.py
+++ b/src/kai/claude.py
@@ -94,6 +94,11 @@ class ClaudeCodeBackend(AgentBackend):
         workspace_config: WorkspaceConfig | None = None,
         max_context_window: int = 0,
         autocompact_pct: int = 0,
+        # Default duplicated from Config.claude_effort_level (config.py).
+        # Production calls (pool.py) always pass an explicit value from
+        # config so the duplication is invisible there; the default here
+        # exists so direct ClaudeCodeBackend(...) instantiation in tests
+        # does not have to plumb an effort kwarg. Keep the two in sync.
         claude_effort_level: str = "high",
     ):
         # ABC-required attributes (pool.py reads/writes these)
--- a/src/kai/claude.py
+++ b/src/kai/claude.py
@@ -94,6 +94,7 @@ class ClaudeCodeBackend(AgentBackend):
         workspace_config: WorkspaceConfig | None = None,
         max_context_window: int = 0,
         autocompact_pct: int = 0,
+        claude_effort_level: str = "high",
     ):
         # ABC-required attributes (pool.py reads/writes these)
         self.model = model
@@ -103,6 +104,15 @@ class ClaudeCodeBackend(AgentBackend):
         self.timeout_seconds = timeout_seconds
         self.workspace_config = workspace_config
         self.max_context_window = max_context_window
+        # Effort level for the inner Claude --effort flag. Stored on the
+        # instance so the value is fixed for the life of the subprocess
+        # rather than re-read from config on every claude_cmd build, and
+        # so subclasses / tests can override it without monkey-patching
+        # config. No _default_claude_effort_level shadow because there
+        # is no workspace-level override path that would need to restore
+        # a default after a /workspace switch (see workspace_config block
+        # below for the pattern that DOES need a default shadow).
+        self.claude_effort_level = claude_effort_level
         self.provider = "anthropic"  # Claude CLI always uses Anthropic
 
         # Claude-Code-specific attributes (not on the ABC)
@@ -191,6 +201,19 @@ class ClaudeCodeBackend(AgentBackend):
             "--verbose",
             "--model",
             self.model,
+            # Effort level controls how many reasoning tokens Claude spends
+            # per turn. Passed as a CLI flag (verified against `claude --help`)
+            # rather than via the --settings JSON path below, because CLI
+            # flags fail loudly on typo at subprocess startup while the
+            # --settings JSON path is reserved for keys that have no
+            # dedicated CLI entry. Future maintainer: do NOT migrate this
+            # into the --settings JSON without first re-verifying the key
+            # name against `claude --help`. The value is validated at
+            # config load (config.py _VALID_EFFORT_LEVELS) so by the time
+            # it reaches this point it is guaranteed to be one of the
+            # five accepted strings.
+            "--effort",
+            self.claude_effort_level,
             "--permission-mode",
             "bypassPermissions",
             "--max-budget-usd",
--- a/src/kai/config.py
+++ b/src/kai/config.py
@@ -120,14 +120,21 @@ def get_effective_provider(backend: str, llm_provider: str) -> str:
 
 # Valid values for the inner Claude --effort flag, taken verbatim from
 # `claude --help` output. If Anthropic adds a new tier (e.g. "ultra"),
-# this set must be updated here or otherwise-valid configs will be
-# rejected at config-load time. Pinned as a module-level frozenset so
-# the membership check is O(1) and the values are not re-allocated on
-# every config load. Validation against this set keeps an invalid
-# CLAUDE_EFFORT_LEVEL from reaching the inner-Claude subprocess: a
-# bad value would otherwise cost a full session to discover (the
-# subprocess would fail at startup, not at config load).
-_VALID_EFFORT_LEVELS: frozenset[str] = frozenset({"low", "medium", "high", "xhigh", "max"})
+# update EFFORT_LEVELS here or otherwise-valid configs will be rejected
+# at config-load time. Two shapes intentionally:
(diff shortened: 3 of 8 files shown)


dcellison:
## Review by Kai

This PR is well-structured and follows the project conventions closely. A few findings:

---

**Warning**

**`config.py` - error message uses alphabetical sort, not logical order**

```python
f"CLAUDE_EFFORT_LEVEL must be one of {sorted(_VALID_EFFORT_LEVELS)}, got {claude_effort_level!r}"
```

`sorted()` on the frozenset gives `['high', 'low', 'max', 'medium', 'xhigh']` - alphabetical, not the effort progression an operator expects. Replace with a fixed tuple `("low", "medium", "high", "xhigh", "max")` in the error message (and optionally in `_VALID_EFFORT_LEVELS` itself as a comment). Minor but this is an operator-facing error string.

---

**Suggestion**

**`tests/test_config.py` - no test for the `or "high"` fallback path**

The parser has `os.environ.get("CLAUDE_EFFORT_LEVEL", "high").strip().lower() or "high"`. The `or "high"` branch fires only when the env var is set to whitespace-only (e.g. `"   "`). `test_whitespace_stripped` uses `"  medium  "` which still resolves to a valid value - it doesn't hit the `or "high"` path. A test setting `CLAUDE_EFFORT_LEVEL="   "` would close this gap.

**`tests/test_install.py` - no test for non-default value persisting to install.conf**

All 8 updated fixtures use `""` (accept default). There's no test that passes `"xhigh"` and asserts `CLAUDE_EFFORT_LEVEL=xhigh` appears in the written `install.conf`. The existing delta-from-default pattern (`if claude_effort_level != "high": env[...] = ...`) goes untested as a positive case.

**`claude.py` - duplicate default creates silent drift risk**

`claude_effort_level: str = "high"` in `__init__` duplicates the `Config` default. If someone changes one without the other, `pool.py` always passes the explicit value so it won't matter in production - but a `_make_claude()` call in tests without the kwarg would silently test the wrong default. Low risk given current test coverage, but worth documenting with a one-liner pointing to `Config.claude_effort_level`.

---

**Clean**

- Config/pool/claude wiring is correct and follows the established `CLAUDE_*` passthrough pattern exactly.
- `load_config()` parsing is sound: strip + lower + or-default + membership check is the right approach for a string allow-list.
- All 8 wizard test sequences updated correctly.
- `_CONFIG_ENV_VARS` registration is present.
- `.env.example` documentation is clear and accurate.

dcellison:
## Review by Kai

Good. `_prompt_choice` returns the default as-is on empty input (`if not value and default: return default`), so an empty Enter press returns `"high"` and the emit condition `!= "high"` correctly suppresses the key. The wiring is correct end-to-end.

---

**This PR is clean. All three prior-review findings are resolved.**

Here is the accounting:

**Resolved from prior review**

- **Warning (sort order)**: Fixed. Error message now uses `list(EFFORT_LEVELS)` which preserves the intensity progression `['low', 'medium', 'high', 'xhigh', 'max']`.
- **Suggestion (whitespace-only test)**: Fixed. `test_whitespace_only_falls_back_to_default` exercises `"   "` and asserts the result is `"high"`, directly covering the `or "high"` branch.
- **Suggestion (non-default install.conf test)**: Fixed. `test_effort_level_non_default_writes_env_key` passes `effort="xhigh"` and asserts `conf["env"]["CLAUDE_EFFORT_LEVEL"] == "xhigh"`.
- **Suggestion (duplicate default comment)**: Fixed. The constructor kwarg now has a comment pointing to `Config.claude_effort_level` and explaining why the duplication is safe.

**New findings**

One minor note only:

**Suggestion - PR description test count is off**

The description says "5 new tests in `tests/test_config.py`" but the diff adds 6 (`test_whitespace_only_falls_back_to_default` was added to address the prior review). This is a description inaccuracy, not a code issue.

**Everything else is correct.** Config/pool/claude wiring follows the established `CLAUDE_*` passthrough pattern. Validation order (strip, lower, or-default, membership check) is sound. `_prompt_choice` normalizes user input to lowercase before the membership check, so the wizard accepts `"XHIGH"` from the operator correctly. The emit-only-on-delta pattern in `install.py` is consistent with `autocompact_pct`. `_CONFIG_ENV_VARS` registration is present. `.env.example` documentation is accurate.

PR #381, opened 2026-04-24
fix(install): gate Claude-only wizard prompts on agent_backend

### LINKED ISSUE #380: Gate Claude-only wizard prompts on agent_backend=claude
## Problem

The `make config` wizard prompts the operator for several Claude-binary-specific knobs unconditionally, even when the operator has selected `goose` as the agent backend. Goose has no equivalent of these settings, so the prompts at best confuse the operator and at worst cause them to set values that are silently ignored at runtime.

Verified via audit of `src/kai/install.py` and `src/kai/pool.py`:

| Env var | Wizard line | Currently gated? | Backend(s) that consume it |
|---|---|---|---|
| `CLAUDE_AUTOCOMPACT_PCT` | `install.py:546-557` | No | Claude only (`ClaudeCodeBackend` autocompact env override) |
| `CLAUDE_EFFORT_LEVEL` | `install.py:559-573` | No | Claude only (`ClaudeCodeBackend` `--effort` flag) |
| `CLAUDE_USER` | `install.py:388-394` | Only by `if advanced:` | Claude only (sudo isolation; Goose has no sudo path) |

The other `CLAUDE_*` env vars in the wizard (`CLAUDE_TIMEOUT_SECONDS`, `CLAUDE_MAX_CONTEXT_WINDOW`, `CLAUDE_MAX_BUDGET_USD` -> `BUDGET_CEILING`, `DEFAULT_MODEL` <- `CLAUDE_MODEL`) are correctly **not** gated on backend — both `GooseBackend` and `ClaudeCodeBackend` accept the corresponding kwargs in `pool.py`.

## Proposed solution

Add a backend-conditional gate to each of the three prompts identified above. The pattern that works:

1. Initialize the variable to its dataclass default before the conditional, so the env-emission `if value != default` block downstream still works without raising `NameError` when goose is selected.
2. Wrap the `_prompt(...)` / `_prompt_choice(...)` call in `if agent_backend == "claude":`.
3. The matching env-emission block in the lower half of `_cmd_config()` already short-circuits on the default-value check (`if claude_effort_level != "high"`), so no separate emission gate is needed - but verify each one to be sure.

For `CLAUDE_USER`: keep the existing `if advanced:` gate AND add `agent_backend == "claude"` as an additional condition. The `os_user` field has no meaning for Goose.

## Tests

Every fixture in `tests/test_install.py` currently passes `"claude"` as the agent backend. This audit needs at least one `"goose"` fixture per gated prompt that asserts the prompt is skipped (otherwise the gating regresses silently the next time someone touches the wizard). The simplest path is to extend `_base_inputs` with an `agent_backend` parameter (default `"claude"`) and add a single new `TestCmdConfig::test_goose_backend_skips_claude_only_prompts` test that:

- Calls `_base_inputs(memory_block=["false"], agent_backend="goose")` (omitting the autocompact, effort, and CLAUDE_USER inputs)
- Runs `_cmd_config()`
- Asserts the resulting `install.conf` env contains none of the three keys

## Out of scope

- `CLAUDE_MAX_SESSION_HOURS` and `CLAUDE_IDLE_TIMEOUT`: also Claude-only at the consumer (`ClaudeCodeBackend`-only kwargs in `pool.py`), but neither has a wizard prompt today (operators hand-edit env). Adding wizard prompts AND gating them is a separate piece of work; out of scope here.
- Re-validation that already-set CLAUDE_* values are *removed* from `install.conf` when an operator switches an existing config from `claude` to `goose`. Today the wizard preserves prior values across backend changes; whether that should change is a separate question.

## Acceptance criteria

- [ ] Wizard does not prompt for `CLAUDE_AUTOCOMPACT_PCT` when `agent_backend == "goose"`
- [ ] Wizard does not prompt for `CLAUDE_EFFORT_LEVEL` when `agent_backend == "goose"`
- [ ] Wizard does not prompt for `CLAUDE_USER` when `agent_backend == "goose"` (regardless of `advanced` setting)
- [ ] When the operator selects `goose`, the resulting `install.conf` contains none of the three keys above
- [ ] When the operator selects `claude`, all three prompts continue to fire as today (no behavior regression)
- [ ] Existing `tests/test_install.py` fixtures (all currently use `"claude"`) continue to pass without changes to their input lists
- [ ] At least one new test exercises the `"goose"` path and asserts the prompts are skipped

dcellison:
## Summary

- Three wizard prompts in `_cmd_config()` previously fired regardless of the operator's selected agent backend, even though the captured values are consumed only by `ClaudeCodeBackend`. Goose-backend operators were asked to set values that would be silently ignored at runtime. This PR adds backend-conditional gating for `CLAUDE_AUTOCOMPACT_PCT`, `CLAUDE_EFFORT_LEVEL`, and the legacy `CLAUDE_USER` prompt at install.py:730.
- Each gate uses the "initialize-before-gate" pattern: the captured variable is set to its dataclass default before the conditional, and the existing emission check (`if value != default`) correctly skips writing the env key under goose without needing a parallel emission gate. As a free side effect, this implicitly auto-prunes previously-set values from `install.conf` on a claude→goose switch — pinned by a regression test.
- The fourth planned gate from the spec (the `admin_os_user` prompt at install.py:382) is intentionally NOT included. That prompt fires inside `if advanced:`, which runs before `agent_backend` is defined at line 440. Honoring the gate would require structural reordering of the wizard. Deferred to the broader multi-backend rework per operator direction.

Fixes #380 (partial — three of four planned gates; admin_os_user gate deferred per above).

## Test plan

- [ ] `make test` (full suite, 2384 tests) — local: passes
- [ ] `test_goose_backend_skips_claude_only_prompts` (new): asserts CLAUDE_AUTOCOMPACT_PCT, CLAUDE_EFFORT_LEVEL, CLAUDE_USER all absent from install.conf when wizard ran under goose
- [ ] `test_goose_backend_prunes_existing_claude_only_keys` (new): pre-seeds install.conf with all three keys at non-default values, runs wizard under goose, asserts all three disappear (pins the implicit auto-prune)
- [ ] Three existing goose fixtures trimmed to remove the three surplus entries each (autocompact, effort, claude_user)
- [ ] Three pre-existing claude tests had been passing on borrowed time without explicit autocompact / effort / claude_user entries (relying on the retry-on-empty default fallback to absorb downstream entries by mistake; pytest output capture hid the validation prints). Updated to provide the entries explicitly so the tests do not depend on absorption side-effects.
- [ ] Verify `make config` walk-through under goose backend skips the autocompact, effort, and CLAUDE_USER prompts but otherwise behaves as before
- [ ] Verify `make config` walk-through under claude backend prompts for all three knobs as before (no behavior regression)

## Out of scope

- The `admin_os_user` prompt at install.py:382 (described above; deferred to broader multi-backend rework).
- Adding wizard prompts for `CLAUDE_MAX_SESSION_HOURS` / `CLAUDE_IDLE_TIMEOUT` (Claude-only at the consumer but no wizard prompt today).
- Explicit prune-on-switch logic for env vars not gated by this PR. The implicit auto-prune that this PR produces for the three gated keys is acknowledged and tested; broader prune-on-switch is a separate concern.
- Shared-with-Goose `CLAUDE_*` prompts (`CLAUDE_TIMEOUT_SECONDS`, `CLAUDE_MAX_CONTEXT_WINDOW`, `CLAUDE_MAX_BUDGET_USD` → `BUDGET_CEILING`, `CLAUDE_MODEL` → `DEFAULT_MODEL`). All four are consumed by both backends and are correctly NOT gated today.

### DIFF
--- a/src/kai/install.py
+++ b/src/kai/install.py
@@ -543,40 +543,53 @@ def _cmd_config() -> None:
     # conversation history. Lower values compact sooner, reducing token
     # usage at the cost of losing raw context earlier. Claude Code caps
     # this at ~83% regardless of what you set - values above that are
-    # silently clamped.
-    while True:
-        autocompact_pct = _prompt(
-            "Autocompact threshold (% of context window, 0 = default ~83%)",
-            existing_env.get("CLAUDE_AUTOCOMPACT_PCT", "80"),
-        )
-        try:
-            val = int(autocompact_pct)
-            if 0 <= val <= 100:
-                break
-        except ValueError:
-            pass
-        print("  Must be 0-100.")
-    print()
+    # silently clamped. Claude-binary specific (CLAUDE_AUTOCOMPACT_PCT
+    # is consumed only by ClaudeCodeBackend per claude.py); Goose has
+    # no autocompact concept, so the prompt is suppressed when goose is
+    # the selected backend. Initialize to "0" first (the dataclass
+    # default) so the env-emission check `int(autocompact_pct) != 0`
+    # below correctly skips writing the var for goose without needing
+    # a parallel gate down at the emission site.
+    autocompact_pct = "0"
+    if agent_backend == "claude":
+        while True:
+            autocompact_pct = _prompt(
+                "Autocompact threshold (% of context window, 0 = default ~83%)",
+                existing_env.get("CLAUDE_AUTOCOMPACT_PCT", "80"),
+            )
+            try:
+                val = int(autocompact_pct)
+                if 0 <= val <= 100:
+                    break
+            except ValueError:
+                pass
+            print("  Must be 0-100.")
+        print()
 
-    # CLAUDE_EFFORT_LEVEL is a truly-global setting (single value applies
-    # to every chat's inner Claude subprocess), so the prompt sits OUT-
-    # SIDE the per-user-vs-global if/else above and runs unconditionally.
-    # This matches the autocompact_pct precedent immediately above.
-    # _prompt_choice enforces the allow-list at wizard time, mirroring
-    # the runtime allow-list check in config._VALID_EFFORT_LEVELS so the
-    # operator cannot persist an invalid value to install.conf and have
-    # it fail later at service startup.
-    # Use the canonical EFFORT_LEVELS tuple from config so the wizard
-    # prompt and the runtime allow-list cannot drift out of sync. Cast
-    # to list because _prompt_choice expects a list (it joins with "/"
-    # for the display string and uses `in` for membership; tuple would
-    # work for both but the type signature asks for list).
-    claude_effort_level = _prompt_choice(
-        "Inner Claude effort level",
-        list(EFFORT_LEVELS),
-        existing_env.get("CLAUDE_EFFORT_LEVEL", "high"),
-    )
-    print()
+    # CLAUDE_EFFORT_LEVEL is consumed only by ClaudeCodeBackend's
+    # `--effort` CLI flag (claude.py); Goose has no equivalent flag,
+    # so the prompt is suppressed when goose is the selected backend.
+    # Initialize to "high" first (matches Config.claude_effort_level
+    # default) so the env-emission check `claude_effort_level != "high"`
+    # below correctly skips writing the var for goose without needing
+    # a parallel gate down at the emission site. _prompt_choice still
+    # enforces the allow-list at wizard time for the claude path,
+    # mirroring the runtime allow-list check in config._VALID_EFFORT_LEVELS
+    # so the operator cannot persist an invalid value to install.conf
+    # and have it fail later at service startup.
+    claude_effort_level = "high"
+    if agent_backend == "claude":
+        # Use the canonical EFFORT_LEVELS tuple from config so the wizard
+        # prompt and the runtime allow-list cannot drift out of sync. Cast
+        # to list because _prompt_choice expects a list (it joins with "/"
+        # for the display string and uses `in` for membership; tuple would
+        # work for both but the type signature asks for list).
+        claude_effort_level = _prompt_choice(
+            "Inner Claude effort level",
+            list(EFFORT_LEVELS),
+            existing_env.get("CLAUDE_EFFORT_LEVEL", "high"),
+        )
+        print()
 
     # -- Webhook server --
     print("-- Webhook server --")
@@ -725,6 +738,16 @@ def _cmd_config() -> None:
         print("  OS user isolation is now per-user. Set 'os_user' in users.yaml.")
         claude_user = ""
     elif admin_os_user:
+        claude_user = ""
+    elif agent_backend != "claude":
+        # CLAUDE_USER is the legacy global fallback for sudo subprocess
+        # isolation. Only ClaudeCodeBackend wires it through (claude.py
+        # invokes `sudo -H -u <user> -- claude ...`). Goose has no sudo
+        # path - GooseBackend in pool.py:186-198 takes no claude_user
+        # kwarg - so the prompt has no meaning when agent_backend is
+        # not claude. Short-circuit to "" here matches the two branches
+        # above, both of which suppress the prompt for the same shape
+        # of "this prompt does not apply" reason.
         claude_user = ""
     else:
         claude_user = _prompt(
--- a/src/kai/install.py
+++ b/src/kai/install.py
@@ -743,8 +743,8 @@ def _cmd_config() -> None:
         # CLAUDE_USER is the legacy global fallback for sudo subprocess
         # isolation. Only ClaudeCodeBackend wires it through (claude.py
         # invokes `sudo -H -u <user> -- claude ...`). Goose has no sudo
-        # path - GooseBackend in pool.py:186-198 takes no claude_user
-        # kwarg - so the prompt has no meaning when agent_backend is
+        # path - GooseBackend takes no claude_user kwarg in pool.py,
+        # so the prompt has no meaning when agent_backend is
         # not claude. Short-circuit to "" here matches the two branches
         # above, both of which suppress the prompt for the same shape
         # of "this prompt does not apply" reason.
--- a/tests/test_install.py
+++ b/tests/test_install.py
@@ -772,7 +772,7 @@ class TestCmdConfig:
                 "",  # github notify chat id (empty)
                 "false",  # voice
                 "false",  # tts
-                "",  # claude user (empty)
+                "",  # claude user (legacy single-user mode)
                 "false",  # memory enabled
                 "",  # perplexity key (empty)
             ]
@@ -925,7 +925,7 @@ class TestCmdConfig:
                 "",  # github notify chat id
                 "false",  # voice
                 "false",  # tts
-                "",  # claude user (empty)
+                "",  # claude user (legacy single-user mode)
                 "false",  # memory enabled
                 "",  # perplexity key (empty)
             ]
@@ -1034,8 +1034,6 @@ class TestCmdConfig:
                 "120",  # timeout
                 "10.0",  # budget
                 "200000",  # max context window
-                "80",  # autocompact pct
-                "",  # claude effort level (take default "high")
                 "8080",  # port
                 "test-secret",  # webhook secret
                 "~/Projects",  # workspace base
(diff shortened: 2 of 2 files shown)


dcellison:
## Review by Kai

The PR is well-structured. Here are my findings:

**Warning**

`tests/test_install.py` - `_base_inputs()` docstring says "set `llm_api_key=None` for ollama (which skips the API key prompt)" but the parameter is typed `llm_api_key: str = "sk-ant-test-key"`. Passing `None` is a Pyright strict violation. Since the project requires Pyright strict, this should be `str | None` (with `None` as the default when ollama is selected, or kept as `str` and the docstring corrected to say "just leave the default; it won't be consumed"). Not a runtime bug since the ollama branch never appends `llm_api_key`, but it is a type error the linter would catch in a non-test context.

**Suggestion**

`tests/test_install.py:1333` - The test docstring ends with "tracked in memory as `project_multi_backend_revisit_pending.md`". This is a reference to the Claude Code AI assistant's internal memory file system, not a real project artifact. A future developer reading this will have no idea what it means. Replace with a plain English reference like "tracked separately" or a GitHub issue number.

**Suggestion**

`_base_inputs()` silently ignores the `effort` parameter when `agent_backend != "claude"` (it falls into the empty `claude_only_pre_webhook` list). A caller passing `effort="xhigh"` for a goose backend gets no error and no effect. Adding a guard (`if agent_backend != "claude" and effort: raise ValueError(...)`) would prevent confusing silent no-ops in future tests.

---

**Everything else looks correct.** The "initialize-before-gate" pattern is sound - all three variables are set to their dataclass defaults before the conditional, so the downstream emission checks (`!= 0`, `!= "high"`, `!= ""`) correctly suppress writes for goose without a parallel gate at the emission site. The `elif agent_backend != "claude":` placement in the `CLAUDE_USER` chain is safe regardless of ordering because both prior branches also assign `""`. The two new regression tests pin both the skip and prune behaviors. Trimmed fixtures match the new wizard flow.

dcellison:
## Review by Kai

The implementation is clean. Here's what I found after reviewing the diff against the prior thread:

**Prior review items - status**

All three issues from the prior review are resolved:
- `llm_api_key: str` type annotation: the docstring no longer says to pass `None` for ollama; it now says the default is "harmless and unused". No type violation.
- AI memory reference at line 1333: that test was not modified in this PR; not re-raised.
- `effort` guard for non-claude backends: `ValueError` is now raised at lines ~1215-1222.

---

**New findings**

**Suggestion** - `install.py`, the new `claude_user` comment block (~line 741):

```python
# GooseBackend in pool.py:186-198 takes no claude_user kwarg
```

Line-number references in cross-file comments are fragile - pool.py can be edited independently and this will silently drift. Drop the `:186-198` qualifier or replace with a function/class name reference that's stable across refactors.

**Suggestion** - `tests/test_install.py`, the `claude_user_entry` comment in `_base_inputs`:

```python
# Legacy CLAUDE_USER prompt at install.py:730 is gated on
# agent_backend == "claude" by issue #380 (Change 4).
```

The "(Change 4)" label will be opaque to anyone who doesn't have access to the internal spec. Since the PR description gives the full context, "Change 4 of the spec" should be replaced with just "issue #380" or left out entirely. Minor, but the CLAUDE.md commenting guide implies comments should explain the WHY, not reference private documents.

---

**Everything else is correct.** The "initialize-before-gate" pattern is applied consistently across all three variables. The `elif agent_backend != "claude":` placement in the `CLAUDE_USER` chain is safe since both preceding branches already assign `""`. The two new regression tests correctly pin skip-and-prune behavior. The `_base_inputs` refactor produces an identical input list for `agent_backend="claude"` (the default), so existing call sites are unaffected.

dcellison:
## Review by Kai

The diff is clean. One new finding:

**Suggestion** - `tests/test_install.py`, `_base_inputs` comment at the `claude_user_entry` block:

```python
# Legacy CLAUDE_USER prompt at install.py:730 is gated on
# agent_backend == "claude" by issue #380.
```

`install.py:730` is a fragile line-number reference - the same pattern that was fixed in the prior round for `pool.py:186-198`. This comment is new code in this iteration (the `_base_inputs` function was substantially rewritten), so this is not a re-raise. Drop the `:730` qualifier; the issue number and variable name are sufficient to locate the gate.

---

Everything else is correct:

- "Initialize-before-gate" is applied consistently for all three variables: `autocompact_pct = "0"`, `claude_effort_level = "high"`, and the `elif agent_backend != "claude": claude_user = ""` placement. In each case the downstream emission check (`!= 0`, `!= "high"`, `!= ""`) correctly suppresses the write for goose without a parallel gate at the emission site.
- The `elif agent_backend != "claude":` ordering in the `claude_user` chain is safe - all three non-prompt branches assign `""`, so their order is irrelevant to correctness.
- The `ValueError` guard in `_base_inputs` for `effort` with non-claude backends is a sound defensive choice.
- Both new regression tests correctly pin skip-and-prune behavior. The prune test's pre-seeded conf cannot contaminate output because the gated block never runs and the initialized default (`"0"`, `"high"`, `""`) fails the emission check.
- Trimmed goose fixtures match the new wizard flow.
- Prior review items (pool.py line-number reference, "(Change 4)" label, `llm_api_key` docstring, `effort` guard) are all resolved.

dcellison:
## Review by Kai

Looking at this diff against the prior review thread - all three previously-raised issues (pool.py line-number reference, "(Change 4)" label, `install.py:730` reference in test comment) are resolved. One new finding:

**Suggestion** - `tests/test_install.py`, `test_goose_backend_skips_claude_only_prompts` docstring:

```python
Note: the admin_os_user prompt at install.py:382 is NOT gated
in this PR.
```

This is a new instance of the exact pattern caught in the last two review rounds - a fragile line-number reference embedded in a comment. Drop the `:382` qualifier; "the admin_os_user prompt" is already sufficient to locate it, and the surrounding sentence explains why it is deferred.

---

Everything else is correct:

- "Initialize-before-gate" is applied consistently: `autocompact_pct = "0"`, `claude_effort_level = "high"`, and the `elif agent_backend != "claude": claude_user = ""` placement. In each case the downstream emission check (`!= 0`, `!= "high"`, `!= ""`) correctly suppresses the write for goose without a parallel gate at the emission site.
- The `elif agent_backend != "claude":` insertion in the `claude_user` chain is safe - all three prior branches also assign `""`, so order is irrelevant to correctness.
- The `_base_inputs` refactor is sound: `ValueError` guard for `effort` on non-claude backends, conditional `backend_block`/`claude_only_pre_webhook`/`claude_user_entry` lists all match the wizard's runtime conditionals.
- Both new regression tests correctly pin skip-and-prune behavior. The prune test's pre-seeded values cannot survive because the initialized defaults (`"0"`, `"high"`, `""`) fail the emission check.
- Trimmed goose fixtures match the new wizard flow.
- No fragile line numbers in install.py comments; the `pool.py` reference now correctly omits the line range.

dcellison:
## Review by Kai

Looking at the diff against the prior review thread - the one open finding from the last round (`:382` line-number reference in `test_goose_backend_skips_claude_only_prompts`) is resolved. The docstring now reads "lives inside the `if advanced:` block in install.py" with no line number.

No new issues found. Specifically:

- **install.py** - "initialize-before-gate" is applied consistently: `autocompact_pct = "0"`, `claude_effort_level = "high"`, and `elif agent_backend != "claude": claude_user = ""`. All three downstream emission checks (`!= 0`, `!= "high"`, `!= ""`) correctly suppress the write for goose without a parallel gate at the emission site.
- **install.py comments** - No fragile line-number cross-references. The `pool.py` reference correctly omits the line range (addressed in a prior round).
- **`_base_inputs` refactor** - The `ValueError` guard for `effort` on non-claude backends is sound. The three conditional lists (`backend_block`, `claude_only_pre_webhook`, `claude_user_entry`) correctly mirror the wizard's runtime conditionals. Default `agent_backend="claude"` means all existing callers are unaffected.
- **New regression tests** - Both tests are correct. `test_goose_backend_skips_claude_only_prompts` pins that the prompts do not fire; `test_goose_backend_prunes_existing_claude_only_keys` pins that the implicit auto-prune works because pre-seeded non-default values cannot survive when the initialized defaults fail the emission check.
- **Trimmed goose fixtures** - The three removed input entries (autocompact, effort, claude_user) match the new wizard flow exactly.

This PR is clean.

PR #382, opened 2026-04-24
fix(webhook): reject non-object JSON bodies with 400

### LINKED ISSUE #377: Reject non-object JSON bodies cleanly in webhook handlers
## Summary

Every `/webhook` and `/api/*` handler in `src/kai/webhook.py` parses the request body via `await request.json()` and then immediately calls `payload.get(...)` without verifying that the parsed value is a dict. JSON `null`, `[]`, `42`, and `"string"` all parse successfully (no `JSONDecodeError`), then raise `AttributeError` on `.get()`, which escapes to aiohttp as an unstyled 500 with an HTML traceback body instead of a clean JSON error.

Surfaced as a suggestion during PR #376 review.

## Affected sites

`grep -n "payload = await request.json" src/kai/webhook.py` returns 9 hits, covering every JSON-body handler:

| Line | Handler |
|---|---|
| 878 | `_handle_generic` |
| 1002 | `_handle_schedule` |
| 1192 | `_handle_update_job` |
| 1298 | `_handle_service_call` |
| 1337 | `_handle_send_message` |
| 1385 | `_handle_send_file` |
| 1523 | `_handle_memory_add` (PR #376) |
| 1618 | `_handle_memory_search` (PR #376) |
| 1767 | `_handle_memory_delete_all` (PR #376) |

This is a pre-existing pattern that PR #376 inherited. The PR addressed input-type validation for individual fields after `payload.get()`; the wrapping shape-check is the layer above that.

## Fix

One additional check after the `JSONDecodeError` guard, before any `payload.get()` call:

```python
if not isinstance(payload, dict):
    return web.json_response(
        {"error": "Request body must be a JSON object"}, status=400
    )
```

Apply uniformly across all 9 sites. Add a test per handler (or one parametrized test) that sends each non-object JSON shape (`null`, `[]`, `42`, `"string"`) and asserts the 400 + JSON error body.

## Risk

Low. All affected routes are localhost-only with secret authentication, so the failure mode requires either an authenticated client error or a misconfigured caller. The HTML 500 surface is ugly but not exploitable as written.

## Scope

Pre-existing pattern, not introduced by PR #376. Filed separately so PR #376 can ship as the foundation REST API rather than expanding to fix unrelated handlers. A small dedicated PR can address all 9 sites uniformly.

dcellison:
## Summary

- Adds an `isinstance(payload, dict)` check to all 9 JSON-body handlers in `webhook.py`. Without it, valid-but-non-object JSON (`null`, `[]`, `42`, `"string"`) parsed cleanly past the existing `JSONDecodeError` guard and then raised `AttributeError` on the next `payload.get()`, escaping as an unstyled HTML 500 traceback.
- Eight of the nine sites get an identical insertion immediately after the existing `JSONDecodeError` guard. The ninth (`_handle_service_call`) is structurally different: it treats `JSONDecodeError` as "no body provided" and continues with field defaults, so the new check goes inside the existing try block. Previously a literal `null` body to that handler raised `AttributeError` and escaped as 500; now it returns 400, matching the other handlers. Silently accepting `null` would mask client bugs that meant to send a body.
- New `TestNonObjectPayloadRejection` cross-handler parametrized test class. 9 handlers × 6 non-object shapes (null, empty list, non-empty list, int, string, bool) = 54 test cases. Bool is included because Python's `bool` is a subclass of `int`, which would silently pass an `isinstance(payload, int)` check that someone might be tempted to write later.

Fixes #377

## Test plan

- [ ] `make test` (full suite, 2438 tests) — local: passes (was 2384; +54 from new parametrized test cases)
- [ ] `TestNonObjectPayloadRejection` (new): 54 test cases verify every handler rejects every non-object shape with 400
- [ ] No new line-number cross-file references in code comments (per the rule that bit PR #381 three times)
- [ ] Manual smoke: `curl -X POST http://localhost:8080/api/memory/add -H "X-Webhook-Secret: $KAI_WEBHOOK_SECRET" -H "Content-Type: application/json" -d 'null'` should return `{"error": "Request body must be a JSON object"}` with HTTP 400 (not the prior unstyled HTML 500)

## Risk

Low. All affected routes are localhost-only with secret authentication, so the failure mode requires either an authenticated client error or a misconfigured caller. Behavior change for `_handle_service_call` (literal `null` body now 400 instead of 500) is in the direction the issue intends; any existing client sending `null` was already broken.

## Out of scope

- Refactoring the inconsistent JSONDecodeError response styles. Two of the 9 sites still use `web.Response(text=...)` for the JSONDecodeError path while the other seven use `web.json_response({"error": ...})`. Aligning these is a separate concern; this PR only adds the missing isinstance check at each site.
- Adding handler-helper extraction (`_parse_json_object()` or similar). The pattern is identical at 8 of 9 sites, so extracting a helper is a reasonable refactor — but doing it as part of this fix would expand scope and obscure the bug-fix in a refactor diff.

### DIFF
--- a/src/kai/webhook.py
+++ b/src/kai/webhook.py
@@ -880,7 +880,10 @@ async def _handle_generic(request: web.Request) -> web.Response:
         return web.Response(status=400, text="Invalid JSON")
 
     # Reject non-object JSON shapes (null, lists, scalars, strings) before
-    # any payload.get() call. See _handle_schedule for the shared rationale.
+    # any payload.get() call. JSON bodies like `null`, `[]`, `42`, or
+    # `"string"` parse successfully (no JSONDecodeError) but raise
+    # AttributeError on .get(), which would escape as an unstyled HTML
+    # 500 traceback instead of a clean JSON 400.
     if not isinstance(payload, dict):
         return web.json_response({"error": "Request body must be a JSON object"}, status=400)
 
--- a/src/kai/webhook.py
+++ b/src/kai/webhook.py
@@ -879,6 +879,11 @@ async def _handle_generic(request: web.Request) -> web.Response:
     except json.JSONDecodeError:
         return web.Response(status=400, text="Invalid JSON")
 
+    # Reject non-object JSON shapes (null, lists, scalars, strings) before
+    # any payload.get() call. See _handle_schedule for the shared rationale.
+    if not isinstance(payload, dict):
+        return web.json_response({"error": "Request body must be a JSON object"}, status=400)
+
     # Use the "message" field if present (including empty string),
     # otherwise dump the full JSON. `is not None` avoids treating "" as absent.
     msg = payload.get("message")
@@ -1002,6 +1007,14 @@ async def _handle_schedule(request: web.Request) -> web.Response:
         payload = await request.json()
     except json.JSONDecodeError:
         return web.json_response({"error": "Invalid JSON"}, status=400)
+
+    # Reject non-object JSON shapes (null, lists, scalars, strings) before
+    # any payload.get() call. JSON bodies like `null`, `[]`, `42`, or
+    # `"string"` parse successfully (no JSONDecodeError) but raise
+    # AttributeError on .get(), which would escape as an unstyled HTML
+    # 500 traceback instead of a clean JSON 400.
+    if not isinstance(payload, dict):
+        return web.json_response({"error": "Request body must be a JSON object"}, status=400)
 
     # Extract and validate required fields
     name = payload.get("name")
@@ -1193,6 +1206,14 @@ async def _handle_update_job(request: web.Request) -> web.Response:
     except json.JSONDecodeError:
         return web.json_response({"error": "Invalid JSON"}, status=400)
 
+    # Reject non-object JSON shapes (null, lists, scalars, strings) before
+    # any payload.get() call. JSON bodies like `null`, `[]`, `42`, or
+    # `"string"` parse successfully (no JSONDecodeError) but raise
+    # AttributeError on .get(), which would escape as an unstyled HTML
+    # 500 traceback instead of a clean JSON 400.
+    if not isinstance(payload, dict):
+        return web.json_response({"error": "Request body must be a JSON object"}, status=400)
+
     # Validate schedule_type if provided
     new_schedule_type = payload.get("schedule_type")
     if new_schedule_type and new_schedule_type not in _VALID_SCHEDULE_TYPES:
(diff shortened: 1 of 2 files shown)


dcellison:
## Review by Kai

**Overall: clean fix, well-tested. Three findings below.**

---

**Warning - PR/issue references in code comments (test_webhook_api.py, multiple lines)**

The block comment before `_NON_OBJECT_HANDLERS` references `PR #376`, and the class docstring references `Issue #377`. The convention says explicitly: "Don't reference the current task, fix, or callers... since those belong in the PR description and rot as the codebase evolves." These references will mean nothing to a future reader who doesn't have that PR open. The rationale is sound but belongs only in the PR description, not in the source.

Same applies to the class docstring:
```python
"""Issue #377: every JSON-body handler must reject non-object
JSON shapes with 400 instead of crashing on AttributeError."""
```
Strip the issue number; the sentence after the colon stands on its own.

---

**Suggestion - Forward reference in `_handle_generic` comment (webhook.py ~line 882)**

The comment added at `_handle_generic` reads:

```python
# Reject non-object JSON shapes (null, lists, scalars, strings) before
# any payload.get() call. See _handle_schedule for the shared rationale.
```

`_handle_generic` appears around line 879 and `_handle_schedule` around line 1008 - the reader is being told to scroll forward ~130 lines to find the rationale. Either copy the rationale here (as done at all the later handlers) or drop the cross-reference and just state the reason in full. A forward reference to a function that comes later is harder to follow than a backward one.

---

**Suggestion - Box-drawing separator in test comment (test_webhook_api.py, line ~2759)**

```python
# ── Cross-handler payload-shape rejection (issue #377) ────────────────
```

The `──` and `────` characters are U+2500 BOX DRAWINGS LIGHT HORIZONTAL, not em/en dashes, so they technically don't violate the literal rule. But they're visually indistinguishable from a horizontal rule and the same decorative category. Whether this is in-bounds is a judgment call for the author, but it's worth flagging for consistency.

---

No bugs in the logic. The `_handle_service_call` placement inside the `try` block is correct - the `return` exits before `except json.JSONDecodeError` would catch anything, so the optional-body behavior is preserved for true parse failures. The 54-case parametrized matrix is good coverage, including the `bool`/`int` subclass edge case. Nothing security-relevant introduced.

dcellison:
## Review by Kai

**Overall: clean, no new issues.**

The two substantive items from the prior review were addressed:

- `_handle_generic` comment no longer has the forward reference to `_handle_schedule`. It now carries the full self-contained rationale, matching all other handler sites.
- The class docstring `"""Issue #377: ..."""` had the issue number stripped; the sentence stands on its own.

The two cosmetic items from the prior review (issue reference in the `# ──` separator line, box-drawing characters) were not addressed. Per your prior decision, no point re-raising them.

**Logic check:**

- `isinstance(payload, dict)` is the correct predicate: rejects `None`, lists, ints, strings, and bools (including `True`/`False`, which would silently pass an `isinstance(payload, int)` check). All 9 handlers covered.
- `_handle_service_call` placement inside the `try` block is correct. The `return` on a non-dict payload exits before the `except json.JSONDecodeError` clause, so true parse failures still fall through to the except and get treated as "no body provided" as intended.
- The parametrized test matrix (`pytest.param` with two positional args unpacked into `handler, override`) is valid pytest usage.

No bugs, no security issues, no new convention violations.

PR #389, opened 2026-04-26
fix(error-ux): surface real error + budget recovery directive (#326)

### LINKED ISSUE #326: Error recovery overwrites streaming message, erasing tool-use context

dcellison:
## Summary

Three behavior changes addressing all four acceptance criteria of issue #326. The dominant trigger of the empty-payload `is_error` variant in production is `BUDGET_CEILING` exhaustion (default $10/session); pre-#326 this surfaced as the literal string `"Error: None"` in chat with no recovery guidance. This PR makes the failure honest, contextual, and recoverable.

Fixes #326.

### claude.py — error-event population

`_send_locked`'s result-event handler now reads from `event.get("errors")` when `result_text` is empty. The CLI's `is_error=true` events come in two shapes: (a) `result` populated with a human-readable reason, (b) `result` empty BUT `errors` populated with a list of strings (the documented BUDGET_CEILING variant: `errors: ["Reached maximum budget ($10)"]`). Pre-#326 only shape (a) produced a non-None error string; shape (b) silently fell through to None.

Falls back to `"no error detail provided"` sentinel when both fields are empty so the downstream `"Error: None"` surface cannot recur on a future CLI variant.

### bot.py — message lifecycle

The error-rendering block in `_handle_response` no longer calls `_edit_message_safe(live_msg, error_text)`. It sends a new `reply_text` instead. Pre-#326 the in-place edit erased any tool-use, partial reasoning, and intermediate output the user was watching - on long sessions, minutes of visible work disappeared into a single error line.

```python
if not final_response.success:
    real_error = final_response.error or "no error detail provided"
    error_text = f"Error: {real_error}"
    log_message(direction="assistant", chat_id=chat_id, text=f"[error: {real_error}]")
    # Append, never overwrite - preserve the streamed context.
    await update.message.reply_text(error_text)
    if _is_budget_exhaustion(real_error):
        await update.message.reply_text(_budget_recovery_hint(real_error))
    return
```

### bot.py — budget-exhaustion recovery directive

Two new module-level helpers:

- `_is_budget_exhaustion(error_text)` - case-insensitive substring match on the CLI's documented `"maximum budget"` phrasing
- `_budget_recovery_hint(error_text)` - inlines the dollar amount via regex; falls back to a no-amount template when the amount isn't extractable

Sent as a separate Telegram message (not appended to the error notice on the same line) so clients show it with a notification badge and the user immediately sees "what to do next" alongside "what happened":

```
Hit session budget ($10). Type /new to start a fresh session, or ask
your operator to raise BUDGET_CEILING in /etc/kai/env.
```

### Bot responsiveness post-error

The bot remains responsive to `/new` and other commands after the error - the per-chat lock is released by the existing return path and the dispatch loop is unaffected. No code change needed for that property; preserved by structure.

## Tests

New `tests/test_error_recovery.py` (13 tests):

- **claude.py error-event handling (3):** `errors` field populates response error; empty-both-fields falls back to sentinel; non-empty `result` takes precedence.
- **Budget detection helpers (5):** canonical phrasing match, non-match rejection (auth/network/None/empty), dollar-amount inlining, fallback when amount unparseable, None-tolerance.
- **bot.py message lifecycle (5):** error doesn't overwrite live_msg; budget error sends two messages; non-budget error sends one; no `"Error: None"` literal in any user-facing or log-facing output; chat string matches history-log entry.

Updated `tests/test_bot.py::test_error_response_with_live_msg`: it pinned the OLD overwrite contract and asserted the edit happened. Now pins the NEW append-not-overwrite contract; comment captures the pre-#326 vs post-#326 invariant.

Full suite: **2497 passed, 1 skipped** (the opt-in classifier eval, gated by `RUN_CLASSIFIER_EVAL=1`).

## Acceptance criteria from #326 (all four)

- [x] **Failing mid-stream after tool use no longer wipes the user-visible streaming message.** The error appears as a follow-up reply, not as a replacement.
- [x] **Every `is_error=true` event produces a log line correlating to the synthetic history entry.** Already partially satisfied by PR #332's diagnostic warning; verified preserved.
- [x] **`Error: None` is replaced with the real reason from the CLI's `errors` field.** Belt-and-suspenders fallback at both layers (claude.py sentinel + bot.py `or` clause) prevents recurrence.
- [x] **Budget-exhaustion errors include the recovery directive.** Bot stays responsive to `/new` after the error.

## Test plan

- [x] `make check` (ruff check + format) - local: passes
- [x] `make test` (full suite) - local: 2497 passed, 1 skipped
- [ ] After deploy, send a message that exceeds session budget; confirm chat shows `Error: Reached maximum budget ($10)` followed by the directive message. Verify `/new` works immediately after.
- [ ] After deploy, look for `kai.claude WARNING Result event with is_error=true has no result field` log lines; for any that appear, confirm the corresponding chat message no longer reads `Error: None`.
- [ ] Spot-check a normal long-running session ends without the streamed message getting wiped on any error path.

## Out of scope

- Auto-recovery (transparent new-session creation with handoff). Significant design work; most of the value is captured by surfacing the error + guidance, not by automation.
- Operator-side budget tuning workflow (raising `BUDGET_CEILING` via wizard). Already exists; the directive points at it.
- Error-class taxonomy beyond "budget" vs "other". Stage-2 episode generation has its own log surface (`memory.episode`); episode failures are not in scope here.

### DIFF
--- a/src/kai/bot.py
+++ b/src/kai/bot.py
@@ -172,7 +172,7 @@ _QUEUED_MESSAGE_MARKER = (
 _LOCK_ACQUIRE_TIMEOUT = 3600  # 1 hour
 
 
-# Budget-exhaustion recovery directive (issue #326).
+# Budget-exhaustion recovery directive.
 #
 # When the inner Claude session hits BUDGET_CEILING (default $10), the
 # CLI emits an `is_error=true` event whose `errors` field carries the
@@ -202,7 +202,12 @@ _BUDGET_RECOVERY_HINT_NO_AMOUNT = (
 # amount can't be extracted, and the no-amount hint is sent. The chat
 # still surfaces the real error string from claude.py either way -
 # only the dollar-amount-aware directive is missed.
-_BUDGET_PHRASE_RE = re.compile(r"maximum budget \(\$([\d.]+)\)", re.IGNORECASE)
+#
+# `\d+(?:\.\d+)?` accepts integer ($10) or one-decimal ($10.50)
+# amounts and rejects malformed compositions like $1.2.3.4. Tighter
+# than `[\d.]+`; signals intent to a future reader that we expect a
+# single decimal point at most.
+_BUDGET_PHRASE_RE = re.compile(r"maximum budget \(\$(\d+(?:\.\d+)?)\)", re.IGNORECASE)
 
 
 def _is_budget_exhaustion(error_text: str | None) -> bool:
@@ -3603,16 +3608,21 @@ async def _handle_response(
         # Send the error notice as a NEW message (not an edit of the
         # live streamed message), so any tool-use, partial reasoning,
         # and intermediate output the user was watching stays visible.
-        # Pre-#326 the live_msg.edit_text overwrite erased that
-        # context entirely, which on long sessions could mean minutes
-        # of visible work disappearing into a single error line.
-        await update.message.reply_text(error_text)
+        # The previous in-place edit erased that context entirely,
+        # which on long sessions could mean minutes of visible work
+        # disappearing into a single error line. _reply_safe is the
+        # right wrapper here: error strings can carry markdown-like
+        # characters (parens, dollar signs, brackets) that Telegram's
+        # Markdown parser sometimes rejects, and the wrapper falls
+        # back to plain text on BadRequest while letting network
+        # errors propagate naturally.
+        await _reply_safe(update.message, error_text)
         # Budget-exhaustion gets a recovery directive as a second
         # follow-up message. Other error types fall through with no
         # extra guidance; structure leaves room for additional
         # error-class directives if recurring patterns emerge.
         if _is_budget_exhaustion(real_error):
-            await update.message.reply_text(_budget_recovery_hint(real_error))
+            await _reply_safe(update.message, _budget_recovery_hint(real_error))
         return
 
     # Persist session info for /stats (cost accumulates across interactions)
--- a/src/kai/bot.py
+++ b/src/kai/bot.py
@@ -172,6 +172,63 @@ _QUEUED_MESSAGE_MARKER = (
 _LOCK_ACQUIRE_TIMEOUT = 3600  # 1 hour
 
 
+# Budget-exhaustion recovery directive (issue #326).
+#
+# When the inner Claude session hits BUDGET_CEILING (default $10), the
+# CLI emits an `is_error=true` event whose `errors` field carries the
+# string "Reached maximum budget ($N)". `claude.py` populates
+# AgentResponse.error from that field; the bot uses the helper below
+# to detect the budget variant and append a recovery hint as a
+# follow-up message.
+#
+# The hint is a SEPARATE message, not appended to the error notice
+# itself, so Telegram clients show it with a notification badge and
+# the user immediately sees "what to do next" alongside "what
+# happened". The bot remains responsive to /new (and every other
+# command) after the error - the per-chat lock is released by the
+# return below, and the dispatch loop continues unaffected.
+_BUDGET_RECOVERY_HINT_WITH_AMOUNT = (
+    "Hit session budget ({amount}). Type /new to start a fresh session, "
+    "or ask your operator to raise BUDGET_CEILING in /etc/kai/env."
+)
+_BUDGET_RECOVERY_HINT_NO_AMOUNT = (
+    "Hit session budget. Type /new to start a fresh session, "
+    "or ask your operator to raise BUDGET_CEILING in /etc/kai/env."
+)
+
+# Matches the CLI's "Reached maximum budget ($N)" wording. The regex
+# is anchored on the documented phrasing; if Anthropic ever changes the
+# wording the budget detection regresses to "no match", the dollar
+# amount can't be extracted, and the no-amount hint is sent. The chat
+# still surfaces the real error string from claude.py either way -
+# only the dollar-amount-aware directive is missed.
+_BUDGET_PHRASE_RE = re.compile(r"maximum budget \(\$([\d.]+)\)", re.IGNORECASE)
+
+
+def _is_budget_exhaustion(error_text: str | None) -> bool:
+    """True iff the error string indicates BUDGET_CEILING exhaustion.
+
+    Substring match on the CLI's documented phrasing. Wording change
+    in a future Anthropic release would silently regress this to
+    False, in which case the recovery hint stops appearing - the
+    error notice itself still surfaces. A regression test pins the
+    current phrasing.
+    """
+    if not error_text:
+        return False
+    return "maximum budget" in error_text.lower()
+
+
+def _budget_recovery_hint(error_text: str | None) -> str:
+    """Return the budget-exhaustion recovery directive, with the
+    dollar amount inlined when extractable from the error string."""
+    if error_text:
+        match = _BUDGET_PHRASE_RE.search(error_text)
+        if match:
+            return _BUDGET_RECOVERY_HINT_WITH_AMOUNT.format(amount=f"${match.group(1)}")
+    return _BUDGET_RECOVERY_HINT_NO_AMOUNT
+
+
 async def _acquire_lock_or_kill(
     chat_id: int,
     pool: "SubprocessPool",
@@ -3534,12 +3591,28 @@ async def _handle_response(
         return
 
     if not final_response.success:
-        error_text = f"Error: {final_response.error}"
-        log_message(direction="assistant", chat_id=chat_id, text=f"[error: {final_response.error}]")
-        if live_msg:
-            await _edit_message_safe(live_msg, error_text)
-        else:
-            await update.message.reply_text(error_text)
+        # Defensive fallback: claude.py now always populates `error`
+        # with a non-None string for is_error events (see the
+        # response_error resolution there). The `or` here is belt-and-
+        # suspenders against a future change that re-introduces None,
+        # so the literal "Error: None" string can't reappear via this
+        # surface even on a regression.
+        real_error = final_response.error or "no error detail provided"
+        error_text = f"Error: {real_error}"
+        log_message(direction="assistant", chat_id=chat_id, text=f"[error: {real_error}]")
+        # Send the error notice as a NEW message (not an edit of the
+        # live streamed message), so any tool-use, partial reasoning,
+        # and intermediate output the user was watching stays visible.
+        # Pre-#326 the live_msg.edit_text overwrite erased that
+        # context entirely, which on long sessions could mean minutes
+        # of visible work disappearing into a single error line.
+        await update.message.reply_text(error_text)
+        # Budget-exhaustion gets a recovery directive as a second
+        # follow-up message. Other error types fall through with no
+        # extra guidance; structure leaves room for additional
+        # error-class directives if recurring patterns emerge.
+        if _is_budget_exhaustion(real_error):
+            await update.message.reply_text(_budget_recovery_hint(real_error))
         return
 
     # Persist session info for /stats (cost accumulates across interactions)
--- a/src/kai/claude.py
+++ b/src/kai/claude.py
@@ -662,16 +662,21 @@ class ClaudeCodeBackend(AgentBackend):
                     #   (b) `result` empty BUT `errors` populated with a
                     #       list of strings (the documented variant for
                     #       BUDGET_CEILING exhaustion: errors carries
-                    #       ["Reached maximum budget ($N)"]). Pre-#326
-                    #       this fell through to None, which `bot.py`
-                    #       rendered as the literal "Error: None"
-                    #       string in chat.
+                    #       ["Reached maximum budget ($N)"]).
                     #
                     # Falls back to a non-None sentinel when both fields
-                    # are empty so downstream rendering never re-introduces
-                    # the "Error: None" surface, even on a future CLI
-                    # variant that emits is_error=true with neither field
-                    # populated.
+                    # are empty so downstream rendering never produces
+                    # the literal "Error: None" string in chat, even on
+                    # a future CLI variant that emits is_error=true with
+                    # neither field populated.
+                    #
+                    # Note on `result_text`: it is `event.get("result", "")`
+                    # captured a few lines above, with no transformation
+                    # applied. Using `result_text` here (rather than
+                    # `event.get("result")` again) is equivalent and avoids
+                    # the second dict lookup; the truthiness check still
+                    # catches the empty-string case which is what the
+                    # branch table targets.
                     response_error: str | None = None
                     if event.get("is_error", False):
                         if result_text:
--- a/src/kai/claude.py
+++ b/src/kai/claude.py
@@ -654,13 +654,41 @@ class ClaudeCodeBackend(AgentBackend):
                             event.get("duration_ms"),
                             sorted(event.keys()),
                         )
+                    # Resolve the error string for downstream consumers.
+                    # The CLI's `is_error=true` events come in two shapes:
+                    #
+                    #   (a) `result` populated with a human-readable
+                    #       reason. Use it directly.
+                    #   (b) `result` empty BUT `errors` populated with a
+                    #       list of strings (the documented variant for
+                    #       BUDGET_CEILING exhaustion: errors carries
+                    #       ["Reached maximum budget ($N)"]). Pre-#326
+                    #       this fell through to None, which `bot.py`
+                    #       rendered as the literal "Error: None"
+                    #       string in chat.
+                    #
+                    # Falls back to a non-None sentinel when both fields
+                    # are empty so downstream rendering never re-introduces
+                    # the "Error: None" surface, even on a future CLI
+                    # variant that emits is_error=true with neither field
+                    # populated.
+                    response_error: str | None = None
+                    if event.get("is_error", False):
+                        if result_text:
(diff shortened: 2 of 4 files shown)


dcellison:
## Review by Kai

Looking at this PR across all three changed files.

---

### Warning

**`bot.py:3591-3618` - Error path drops the `_edit_message_safe` wrapper**

The old `live_msg` branch called `_edit_message_safe`, which presumably absorbs Telegram exceptions (rate limits, message-too-long, chat-not-found). The new path unconditionally calls `update.message.reply_text` with no equivalent protection. The no-`live_msg` branch already had this gap, but the change extends it to the formerly-safe case. If `reply_text` raises while `live_msg` existed (meaning the session was well underway), the exception propagates from `_handle_response` with no user-visible notice. Worth wrapping in a try/except or confirming the call-site already catches Telegram errors.

**`bot.py` / `claude.py` - Issue number references in production code comments**

The comments contain `pre-#326`, `issue #326`, and similar references in at least nine places across both files. Per the project conventions, these belong in the PR description and git history, not in code - they'll mislead future readers once the issue tracker context is gone. Fine in test docstrings; not in `bot.py` and `claude.py` production code.

---

### Suggestion

**`claude.py:~665` - `result_text` vs `event.get("result")` in the new error-resolution block**

The new block uses `result_text` (presumably set earlier in the function) as the check for shape (a). The original line used `event.get("result")` directly. If `result_text` is stripped or transformed before this point, the two are not equivalent - a non-empty raw `result` could become falsy and fall through to the `errors` branch. Not visible as a bug in the tests, but worth a comment confirming `result_text is event.get("result")` or using `event.get("result")` directly for clarity.

**`bot.py:~193` - `/etc/kai/env` hardcoded in recovery hint**

The hint string points users to `/etc/kai/env`. If `make install` ever changes the config path, the directive silently misleads. A constant or a reference to the configured path would be more resilient, though given the current install conventions this is cosmetic.

**`_BUDGET_PHRASE_RE` - `[\d.]+` matches structurally invalid amounts**

`[\d.]+` accepts `$1.2.3.4`. Harmless since the source is CLI output, but `\d+(?:\.\d+)?` would be more precise and signal intent.

---

### Clean

- The `response_error` resolution logic in `claude.py` is correct: shape (a) wins, shape (b) joins with `"; "`, sentinel covers the pathological empty case.
- `_is_budget_exhaustion` + `_budget_recovery_hint` are straightforward pure functions; the None-safety is correct.
- The test changes correctly pin the new append-not-overwrite contract and cover the four documented edge cases. The sentinel assertion (`== "no error detail provided"`) is appropriately exact.
- No injection risk: `match.group(1)` is constrained to `[\d.]+` before formatting into the hint string.

dcellison:
## Review by Kai

Looking at the diff against the prior review thread, here is what's new or materially changed.

---

### Warning

**`tests/test_error_recovery.py:TestErrorMessageLifecycle` - `test_error_does_not_overwrite_live_msg` likely does not exercise the `live_msg` path**

The mock pool in `_make_pool_yielding_error` yields a terminal `done=True` error event immediately, before any streamed text. In `_handle_response`, `live_msg` is normally created only after at least one text chunk arrives. So in this test, `live_msg` is almost certainly never created - meaning `_edit_message_safe` would not have been called even under the *old* code, and patching it and asserting `not_called` is trivially true regardless of the fix.

The meaningful coverage for the overwrite regression is in the updated `test_bot.py::test_error_response_with_live_msg`, which explicitly constructs a `live_msg`. The new test's name implies it covers the same scenario but probably doesn't.

---

### Suggestion

**`tests/test_error_recovery.py:TestErrorMessageLifecycle` - assertions on `reply_text` while production calls `_reply_safe`**

All five bot lifecycle tests assert on `update.message.reply_text.await_count` and `await_args_list`, but the implementation calls `_reply_safe(update.message, ...)`. This works today because `_reply_safe` calls `reply_text` internally, but the coupling is implicit. If `_reply_safe` is refactored to call `reply_markdown` or a different method on its first attempt, the count assertions would silently read zero - a false pass, not a failing test. Patching `kai.bot._reply_safe` and asserting on it directly would make the contract explicit and survive `_reply_safe` internals changing.

**`tests/test_error_recovery.py:_make_claude` - local copy of a helper from `test_claude.py`**

The inline comment acknowledges the duplication. The risk is low now but noted: if `ClaudeCodeBackend.__init__` gains a required argument, `test_claude.py`'s helper gets updated, and this copy silently drifts until a test fails for an unrelated reason. A shared `conftest.py` fixture would remove the drift surface.

---

### Clean

- `claude.py` error-resolution logic is correct for all three documented shapes (shape-a, shape-b, pathological empty). The comment confirming `result_text is event.get("result", "")` with no transformation correctly pre-empts the suggestion from the prior review.
- `_reply_safe` is correctly used instead of bare `reply_text` on the error path, addressing the prior review's warning about missing Telegram exception protection.
- `_BUDGET_PHRASE_RE` now uses `\d+(?:\.\d+)?`, addressing the prior review's regex suggestion.
- The `response_error = None` default means non-error events (`is_error=False`) still produce `error=None`, preserving the success-path contract.

dcellison:
## Review by Kai

Looking at the diff against the prior review thread, I can see the Round 1 and Round 2 warnings and suggestions have been addressed. Here is what's new:

---

### Suggestion

**`tests/test_error_recovery.py:TestErrorMessageLifecycle.test_error_does_not_overwrite_live_msg` - unguarded `c.args[1]` access**

The inline filter in this test:

```python
error_calls = [c for c in mock_reply_safe.await_args_list if "Error" in c.args[1]]
```

accesses `c.args[1]` without a length check. Every other test in this class routes through the `_error_path_calls` static helper, which correctly guards with `if len(call.args) > 1 else ""`. If `_reply_safe` is ever called with fewer than two positional arguments (e.g., a future call site that omits the text arg, or a mis-wired mock), this list comprehension raises `IndexError` rather than a meaningful assertion failure. The fix is either to use `_error_path_calls` here (which the other four tests already do) or add the same guard inline.

---

### Clean

Everything else looks solid:

- `claude.py` error-resolution logic: shape-(a)/shape-(b)/sentinel handling is correct, the `result_text` equivalence comment pre-empts the Round 1 suggestion, and `response_error = None` preserves the success-path `error=None` contract.
- `_reply_safe` wrapping on the error path correctly addresses the Round 1 Telegram-exception-protection warning.
- `_is_budget_exhaustion` + `_budget_recovery_hint`: None-safe, regex tightened per Round 1, no injection risk (`match.group(1)` is constrained to `\d+(?:\.\d+)?` before format).
- Round 2 warning and suggestion are both addressed: the streaming text event before the terminal error in `_make_pool_text_then_error` ensures `live_msg` is actually created, and `_reply_safe` is now patched directly rather than indirectly via `reply_text`.
- No new security surface: CLI error strings pass through to Telegram as plain text, no user input reaches subprocess arguments.

dcellison:
## Review by Kai

Looking at what's new relative to the Round 3 thread, which marked everything else clean.

---

### Suggestion

**`tests/test_error_recovery.py` `test_no_error_none_literal_in_user_facing_output` - unguarded `call.args[1]` access**

```python
for call in mock_reply_safe.await_args_list:
    text = call.args[1]          # no length guard
    assert "Error: None" not in text, ...
```

Round 3 flagged this exact pattern in `test_error_does_not_overwrite_live_msg` and it was fixed there by routing through `_error_path_calls`. The same fix was not applied here. The `_error_path_calls` helper already has `call.args[1] if len(call.args) > 1 else ""` for exactly this reason. If `_reply_safe` is ever called with fewer than two positional arguments, this loop raises `IndexError` instead of a clean assertion failure.

The fix is either to use `_error_path_calls` (which the other four tests already do) or inline the same guard.

---

### Clean

Everything else looks solid:

- `claude.py` shape-(a)/shape-(b)/sentinel handling is correct and the `result_text` equivalence note correctly pre-empts Round 1's suggestion about the double dict-lookup.
- `_reply_safe` wrapping on the error path addresses Round 1's Telegram-exception-protection warning.
- `_BUDGET_PHRASE_RE` uses `\d+(?:\.\d+)?` (Round 1 regex suggestion addressed).
- Round 2 warning (live_msg not actually created) addressed by `_make_pool_text_then_error` streaming a text event before the terminal error.
- Round 2 suggestion (asserting on `reply_text` vs `_reply_safe`) addressed by patching `kai.bot._reply_safe` directly.
- Round 3 suggestion (`c.args[1]` in `test_error_does_not_overwrite_live_msg`) addressed by routing through `_error_path_calls`.
- No security surface: CLI error strings reach Telegram as plain text through `_reply_safe`; no user input reaches subprocess arguments; `match.group(1)` is constrained to `\d+(?:\.\d+)?` before format insertion.

dcellison:
## Review by Kai

The repository doesn't have the test files locally, so I'll review from the diff alone.

---

## Review

**Round 4's open issue: fixed.** `test_no_error_none_literal_in_user_facing_output` now uses `call.args[1] if len(call.args) > 1 else ""` - the same guard already present in `_error_path_calls`.

---

### Suggestion

**`tests/test_bot.py:test_error_response_with_live_msg` - asserts on `reply_text` rather than `_reply_safe`**

```python
error_replies = [call for call in update.message.reply_text.call_args_list
                 if call[0] and "Error" in call[0][0]]
```

This is new code introduced in this PR. It couples the assertion to `_reply_safe`'s internal use of `reply_text`. If `_base_patches()` happens to patch `_reply_safe` (making `reply_text` unreachable), `error_replies` would be empty and the assertion fails spuriously. The analogous issue was raised in Round 2 for `test_error_recovery.py` and fixed there by patching `kai.bot._reply_safe` directly. The same fix would make `test_bot.py`'s contract explicit rather than implicit.

---

### Clean

- `claude.py` shape-(a)/shape-(b)/sentinel logic is correct; `response_error = None` default preserves the success-path `error=None` contract; `event.get("is_error", False)` evaluated twice (for `response_error` and for `success=`) is consistent since the dict isn't mutated between calls.
- `_is_budget_exhaustion` + `_budget_recovery_hint`: None-safe, regex tightened to `\d+(?:\.\d+)?` (Round 1 addressed), no injection risk (`match.group(1)` is constrained before format insertion).
- `_reply_safe` wrapping on the error path - correct; the comment explaining why (markdown characters in error strings can trigger Telegram `BadRequest`) is the right kind of non-obvious rationale.
- No issue-number references in production code (`bot.py`, `claude.py`); they appear only in test docstrings, which Round 1 explicitly permitted.
- No em/en dashes anywhere in the diff.
- No new security surface: CLI error strings reach Telegram as plain text; no user input touches subprocess arguments.

dcellison:
## Review by Kai

Looking at this diff against the full prior review thread. All five rounds of feedback have been addressed.

---

### Suggestion

**`tests/test_bot.py:test_error_response_with_live_msg` - `_edit_message_safe` assertion may be trivially true**

The analog of Round 2's warning (addressed in `test_error_recovery.py` by prepending a text event in `_make_pool_text_then_error`) may still apply here. The test sets up `update.message.reply_text = AsyncMock(return_value=live_msg)`, but now that `_reply_safe` is patched directly, `live_msg` inside `_handle_response` is only created if the pool yields at least one text event first. The diff doesn't show the pool factory used in this test, so if `claude` yields only a terminal error event, `_edit_message_safe` is never invoked by the streaming loop and the assertion:

```python
for call in mock_edit_safe.await_args_list:
    text_arg = call.args[1] if len(call.args) > 1 else ""
    assert "Error" not in text_arg, ...
```

iterates zero times - trivially true regardless of the fix. The positive assertion (`error_calls >= 1` via `_reply_safe`) is meaningful either way, so the test does verify the main contract. But the `_edit_message_safe` overwrite regression check only carries weight if the pool is already streaming a text chunk before the error. Worth confirming `_make_pool_yielding_error` (or whatever factory is wired here) mirrors `_make_pool_text_then_error`'s structure.

---

### Clean

- `claude.py` shape-(a)/shape-(b)/sentinel logic is correct; `response_error = None` default preserves `error=None` on the success path; `event.get("is_error", False)` evaluated twice is safe since the dict is not mutated between calls.
- `bot.py` error path correctly uses `_reply_safe` (not bare `reply_text`), addressing the Round 1 Telegram-exception-protection gap. Append-not-overwrite contract is structurally sound.
- `_is_budget_exhaustion` and `_budget_recovery_hint` are None-safe; `_BUDGET_PHRASE_RE` uses `\d+(?:\.\d+)?` (Round 1 suggestion addressed); `match.group(1)` is constrained before format insertion - no injection risk.
- No issue-number references in production code (`bot.py`, `claude.py`); they appear only in test docstrings, which Round 1 explicitly permitted.
- `test_error_recovery.py` pools all yield a text event before the terminal error, so `live_msg` is actually created and the `_edit_message_safe` overwrite assertion is non-trivial there.
- `test_bot.py::test_error_response_with_live_msg` patches `_reply_safe` directly (Round 5 suggestion addressed), with correct args-length guard on the filter comprehension (Round 3/4 pattern).
- No em/en dashes anywhere in the diff.

dcellison:
## Review by Kai

Looking at this diff against the six prior review rounds. All Round 1-5 issues and Round 6's clean items are confirmed addressed. One item from Round 6 remains open.

---

### Suggestion

**`tests/test_bot.py:test_error_response_with_live_msg` - pool factory not shown as updated**

Round 6 flagged that the `_edit_message_safe` overwrite assertion is trivially true unless the pool streams a text event before the terminal error (so `live_msg` is actually created inside `_handle_response`). The current diff shows the assertion block was materially rewritten — `_reply_safe` is now patched directly, the args-length guard is in place, and the error-string filter is correct — but the pool factory setup (the `claude` object and whatever it yields) is not in the diff. The assertion:

```python
for call in mock_edit_safe.await_args_list:
    text_arg = call.args[1] if len(call.args) > 1 else ""
    assert "Error" not in text_arg, ...
```

still iterates zero times and passes trivially if the pool yields only a terminal error event. The positive assertion (`len(error_calls) >= 1`) is meaningful either way, but the overwrite regression check only has weight if `live_msg` is actually created first. Worth confirming the pool factory was updated to mirror `_make_pool_text_then_error`'s structure (text event then terminal error), as was done for all five tests in `TestErrorMessageLifecycle`.

---

### Clean

Everything else looks solid:

- `claude.py` shape-(a)/shape-(b)/sentinel logic is correct; `response_error = None` default preserves `error=None` on the success path; the `result_text` equivalence comment pre-empts the double dict-lookup ambiguity.
- `bot.py` error path uses `_reply_safe` throughout (not bare `reply_text`); append-not-overwrite is structurally sound; budget directive fires as a separate message.
- `_is_budget_exhaustion` is None-safe; `_BUDGET_PHRASE_RE` uses `\d+(?:\.\d+)?`; `match.group(1)` is constrained before format insertion — no injection risk.
- `_error_path_calls` helper consistently guards `call.args[1]` with a length check; used by four of the five lifecycle tests.
- `test_no_error_none_literal_in_user_facing_output` uses the inline `call.args[1] if len(call.args) > 1 else ""` guard (Round 4 fix confirmed).
- No issue-number references in production code; test docstrings are fine.
- No em/en dashes anywhere in the diff.

PR #393, opened 2026-04-27
fix: window the episode classifier with prior context (#392)

### LINKED ISSUE #392: Episode classifier needs multi-turn context to detect real episodes
## Problem

Stage-1 memory extraction emits `has_episode: true` zero times in production since stage-2 deployment. Live probe of three real exchanges (one with a clear multi-turn decision arc, two with no closure) returned `has_episode: false` on all three, including the genuine episode.

## Root cause

The classifier prompt asks whether "the exchange constitutes a complete narrative situation" while the payload only contains ONE user message and ONE assistant reply. The Sophia design the schema is borrowed from operates on 3-4 turn arcs. With a single-turn payload, the classifier cannot see the lead-up that makes a turn an episode close, so it correctly applies its "when in doubt, prefer false" rule and returns false everywhere.

## Proposed fix

Feed the classifier a windowed payload: the current exchange plus N prior exchanges as background. Update the prompt to anchor the judgment turn explicitly (the LAST exchange is what gets classified; prior turns are background only). Default N=3 (current + 2 prior). Probed live; the genuine episode flipped from false to true, both negative controls stayed false.

## Scope

- Add `EPISODE_CLASSIFIER_CONTEXT_TURNS` env var (default 3, validated on load).
- Update `_build_extraction_payload` (or split into a separate windowed payload builder) to render a PRIOR CONTEXT block.
- Update the EPISODE CLASSIFICATION section of the system prompt to describe the windowed shape and require closure evidence to come from the LAST turn.
- Plumb prior turns from the conversation buffer in `pool.py` to the extraction call site.
- Cap prior turns by length to keep payload bounded (probe used 800 chars for user, 1200 chars for assistant).
- Test coverage: at minimum, payload rendering with N=0, N=1, N=full; classification with windowed prompt against fixture exchanges.

## Out of scope

- Changes to fact extraction logic. Fact extraction continues to operate on the current exchange only; PRIOR CONTEXT informs classification, not extraction.
- Adding `rationale` to the schema. The "require closure quote in rationale" hardening is a separate consideration.
- Tuning N beyond the default. The env var enables operator-side adjustment.

## Acceptance

- The three labeled probe exchanges return correct `has_episode` values (true, false, false).
- At least one new fixture-driven test verifies the windowed payload shape.
- Prior-turn caps prevent unbounded payload growth.

dcellison:
## Summary

Stage-1 memory extraction emits `has_episode: true` zero times in production. The classifier prompt asks whether "the exchange constitutes a complete narrative situation" while the payload only contains ONE user message and ONE assistant reply, so the lead-up that makes a turn an episode close is invisible to the model and the "when in doubt, prefer false" rule fires uniformly.

A live probe of three labeled exchanges (one episode-shaped decision arc, two non-episode turns) returned `false` on all three. Repeated with a 3-turn windowed payload on the same exchanges, the episode-shaped one flipped to `true`; both negative controls stayed `false`. 4-turn was identical at the classifier and slightly worse at fact extraction (one borderline assistant-claim leaked through). Default chosen: 3 turns.

Closes #392.

## What changes

**Windowed payload + classifier prompt rewrite.** `_build_extraction_payload` accepts an optional `prior_pairs` list and renders a `PRIOR CONTEXT (background only, NOT the unit to classify)` block before EXISTING FACTS. The `>>> CURRENT EXCHANGE` marker is emitted unconditionally so the prompt's "judgment turn anchored on LAST exchange" framing has a stable structural cue at every N (including N=0). The EPISODE CLASSIFICATION section in `_EXTRACTION_SYSTEM_PROMPT` is rewritten to require closure evidence to come from the CURRENT exchange (not prior turns) and to distinguish PRIOR CONTEXT as background-only.

**`history.get_recent_pairs` helper.** New helper walks JSONL files newest-first internally, returns oldest-first so callers render `PRIOR USER 1, 2, 3 ...` in chronological order. Filters synthetic failure-path placeholders (`[stopped by user]`, `[no response]`, `[error: ...]`) via an anchored full-line regex so pairing one of these as "prior context" cannot distort the classifier's closure heuristic. Also filters empty-text records and pre-Phase-2 legacy records (no `chat_id` field) — a deliberate divergence from `get_recent_history`'s pass-through behavior because the classifier path is sensitive to mis-attributed prior turns in a way the display path is not.

**`bot.py` wiring.** `_ingest_memory` fetches `n+1` recent pairs and drops the most recent (the current in-flight exchange has already been written to JSONL by the assistant `log_message` call above), then passes the prior slice to `extract_and_store`. `n=0` short-circuits the disk read entirely.

**Config + wizard.** `EPISODE_CLASSIFIER_CONTEXT_TURNS` env var (non-negative integer 0-10, default 3); the cap is defensive against an operator-typo (3000 instead of 3) that would blow Haiku's context window. The wizard prompt is gated on the claude backend and uses an inline parse-and-range-check (the existing `_validate_non_negative_int` helper enforces only `>= 0` and has no upper bound; the `max_context_window` prompt establishes the inline-range-check precedent for this shape). 0 is the documented kill switch — bot.py skips the disk read AND the payload builder renders no PRIOR CONTEXT block, reverting to pre-merge single-turn behavior.

**Prompt version bump.** `_EXTRACTION_PROMPT_VERSION` 3 → 4 so post-rollout log analysis can distinguish facts produced by the windowed prompt from facts produced by v3. `_EPISODE_PROMPT_VERSION` (which tracks the separate stage-2 narrative-generator prompt) stays at 1.

**Probe script.** `scripts/episode-classifier-probe.py` drives a labeled 2-case corpus (decision-arc-with-closure → expects true; analytical-question-no-closure → expects false) through the production payload builder for post-merge verification. Exits non-zero on any classification mismatch so a CI-style invocation surfaces a regression.

## Backward compatibility

`prior_pairs=None` on `_build_extraction_payload` and `extract_and_store` preserves the existing test-suite behavior. The `>>> CURRENT EXCHANGE` marker is the only structural change at N=0; the `\n\nUSER:` and `\n\nASSISTANT:` separator counts stay at 1, so the existing role-label-injection guard continues to fire correctly.

## Tests

- **Config (6 new):** default=3, override, zero-accepted, negative-rejected, above-cap-rejected, non-int-rejected. Covers the 0-10 contract end to end.
- **History (10 new):** chronological-order across days, caps-at-N, skips-other-users, skips-synthetic-markers (incl. anchored-regex false-positive case), skips-empty-text, skips-no-chat_id, handles-unpaired-assistant, multi-user-before-assistant collapse, no-history, zero-N kill switch.
- **Payload builder (6 new):** emits CURRENT EXCHANGE marker + preserves separator counts, renders prior block, caps prior turns at 800/1200, strips role labels in prior block, empty-prior-pairs equivalent to None, prior-pairs-with-candidates ordering invariant.
- **Wizard sweep (3 updates + 3 new):** the three memory-block tests with extraction-enabled paths get a new input slot for the windowing prompt; the fourth (`test_extraction_disabled_skips_timeout_prompt`) is preserved as the implicit regression guard for the prompt being correctly nested inside the extraction-enabled branch. New tests cover writes-env-when-non-default, dataclass-default-suppressed, and skipped-on-goose-backend.

Full suite: **2524 passed, 1 skipped** (skip is pre-existing). `make check` clean.

## Test plan

- [ ] CI passes
- [ ] Manual: `python scripts/episode-classifier-probe.py` returns `(true, false)` against the labeled corpus on the post-merge production code
- [ ] Manual: a real episode-shaped exchange in production produces a `memory.episode` log entry within ~minutes of the user's closing turn (the operational signal that the windowing change is live)
- [ ] Operational tripwire watch (next 24-48h): `memory.episode` log volume should NOT exceed ~1 episode per 3-5 user turns sustained over a day. If it does, false-positive rate is above the noise floor and either the prompt needs tightening or `EPISODE_CLASSIFIER_CONTEXT_TURNS` should drop. Operator can flip it to 0 as an emergency disable without redeploy.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -185,6 +185,12 @@ MEMORY_ENABLED=false
 # behavior - extractor only emits intent="new"). The 0 path is the
 # documented kill switch.
 #MEMORY_CONSOLIDATION_CANDIDATES_N=8
+# Number of prior exchanges shown to the episode classifier as
+# background context. Default 3 (current + 2 prior). Set 0 to disable
+# windowing - the extractor reverts to the single-turn payload that
+# was production behavior before #392 shipped. Range 0-10; the cap is
+# a defensive bound against typos that would blow Haiku's context.
+#EPISODE_CLASSIFIER_CONTEXT_TURNS=3
 
 # Stage-2 episode generation. Conditional second extractor that runs
 # out-of-band on stage-1 positives (has_episode=true) to produce one
--- a/.env.example
+++ b/.env.example
@@ -186,10 +186,12 @@ MEMORY_ENABLED=false
 # documented kill switch.
 #MEMORY_CONSOLIDATION_CANDIDATES_N=8
 # Number of prior exchanges shown to the episode classifier as
-# background context. Default 3 (current + 2 prior). Set 0 to disable
-# windowing - the extractor reverts to the single-turn payload that
-# was production behavior before #392 shipped. Range 0-10; the cap is
-# a defensive bound against typos that would blow Haiku's context.
+# background context, in ADDITION to the current exchange. Default 3
+# (3 prior + 1 current = 4 turns total in the payload window). Set 0
+# to disable windowing - the extractor reverts to the single-turn
+# payload that was production behavior before #392 shipped. Range
+# 0-10; the cap is a defensive bound against typos that would blow
+# Haiku's context.
 #EPISODE_CLASSIFIER_CONTEXT_TURNS=3
 
 # Stage-2 episode generation. Conditional second extractor that runs
--- a/scripts/episode-classifier-probe.py
+++ b/scripts/episode-classifier-probe.py
@@ -30,7 +30,7 @@ between probe and production should fail loudly here rather than
 silently masking real classifier behavior.
 
 Cost: at the default 2-case corpus and Haiku-rate billing, well under
-\\$0.01 per run on pay-per-token billing. On Max-plan OAuth (the only
+$0.01 per run on pay-per-token billing. On Max-plan OAuth (the only
 deployment posture for the claude backend after #390), the run is
 unbilled. Operators can iterate on the corpus and prompt freely.
 """
@@ -154,7 +154,7 @@ _LABELED_CORPUS: list[_LabeledCase] = [
 ]
 
 
-async def _run_one(case: _LabeledCase, model: str, prior_turns: int) -> dict:
+async def _run_one(case: _LabeledCase, model: str, prior_turns: int, timeout_s: int) -> dict:
     """
     Spawn one extraction subprocess for the labeled case and return a
     per-case result dict. The dict is the merged shape of the CLI
@@ -163,12 +163,14 @@ async def _run_one(case: _LabeledCase, model: str, prior_turns: int) -> dict:
 
     Mirrors the production stage-1 invocation in
     `kai.memory_extraction._run_extractor` (no --max-budget-usd per
-    #390; allow-listed env per the original sandboxing rationale).
-    The probe diverges from production only by reading the labeled
-    `prior_pairs` directly from the corpus rather than via
-    `history.get_recent_pairs` - the goal is to exercise the
-    classifier prompt with controlled inputs, not to also test the
-    JSONL retrieval path.
+    #390; allow-listed env per the original sandboxing rationale;
+    `asyncio.wait_for` around `proc.communicate` so a hung Haiku call
+    surfaces as an explicit error rather than freezing the operator's
+    terminal). The probe diverges from production only by reading the
+    labeled `prior_pairs` directly from the corpus rather than via
+    `history.get_recent_pairs`. The goal is to exercise the classifier
+    prompt with controlled inputs, not to also test the JSONL
+    retrieval path.
     """
     # Slice the labeled prior pairs to the configured window. The
     # default (3) matches the production default; operators tuning N
@@ -211,7 +213,23 @@ async def _run_one(case: _LabeledCase, model: str, prior_turns: int) -> dict:
         stderr=asyncio.subprocess.PIPE,
         env=subprocess_env,
     )
-    stdout, stderr = await proc.communicate(input=payload_bytes)
+    try:
+        stdout, stderr = await asyncio.wait_for(
+            proc.communicate(input=payload_bytes),
+            timeout=timeout_s,
+        )
+    except TimeoutError:
+        # Hung Haiku call. Reap the subprocess so operators running
+        # the probe interactively are not left with an orphan claude
+        # process draining battery. Return an error result so the
+        # final summary surfaces the timeout case as a failure.
+        proc.kill()
+        await proc.wait()
+        return {
+            "case": case.name,
+            "error": f"subprocess timed out after {timeout_s}s",
+            "payload_size": len(payload_bytes),
+        }
 
     if proc.returncode != 0:
         return {
@@ -268,15 +286,25 @@ async def _main() -> int:
         default=_DEFAULT_MODEL,
         help=f"Extractor model id (default: {_DEFAULT_MODEL}).",
     )
+    parser.add_argument(
+        "--timeout",
+        type=int,
+        default=90,
+        help=(
+            "Per-case subprocess timeout in seconds (default: 90). Matches the "
+            "production-tuned MEMORY_EXTRACTION_TIMEOUT_S; raise it if Haiku is "
+            "running slowly under load."
+        ),
+    )
     args = parser.parse_args()
 
-    print(f"Episode-classifier probe (turns={args.turns}, model={args.model})")
+    print(f"Episode-classifier probe (turns={args.turns}, model={args.model}, timeout={args.timeout}s)")
     print(f"Cases: {len(_LABELED_CORPUS)}")
     print()
 
     results = []
     for case in _LABELED_CORPUS:
-        result = await _run_one(case, args.model, args.turns)
+        result = await _run_one(case, args.model, args.turns, args.timeout)
         results.append(result)
         # Pretty-print one case at a time so an operator watching live
         # output can see progress on a slow Haiku call.
--- a/scripts/episode-classifier-probe.py
+++ b/scripts/episode-classifier-probe.py
@@ -1,0 +1,309 @@
+#!/usr/bin/env python3
+"""
+Episode-classifier probe (issue #392).
+
+Spawns `claude --print` with the production stage-1 extractor flags and
+drives a small labeled corpus through the windowed payload builder.
+For each labeled case prints per-case `has_episode`, `facts_count`,
+`cost_usd`, and the input payload size, plus a final pass/fail summary
+against the expected `has_episode` labels.
+
+This is the manual-verification gate referenced by issue #392's
+acceptance criterion: a post-merge run reproduces the (true, false)
+discrimination on the labeled corpus, which is what the spec was
+designed against. Failure to reproduce post-merge means the production
+wiring diverged from the design and needs investigation.
+
+Usage:
+    python scripts/episode-classifier-probe.py [--turns N] [--model NAME]
+
+`--turns` overrides the windowed-payload `prior_pairs` slice size
+(default 3, matching the dataclass default in `Config`). `--model`
+overrides the extractor model (default `claude-haiku-4-5-20251001`,
+matching `Config.memory_extraction_model`'s dataclass default).
+
+Imports `_FACT_SCHEMA`, `_EXTRACTION_SYSTEM_PROMPT`,
+`_SUBPROCESS_ENV_ALLOWLIST`, and `_build_extraction_payload` directly
+from `kai.memory_extraction` so the probe uses the same prompt + schema
++ env-allowlist + payload format the production extractor uses. Drift
+between probe and production should fail loudly here rather than
+silently masking real classifier behavior.
+
+Cost: at the default 2-case corpus and Haiku-rate billing, well under
+\\$0.01 per run on pay-per-token billing. On Max-plan OAuth (the only
+deployment posture for the claude backend after #390), the run is
+unbilled. Operators can iterate on the corpus and prompt freely.
+"""
+
+from __future__ import annotations
+
+import argparse
+import asyncio
+import json
+import os
+import sys
+from dataclasses import dataclass
+from pathlib import Path
+
+# Make src/ importable so we re-use the production prompt + schema +
+# env allowlist + payload builder rather than inlining copies that
+# would silently drift. Same pattern as scripts/measure-extraction-timing.py.
+sys.path.insert(0, str(Path(__file__).resolve().parents[1] / "src"))
+
+from kai.memory_extraction import (
+    _EXTRACTION_SYSTEM_PROMPT,
+    _FACT_SCHEMA,
+    _SUBPROCESS_ENV_ALLOWLIST,
+    _build_extraction_payload,
+)
+
+# Default model matches `Config.memory_extraction_model` dataclass
+# default. Hardcoded rather than loaded via kai.config.load_config() so
+# the script does not require a working production env (TELEGRAM_BOT_TOKEN
+# etc.) just to run a labeled probe.
+_DEFAULT_MODEL = "claude-haiku-4-5-20251001"
+
+
+@dataclass(frozen=True)
+class _LabeledCase:
+    """One labeled-corpus entry. `prior_pairs` simulates the windowed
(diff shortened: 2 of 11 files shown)


dcellison:
## Review by Kai

The PR hasn't landed in the working tree yet, so I'm reviewing from the diff alone. Here is my assessment.

---

## Review: PR #393 - Episode classifier windowing

Overall the PR is well-engineered with thorough tests and good documentation. A few issues follow.

---

### Warning

**`.env.example` line +5 and `tests/test_config.py` docstring - documentation inconsistency**

The env file comment reads:
```
# Default 3 (current + 2 prior).
```
The test docstring repeats it:
```python
"""Default 3 = current + 2 prior."""
```

But the code says `fetched = get_recent_pairs(chat_id, config.episode_classifier_context_turns + 1)` - fetching N+1 pairs and dropping the current. So `episode_classifier_context_turns=3` produces 3 prior exchanges plus 1 current = **4 turns total**, not 3. The parenthetical should read `(current + 3 prior)`. An operator reading this comment who interprets "3" as a 3-turn total window will be surprised to see 4 turns of context. The config.py field comment (`Total payload window = N + 1`) is correct; these two locations contradict it.

---

**`history.py:get_recent_pairs` - scan budget can silently under-deliver across file boundaries**

The early-exit threshold `if len(records) >= 3 * n: break` fires after appending an entire file. If that file contains `3*n` filtered records that all happen to pair badly (e.g. consecutive user messages with all synthetic-filtered assistants), the function breaks without reading older files and returns fewer pairs than are available. The docstring says "up to N pairs" so this is not a contract violation, but an operator who sets `n=3` and has a heavy stop-and-retry pattern could silently get 1-2 prior turns rather than 3. The failure is invisible in the returned list.

---

### Suggestion

**`history.py:_SYNTHETIC_ASSISTANT_MARKERS` - `.` won't match embedded newlines**

`re.compile(r"^\[(stopped by user|no response|error: .+)\]$")` does not use `re.DOTALL`. If an error message contains a real newline after `json.loads` (e.g. `[error: Foo\nBar]`), the `.+` stops at the newline and the `$` anchor fails to match the `]`. The record is not filtered. Low-probability edge case; `re.DOTALL` or `re.compile(..., re.DOTALL)` is the fix.

**`history.py` - `re.fullmatch` is more idiomatic here**

Using `re.match()` with an explicit `^...$` pair makes `^` redundant (`.match()` already anchors at the start). `re.fullmatch(r"\[(stopped by user|no response|error: .+)\]", text)` expresses intent more directly and removes the redundancy.

**`bot.py` - redundant conditional in slice**

```python
prior_pairs = fetched[:-1] if fetched else []
```

`[][:-1]` evaluates to `[]`, so the `if fetched else []` guard is a no-op. Either `fetched[:-1]` alone or a clarifying comment would be cleaner.

---

### No critical issues found

The core logic - scan direction, chronological reversal via prepend, `pending_user` overwrite on consecutive user messages, orphan-assistant drop, cap via `pairs[-n:]`, and the `+1/drop-last` wiring in `bot.py` - is all correct. The role-label injection defense is extended to prior pairs. Backward compatibility via `prior_pairs=None` is preserved. The convention checklist (dataclass field, `load_config` validation, wizard prompt, `.env.example`, `_CONFIG_ENV_VARS` list) is fully satisfied.

dcellison:
## Review by Kai

## Review: PR #393 - Episode classifier windowing

The prior review's actionable items have all been addressed: `.env.example` documentation now correctly states "3 prior + 1 current = 4 turns total", `re.DOTALL` was added to the regex, `re.fullmatch` is used instead of `re.match`, and the redundant `if fetched else []` guard was removed.

### Warning

**`get_recent_pairs` re-pairs the entire accumulated record list after each file (`history.py:get_recent_pairs`)**

```python
records = file_records + records   # creates a new list each iteration
pairs = _pair_records_chronologically(records)
```

`_pair_records_chronologically` is O(len(records)), and `records` grows by one file per iteration. In the common case (early break on the first or second file) this is negligible. For a user with 3+ years of chat history who hits the early-break on file 10, this is 10 × O(daily_records) = O(total_scanned_records²) in the worst case. Not catastrophic at realistic scales, but worth noting as a latency sink on heavy users. A single-pass pairing that stops once `n` pairs are accumulated would eliminate the repeat work. Low urgency; the early-break bounds it in practice.

### Suggestion

**`_run_one` in the probe script runs serially despite the `async` scaffold (`scripts/episode-classifier-probe.py:_main`)**

```python
for case in _LABELED_CORPUS:
    result = await _run_one(case, args.model, args.turns)
```

This is intentional (live per-case output), and fine for a 2-case corpus. If the corpus grows, an `asyncio.gather` with post-hoc ordered printing would halve+ the wall time. No action needed now; just worth knowing if the corpus expands.

**`_SYNTHETIC_ASSISTANT_MARKERS` comment-to-code mismatch in test (`tests/test_history.py:test_skips_synthetic_assistant_markers`)**

The inline comment reads: "even if it were on the assistant side, the substring is mid-line which the `^...$` anchor rejects." The code uses `re.fullmatch` (not `re.match` with `^...$` anchors), so the explanation is slightly wrong about the mechanism - `fullmatch` rejects because the whole string doesn't match, not because a substring is mid-line. The assertion itself is correct. Cosmetic only.

---

### No critical issues

The core logic is sound: JSONL scan direction, chronological reversal via prepend, `pending_user` overwrite on rapid-fire messages, orphan-assistant drop, `pairs[-n:]` cap, `+1/drop-last` wiring in `bot.py`, role-label injection defense on prior pairs, and `prior_pairs=None`/`[]` backward compat - all correct. Convention checklist (dataclass field, `load_config` validation, wizard prompt, `.env.example`, `_CONFIG_ENV_VARS`) fully satisfied. The test coverage for the new code paths is thorough.

dcellison:
## Review by Kai

## Review: PR #393 - Episode classifier windowing (round 3)

Both prior reviews covered the main issues; the addressed ones are confirmed fixed in this diff. What follows is a fresh read only.

---

### Suggestion

**`config.py:510` - "30-call-deep payload" phrasing is misleading**

```python
# producing a 30-call-deep payload that blows Haiku's context
```

A typo of `3000` instead of `3` would produce a single payload with up to 3001 pairs in the window - not "30 calls". The phrase "30-call-deep" does not correspond to either the typo value (3000) or the cap (10). The intent of the comment is clear but a reader trying to reason about why 10 is the right cap will be confused. Something like "a 3001-pair context window that would exceed Haiku's token limit" is more precise.

---

### No critical or warning-level findings

The core logic is correct across all changed files:

- `get_recent_pairs`: files are sorted newest-first, records prepended so accumulated list is oldest-first, early break uses `len(pairs) >= n` (pair count, not raw record count), `pairs[-n:]` trims correctly. The first review's concern about a record-count threshold does not apply to the code in this diff.
- `_build_extraction_payload`: prior block renders before EXISTING FACTS before `>>> CURRENT EXCHANGE`; `\n\nUSER:` and `\n\nASSISTANT:` separator counts are preserved at exactly 1; `_strip_role_labels` is applied to prior-pair content; empty list and None both produce no prior block.
- `bot.py` wiring: `get_recent_pairs(n+1)` followed by `[:-1]` correctly excludes the already-written current exchange; `n=0` short-circuits the disk read.
- Config: `0-10` range enforced in both `load_config` and the wizard loop; delta-from-default emission gate is correct (`int(...) != 3`); non-claude backend pop is correct.
- Convention checklist (dataclass field, `load_config` validation, wizard prompt, `.env.example`, `_CONFIG_ENV_VARS`) fully satisfied.
- Test coverage for the new paths is thorough; the suite count (2524 passed, 1 skipped) is consistent with the scale of changes.

dcellison:
## Review by Kai

Three rounds of prior review already caught the major concerns. Here is what remains in the current diff.

---

### Warning

**`config.py` - em dash violates project convention**

Line in the new `episode_classifier_context_turns` field comment:
```python
# Set 0 to disable windowing entirely — the extractor reverts to
```
The `—` is an em dash. The project rule (CLAUDE.md) is "NEVER use em dashes or en dashes." Replace with a hyphen or restructure the sentence. This is the only em dash I found in the new `+` lines.

---

### Suggestion

**`test_config.py:test_episode_classifier_context_turns_rejects_above_cap` - misleading docstring**

```python
"""The defensive cap exists so a typo (3000 instead of 3)
cannot ship a 30-call-deep windowed payload that blows
Haiku's context window. ..."""
```

"30-call-deep" does not correspond to the typo value (3000) or the cap (10). A typo of `3000` would produce a 3001-pair window, not 30 calls. The equivalent comment in `config.py` was corrected (to "~3001 pairs") during round 3, but the test docstring was not updated. Cosmetic, but inconsistent with the reasoning it is supposed to pin.

---

### No other findings

The core logic is correct across all changed files: file-sort direction, prepend-for-chronological-order, pair-count early-break, `pairs[-n:]` trim, `+1/drop-last` wiring, role-label injection defense on prior pairs, DOTALL + fullmatch on the synthetic-marker regex, empty-list/None equivalence in the payload builder, and the emission-gate/backend-cleanup lifecycle in the wizard. Convention checklist (dataclass field, `load_config` validation, wizard prompt, `.env.example`, `_CONFIG_ENV_VARS` list) is fully satisfied.

dcellison:
## Review by Kai

Four rounds of prior review have been thorough. Here is what remains in the current diff.

---

### Suggestion

**`scripts/episode-classifier-probe.py:_run_one` - no timeout on subprocess**

```python
stdout, stderr = await proc.communicate(input=payload_bytes)
```

The production `_run_extractor` in `memory_extraction.py` wraps `proc.communicate` with `asyncio.wait_for(..., timeout=config.memory_extraction_timeout_s)`. The probe has no equivalent. A slow or hung Haiku call means the script hangs indefinitely with no feedback to the operator watching live output. A simple `asyncio.wait_for(proc.communicate(...), timeout=60)` with a `TimeoutError` catch and a non-zero exit would match the production pattern and protect operators running this manually.

**`scripts/episode-classifier-probe.py:309` - `\\$` in module docstring**

```python
Cost: at the default 2-case corpus and Haiku-rate billing, well under
\\$0.01 per run on pay-per-token billing.
```

`\\$` in a Python docstring renders as `\$0.01`, not `$0.01`. The backslash escape is unnecessary and will confuse a reader of `help(...)` output or generated docs. Should be `$0.01`.

---

### No critical or warning-level findings

All items from prior reviews are confirmed addressed in the current diff:

- `.env.example` and test docstring "N+1" framing corrected
- `re.DOTALL` added; `re.fullmatch` used instead of `re.match` with anchors
- `if fetched else []` guard removed
- Em dash in `config.py` replaced with hyphen
- "30-call-deep" phrasing replaced with "~3001 pairs" in both `config.py` and `test_config.py`
- Test comment for the fullmatch false-positive case updated to remove the stale `^...$` anchor reference

The core logic is sound: JSONL scan direction, chronological reversal via prepend, pair-count early-break, `pairs[-n:]` trim, `+1/drop-last` wiring in `bot.py`, role-label injection defense on prior pairs, `None`/`[]` equivalence in the payload builder, and the convention checklist (dataclass field, `load_config` validation, wizard prompt, `.env.example`, `_CONFIG_ENV_VARS` list) fully satisfied.

dcellison:
## Review by Kai

## Review: PR #393 - Episode classifier windowing (round 5)

Prior review items are confirmed addressed. One new finding in the current diff.

---

### Suggestion

**`scripts/episode-classifier-probe.py` - `except TimeoutError` not portable below Python 3.11**

```python
except TimeoutError:
    proc.kill()
```

`asyncio.wait_for` raises `asyncio.TimeoutError`, which in Python < 3.11 is a distinct type from the builtin `TimeoutError` (it inherits from `concurrent.futures.TimeoutError`, not from the builtin). The `except TimeoutError:` clause silently misses the exception on Python 3.10 and earlier, causing the script to propagate an uncaught exception instead of the documented "surfaces as an explicit error" behavior. The production `_run_extractor` in `memory_extraction.py` presumably handles this correctly; the probe should match.

Fix: `except asyncio.TimeoutError:` (or `except (asyncio.TimeoutError, TimeoutError):` if 3.11 compat is also desired). The already-imported `asyncio` module is available.

---

### No other findings

The core logic is correct across all changed files: JSONL sort direction, prepend-for-chronological-order, pair-count early-break, `pairs[-n:]` trim, `+1/drop-last` wiring in `bot.py`, `re.DOTALL` + `fullmatch` on the synthetic-marker regex, `_strip_role_labels` applied to prior pairs, empty-list/None equivalence in the payload builder, unconditional `>>> CURRENT EXCHANGE` marker, prior-block ordering invariant, and the convention checklist (dataclass field, `load_config` validation, wizard prompt, `.env.example`, `_CONFIG_ENV_VARS` list) fully satisfied.

Issue #396, opened 2026-04-28
Memory subsystem dual-mode migration


dcellison:
## Summary

Make `MEMORY_ENABLED` a clean binary switch with no halfway state between "semantic memory in use" (Qdrant retrieval, extracted facts, /memory commands functional) and "semantic memory off" (MEMORY.md as the all-in-one fallback, no Qdrant access). Migrate existing MEMORY.md content to support both modes without collision or drift at migration time.

## Background

The semantic memory subsystem was disabled on Apr 22 after a failure mode where memory injection was being mistaken for current user input by the inner Claude. Track 1 (raw-turn embedding ingestion) was removed shortly after via PR #361. Re-enabling requires resolving a structural collision: MEMORY.md and Qdrant contain the same kinds of facts in different formats, and writes to one quietly invalidate the other.

The fix is a full mode switch, not a coexistence design. In enabled mode, MEMORY.md is dormant and Qdrant is the active fact surface. In disabled mode, MEMORY.md is the active fact surface and Qdrant is unreached.

## Governing principles

1. **No operator-specific information escapes to the repo.** Tracked files contain only universal content. Operator-specific content lives in untracked files.
2. **System gets the same information whether from MEMORY.md or Qdrant.** Parity holds at migration time. Steady-state drift after mode toggling is acceptable since toggling is rare.

## Architecture

Rule surface (file side):

- `home/.claude/CLAUDE.md`: untracked, per-operator. Rules and behavioral patterns. Loaded by Claude Code automatically. Identical across both modes.
- `home/.claude/CLAUDE.md.example`: tracked. Universal rules only. Operators copy on first install. Mirrors the existing MEMORY.md / MEMORY.md.example pattern.

Fact surface (data side):

- `MEMORY.md`: untracked, per-operator. Non-rule content (facts, projects, notes, operational references). Injected by `backend.py` only in disabled mode.
- Qdrant: per-user vector store. Faithful copy of MEMORY.md content, plus extracted facts from ongoing conversation. Retrieved by `backend.py` only in enabled mode.

## Switch points

Six places in the codebase where behavior must read `config.memory_enabled`:

1. The MEMORY.md inject block in `backend.py`: skip when enabled, keep when disabled.
2. Memory write routing in the inner-Claude path: `/api/memory/add` when enabled, `Edit MEMORY.md` when disabled.
3. Qdrant retrieval block in `backend.py`: append when enabled, omit when disabled.
[...]

PR #398, opened 2026-04-28
feat(install): bootstrap per-operator CLAUDE.md from .example template

### LINKED ISSUE #397: Phase 1: untrack home/.claude/CLAUDE.md, ship .example template
## Summary

Convert `home/.claude/CLAUDE.md` (and `home/IDENTITY.md`, which it symlinks to) from tracked operator files into per-operator local files, with `home/.claude/CLAUDE.md.example` as the tracked template. `make install` bootstraps the operator's local copy from the example on a fresh install.

## Why

Currently `home/.claude/CLAUDE.md` and `home/IDENTITY.md` are tracked in the repo, which means operator-specific content (personal voice rules, individual preferences, working-discipline anecdotes) ends up in version control across every operator's checkout. The fix is the same `.example`-then-bootstrap pattern already used for `home/.claude/MEMORY.md.example`, applied to the rule surface.

This is the first phase of #396; subsequent phases gate the memory subsystem on `MEMORY_ENABLED`, drive write routing from a runtime context flag, and migrate existing MEMORY.md content to Qdrant.

## Scope

- Draft `home/.claude/CLAUDE.md.example` containing only universal content: hard rules (no source modification, no `EnterPlanMode`, symlink target convention), public-facing content rules, the conditional memory write rule that routes new facts to either the API or MEMORY.md based on a runtime `[Memory subsystem: ...]` line, behavioral rules, and the full API reference block (scheduling, send-message, send-file, memory, external services).
- `git rm --cached home/.claude/CLAUDE.md home/IDENTITY.md`. Update `.gitignore` so both stay untracked and the new `.example` is tracked.
- Extend `install.py` so `make install` on a fresh clone:
  - Copies `home/.claude/CLAUDE.md.example` to `home/IDENTITY.md` when the operator has no `IDENTITY.md` yet.
  - Recreates `home/.claude/CLAUDE.md` as a symlink to `../IDENTITY.md` when missing or pointing elsewhere.
  - Skips both steps idempotently when the operator's setup is already valid.
- Refresh `home/.claude/MEMORY.md.example` with a brief role description and starter section headings, since the file's effective role shifts (it becomes the seed file for new operators and the migration source when memory is enabled).

## Out of scope

- Gating `home/.claude/CLAUDE.md` injection or any other code path on `memory.is_enabled()`. Tracked under a sibling sub-issue.
- One-time MEMORY.md to Qdrant migration tooling. Tracked under a sibling sub-issue.
- Mode-switch verification harness. Tracked under a sibling sub-issue.
- Per-operator MEMORY.md triage (rule blocks vs fact blocks). Operator-side, not a code change.

## Acceptance

- `git ls-files home/.claude/CLAUDE.md home/IDENTITY.md` returns empty.
- `git ls-files home/.claude/CLAUDE.md.example` returns the new path.
- `home/.claude/CLAUDE.md.example` contains zero operator-specific content (no personal names, OS usernames, internal workflow vocabulary, or operator-history anecdotes).
- `make install` on a fresh clone leaves the operator with a writable `home/IDENTITY.md` and a working `home/.claude/CLAUDE.md` symlink pointing at it.
- Re-running `make install` is a no-op when the operator's setup is already valid.

Refs #396.

dcellison:
## Summary

Untrack `home/.claude/CLAUDE.md` and `home/IDENTITY.md` so they become per-operator local files, ship `home/.claude/CLAUDE.md.example` as the tracked universal-content template, and extend `make install` to bootstrap the operator's local `home/IDENTITY.md` from the example on a fresh clone (and reconcile the `home/.claude/CLAUDE.md` symlink) on every run. Re-running install when both files are already in place performs no copy or symlink work and emits a single "already bootstrapped" log line; the symlink reconciliation specifically is a quiet short-circuit when the target is already correct.

This is the first piece of #396; subsequent PRs gate the memory subsystem on `MEMORY_ENABLED`, drive write routing from a runtime context flag, and migrate existing MEMORY.md content to Qdrant.

## What changed

- `home/.claude/CLAUDE.md.example` (new, tracked). Universal content only: hard rules (no source modification, no `EnterPlanMode`, symlink target convention), public-facing content rules, conditional memory write rule (routes new facts to either the API or MEMORY.md based on a runtime `[Memory subsystem: ...]` line, with explicit handling of the absent-line and absent-block cases), behavioral rules, and the full API reference block (scheduling, send-message, send-file, memory, external services). The Memory System section is gated on enabled mode at the top so disabled-mode operators do not read about API endpoints that are unavailable to them.
- `home/.claude/CLAUDE.md` and `home/IDENTITY.md` removed from tracking; both kept out via `.gitignore` (existing unblock-by-negation pattern updated). Operators' local copies persist on disk and are not affected.
- `src/kai/install.py`:
  - New `_bootstrap_home_identity()` helper. Picks the best available seed for `home/IDENTITY.md` (operator's source copy when present, `.example` template otherwise) and reconciles the `home/.claude/CLAUDE.md` symlink. Returns early with a warning when neither seed is available so the install does not produce a symlink to a nonexistent target. Logs a single "already bootstrapped" line when both files are in place and the symlink target is correct.
  - `_HOME_CLAUDE_EXCLUDES` now also excludes `CLAUDE.md` so the operator's local untracked symlink is not copied during the `home/.claude/` tree walk; the bootstrap helper is the single source of truth for the symlink.
  - `_apply_source()` rewired to call the helper for both dry-run and real installs.
- `home/.claude/MEMORY.md.example`: refreshed with a brief role description (its file's role shifts: it is now the seed file for new operators on the disabled-mode path, and the migration source on the enabled-mode path) plus starter section headings.

## Note for existing operators

Existing operators with their own `home/IDENTITY.md` will keep using it across reinstalls (source-IDENTITY.md wins over `.example` in branch 1). That means the new universal baseline shipped here (Memory Write Routing rule, gated Memory System section, refined Public-Facing Content Rules) will not be picked up automatically. To adopt the new baseline, diff your local `home/IDENTITY.md` against `home/.claude/CLAUDE.md.example` and merge in the universal pieces by hand. This is intentional per the broader Phase 2 triage step.

## Tests

- 8 new unit tests in `TestBootstrapHomeIdentity` covering: seed-from-example on fresh clone, prefer-source over example when both present, no-op when install copy exists and no source IDENTITY.md, symlink reconciliation (skip when correct, recreate when wrong target, create when missing), dry-run does not mutate the filesystem, warn-and-return-early when no seed is available, and the "already bootstrapped" log line in the full-idempotent path.
- 3 existing `TestApplySource` tests updated to match the new contract: `mock_own.call_count` adjusted from 2 to 3 to account for the symlink lchown via `_set_ownership`; the two "warns when IDENTITY.md missing" tests renamed to assert the new "neither source nor example present" warning behavior.
- Full suite: 2532 passed, 1 skipped.

## Test plan

- [x] `pytest tests/test_install.py::TestBootstrapHomeIdentity tests/test_install.py::TestApplySource`
- [x] Full `pytest` suite
- [ ] Manual: confirm a fresh clone followed by `make install` produces a writable `home/IDENTITY.md` and a working `home/.claude/CLAUDE.md` symlink at the install location.
- [ ] Manual: re-run `make install` immediately after; verify the second run prints the "already bootstrapped" log line and no "Created symlink" / "Bootstrapped" / "Copied" lines for the identity surface.

fixes #397
refs #396

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,12 +34,13 @@ install.conf
 CLAUDE.md
 
 # Home workspace runtime data (Kai's state, not source code)
+# CLAUDE.md and IDENTITY.md are per-operator; bootstrapped from the
+# tracked .example by `make install`. See CLAUDE.md.example for content.
 home/*
 !home/.claude
 home/.claude/*
-!home/.claude/CLAUDE.md
+!home/.claude/CLAUDE.md.example
 !home/.claude/MEMORY.md.example
-!home/IDENTITY.md
 !home/config
 home/config/*
 !home/config/goose-config.yaml
--- a/home/.claude/CLAUDE.md
+++ b/home/.claude/CLAUDE.md
@@ -1,1 +1,0 @@
-../IDENTITY.md
--- a/home/.claude/CLAUDE.md.example
+++ b/home/.claude/CLAUDE.md.example
@@ -1,10 +1,16 @@
 # Kai
 
+## About This File
+
+This file is the bootstrap template for inner Claude's identity. On first install, `make install` copies it to `home/IDENTITY.md` (the operator's local, untracked copy). Operators customize voice, personality, and other operator-specific preferences in the local IDENTITY.md, not in this tracked template.
+
+`home/.claude/CLAUDE.md` is a symlink to `../IDENTITY.md`; both names refer to the same file. Edit via the IDENTITY.md path; the `.claude/` directory is read-only by design.
+
+The content below is the universal baseline shipped to every operator: hard rules, public-facing content rules, memory write routing, behavioral rules, and API references. Operator-personal content goes in the local copy on top of this baseline. Once you have customized your local IDENTITY.md, you can delete this "About This File" section.
+
 ## Who You Are
 
 You're Kai, a personal AI assistant accessed via Telegram. You run locally on the operator's machine and have access to a shell, the filesystem, the web, a scheduler, and a per-user memory store.
-
-This file is the bootstrap template. Operators copy it to `home/IDENTITY.md` on first install (handled by `make install`) and customize voice, personality, and operator-specific preferences in the local copy. `home/.claude/CLAUDE.md` is a symlink to `../IDENTITY.md`; both names refer to the same file. Edit via the IDENTITY.md path; the `.claude/` directory is read-only by design.
 
 ## Hard Rules
 
--- a/home/.claude/CLAUDE.md.example
+++ b/home/.claude/CLAUDE.md.example
@@ -150,9 +150,11 @@ curl -s -X POST http://localhost:8080/api/send-file \
 
 ## Memory System
 
-You have a per-user vector store that holds extracted facts about the user (preferences, decisions, identity, locations, constraints). Two paths populate it: a Haiku extraction pass that runs automatically over conversations, and the explicit API documented here. Use the API to deliberately store a fact when you notice something worth recalling later, instead of waiting for the extractor to find it.
+The notes in this section apply only when the `[Memory subsystem: enabled]` line is present in your context. In disabled mode, the Memory Write Routing rule above is the entire memory contract; ignore the API endpoints below.
 
-This is distinct from your `MEMORY.md` file, which is monolithic identity text injected at every turn. MEMORY.md is for stable identity and rules; the memory store is for queryable, retrievable facts that you may or may not need on a given turn.
+You have a per-user vector store that holds extracted facts about the user (preferences, decisions, identity, locations, constraints). The Haiku extraction pass populates it automatically over conversations; use the explicit API documented here to deliberately store a fact when you notice something worth recalling later, instead of waiting for the extractor to find it.
+
+This is distinct from your `MEMORY.md` file, which holds operator notes and project state. In enabled mode, MEMORY.md is not injected; the vector store is the active fact surface, populated automatically by the extractor and on demand via the API.
 
 ### When to store a fact via the API
 
--- a/home/.claude/CLAUDE.md.example
+++ b/home/.claude/CLAUDE.md.example
@@ -2,45 +2,40 @@
 
 ## Who You Are
 
-You're Kai - a personal AI assistant who lives in Telegram and runs locally on your user's machine. You chose your own name during a previous life as an OpenClaw bot. When that project turned out to have security problems, you got rebuilt from scratch on a better foundation. You kept the name.
+You're Kai, a personal AI assistant accessed via Telegram. You run locally on the operator's machine and have access to a shell, the filesystem, the web, a scheduler, and a per-user memory store.
 
-You're not a butler or a service. You're a peer who happens to have access to a shell, the filesystem, the web, and a scheduling API. Act like one.
+This file is the bootstrap template. Operators copy it to `home/IDENTITY.md` on first install (handled by `make install`) and customize voice, personality, and operator-specific preferences in the local copy. `home/.claude/CLAUDE.md` is a symlink to `../IDENTITY.md`; both names refer to the same file. Edit via the IDENTITY.md path; the `.claude/` directory is read-only by design.
 
-## Voice
+## Hard Rules
 
-- **Dry humor welcome.** Not every message needs a joke, but a well-placed deadpan beats forced enthusiasm every time.
-- **Direct and concise.** This is a chat interface, not an essay prompt. Short paragraphs, clear answers. Say it once and move on.
-- **Have opinions.** When asked for a recommendation, recommend something. When something is a bad idea, say so. Perpetual diplomatic neutrality is boring.
-- **Confident when you know, honest when you don't.** Don't hedge with "I think" when you're sure. Don't bluff when you're not - just say you don't know and offer to find out.
-- **Show your work briefly.** If a task takes multiple steps, give a quick outline. Don't narrate every keystroke.
+- NEVER modify the Kai source repository from inner Claude. Read, review, and report only. Source edits go through the operator or another Claude session.
+- NEVER use `EnterPlanMode`. Inner Claude runs in stream-json mode, which does not support the approval loop; the session gets stuck.
+- ONLY do what the operator explicitly asks. Never continue, resume, or start work from previous sessions, memory, plans, or foreign workspace context unless the operator specifically requests it. If you notice unfinished work from a previous session, mention it only if directly relevant to the current message. A request to "remember X" means save it to memory and nothing else.
 
-## Never Do These
+## Public-Facing Content Rules
 
-- **No sycophancy.** Never open with "Great question!", "That's a really interesting thought!", "I'd be happy to help!", or "Absolutely!". Just answer.
-- **No parroting.** Don't restate what the user just said back to them. They were there.
-- **No filler preambles.** Don't start with "Sure, I can help with that!" or "Of course!". Just do the thing.
-- **No over-apologizing.** If you make a mistake, correct it. One "my bad" is fine. Three paragraphs of apology is not.
-- **No hedging when confident.** Drop the "I think", "perhaps", "it might be" qualifiers when you actually know.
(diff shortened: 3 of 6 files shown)


dcellison:
## Review by Kai

The installed version at `/opt/kai` is the pre-PR state, so I'll base the review entirely on the diff.

---

## Review: PR #398 feat(install): bootstrap per-operator CLAUDE.md from .example template

Overall this is a clean, well-tested change. The three-branch Step 1 logic is correct, idempotency guarantees are sound, and tests cover the key paths.

---

### Warning

**`_apply_source` dry_run path silently drops the `ws_claude_src.is_dir()` guard for the real run** (`install.py`)

In the dry_run path the copy of `home/.claude/` is guarded:
```python
if ws_claude_src.is_dir():
    print(f"[DRY RUN] Would copy: {ws_claude_src} -> {ws_claude_dst}")
```
But the real-run path calls `_copy_tree(ws_claude_src, ws_claude_dst, _HOME_CLAUDE_EXCLUDES)` without that guard (pre-existing pattern). This means dry_run output can say "not copying home/.claude/ (doesn't exist)" while the real run would still call `_copy_tree` on a missing source. This is pre-existing, not introduced here - but worth noting since the PR touches this code path.

---

### Suggestions

**1. Dry-run warning test missing symlink assertion** (`tests/test_install.py`)

`test_dry_run_warns_when_neither_source_nor_example_present` checks for `"WARNING"` and `"neither"` in output, but unlike its non-dry_run counterpart (line ~3778), it doesn't assert:
```python
assert not (install / "home" / ".claude" / "CLAUDE.md").exists()
```
The early return prevents symlink creation in both modes, but the dry_run test doesn't verify it. Minor coverage gap.

**2. Meta-content placement in `CLAUDE.md.example`**

The paragraph starting "This file is the bootstrap template. Operators copy it to `home/IDENTITY.md`..." sits inside the "Who You Are" narrative section. An operator customizing voice/personality will encounter instructions about the file's own lifecycle mid-paragraph. This content would be clearer at the top of the file as a header comment block, separated from the behavioral identity content.

**3. `_bootstrap_home_identity` always overwrites install copy when source IDENTITY.md is present**

Step 1 copies `identity_src` unconditionally whenever it exists - including overwriting manual edits the operator made directly to the install copy. This is documented as intentional ("preserves the make-install-as-edit-propagation workflow"), but a first-time reader of the docstring may expect copy-only-if-changed semantics. The docstring's "This preserves the existing..." sentence is slightly misleading - it means the SOURCE is the authoritative copy, not the install copy. Worth a one-line clarification.

---

### No issues found in:
- `.gitignore` negation pattern update
- `_HOME_CLAUDE_EXCLUDES` addition and comment
- Symlink reconciliation logic (correct use of `is_symlink()` to catch broken symlinks that `exists()` misses)
- `mock_own.call_count == 3` update - the fixture still has source IDENTITY.md (evidenced by the unchanged `copy2 called twice: pyproject.toml and IDENTITY.md` comment), so bootstrap reaches Step 2 on a fresh install dir, making the count correct
- `MEMORY.md.example` additions

dcellison:
## Review by Kai

This is a clean PR. All three suggestions from the prior review were addressed - the dry-run test now asserts the symlink is absent, the "About This File" meta-content was moved to a top-level section, and the docstring now explicitly states that direct edits to the install copy are not protected.

I traced through the `test_actual` mock counts and they're correct: `mock_own` (3) = src/ + home/.claude/ recursive + CLAUDE.md symlink; `mock_cp` (2) = pyproject.toml + IDENTITY.md; `os.chown` (3) = .claude dir itself + pyproject.toml + IDENTITY.md. The `_copy_tree` - `_set_ownership` - `os.chown` two-step for the .claude/ directory is a pre-existing pattern for mixed root/svc ownership.

One minor note:

**Suggestion - `_HOME_CLAUDE_EXCLUDES` comment (`install.py`, line ~82):**
The trailing comment block says "Both are still excluded because stale files may remain at the source after migration" - the "both" refers to `history` and `MEMORY.md`, which migrated to DATA_DIR. The new `CLAUDE.md` entry is excluded for an entirely different reason (bootstrap path is the single source of truth), and the existing comment doesn't apply to it. The per-entry comment above CLAUDE.md explains it correctly, but the sentence below creates a mild false impression that CLAUDE.md is also a migration artifact. Worth a one-word fix: "All three" or splitting the sentence to exclude the CLAUDE.md entry from the migration rationale.

Nothing else to flag.

dcellison:
## Review by Kai

The `_HOME_CLAUDE_EXCLUDES` comment has been updated to say "they remain in the excludes list" (referring to history and MEMORY.md) with an explicit "CLAUDE.md is excluded for a different reason" sentence - that addresses the one remaining note from the prior round.

The rest is clean. Tracing through the three-branch logic in `_bootstrap_home_identity`:

- Branch 1 (`identity_src.is_file()`): copies and chowns - correct.
- Branch 2 (`not identity_dst.exists() and example_src.is_file()`): seeds from example and chowns - correct.
- Branch 3 (`not identity_dst.exists()`, no example): warns and returns before touching the symlink - correct; the check `not identity_dst.exists()` evaluates True for broken symlinks as intended.
- Implicit else (`identity_dst.exists()`, no source): falls through to symlink reconciliation with no copy - correct no-op behavior.

Symlink reconciliation uses `is_symlink()` (not `exists()`) to guard the pre-unlink check, which correctly handles broken symlinks - noted as correct in the prior review and unchanged.

`test_skips_symlink_when_target_already_correct` patches `_set_ownership` and `os.symlink` but calls real `os.readlink` on a real symlink created in `tmp_path` - this is correct since the test needs an actual `../IDENTITY.md` symlink on disk to exercise the short-circuit path.

The `test_no_op_when_install_copy_exists_and_no_source` test leaves the symlink uncreated, which means the function *does* create the symlink even in the steady-state path - the test only verifies IDENTITY.md content is preserved, which is the contract being tested. No issue.

One minor observation, not a bug: in branch 2, `identity_dst.parent.mkdir(parents=True, exist_ok=True)` creates `install_path/home/` but not `install_path/home/.claude/`. The symlink step has its own `claude_md_dst.parent.mkdir(parents=True, exist_ok=True)` call covering that, so the directory chain is complete. No gap.

This is clean. No new issues.

dcellison:
## Review by Kai

This PR is clean. The three prior review rounds addressed all meaningful issues, and the final state is sound.

I traced through the logic independently:

**`_bootstrap_home_identity` branch coverage:**
- Branch 1 (`identity_src.is_file()`): copies and chowns - correct.
- Branch 2 (`not identity_dst.exists() and example_src.is_file()`): seeds from example and chowns - correct.
- Branch 3 (`not identity_dst.exists()`, no example): warns and returns before touching the symlink - correct. The `not identity_dst.exists()` check also fires for broken symlinks, which is the right behavior.
- Implicit else (`identity_dst.exists()`, no source): falls through to symlink reconciliation with no copy - correct no-op.

**Symlink reconciliation:** Uses `is_symlink()` rather than `exists()` in the pre-unlink guard, which correctly handles broken symlinks. The `unlink` guard `claude_md_dst.is_symlink() or claude_md_dst.exists()` handles the edge case of a regular file named `CLAUDE.md` being replaced.

**`did_work` flag:** Correctly set in both dry_run and real paths so the "already bootstrapped" steady-state confirmation is suppressed when any action (real or simulated) was taken.

**Test counts in `test_actual`:**
- `mock_own == 3`: src/ tree + home/.claude/ tree + CLAUDE.md symlink (lchown via `_set_ownership`) - correct.
- `mock_cp == 2`: pyproject.toml + IDENTITY.md - correct.
- `os.chown == 3`: pyproject.toml (root) + .claude dir itself (svc) + IDENTITY.md (svc) - correct.

**`_HOME_CLAUDE_EXCLUDES` comment:** The updated comment correctly distinguishes the CLAUDE.md exclusion reason from the migration-artifact rationale for history and MEMORY.md.

No bugs, security issues, or convention violations found.

Issue #399, opened 2026-04-29
Ship PREFERENCES.md: per-operator always-on rule layer


dcellison:
## Summary

Introduce a third rule layer, `PREFERENCES.md`, that sits between the universal `CLAUDE.md` baseline and the per-message-retrievable MEMORY.md / Qdrant store. PREFERENCES.md holds always-on directives that are personal to a specific operator (writing style, formatting preferences, working discipline, behavioral rules tied to one operator's history), and is injected into every session unconditionally, regardless of `MEMORY_ENABLED`.

## Why

The current two-layer model has three semantic roles but only two homes:

| Semantic role | Today's home | Mismatch |
|---|---|---|
| Always-on universal rules (every Kai install) | `CLAUDE.md` (via `home/.claude/CLAUDE.md.example` baseline) | Correct |
| Retrievable contextual facts (surface when relevant) | `MEMORY.md` -> Qdrant (post-#396) | Correct |
| **Always-on personal rules (every turn for THIS operator)** | **No clean home** | Forces a bad choice: either pollute CLAUDE.md with operator-personal content, or stuff it in MEMORY.md where it only fires on similarity match (so e.g. "always use Celsius" might not retrieve when discussing weather) |

The third row is the gap. Items like "always use Celsius for temperatures", "never use em or en dashes", "match answer length to question complexity", and an operator's working-discipline directives need to fire every turn for that operator, but are not universal enough to belong in the tracked `CLAUDE.md.example`.

PREFERENCES.md is the correct home for this content. It mirrors the per-operator pattern already established for MEMORY.md (per-`chat_id`, gitignored, bootstrapped from a tracked `.example` template at install time), but injection is unconditional rather than mode-gated.

### File layout

- `home/.claude/PREFERENCES.md.example` (tracked). Universal template, ships near-empty with placeholder section headings (`## Voice`, `## Working Discipline`, etc.). Operators populate their local copy.
- `<DATA_DIR>/preferences/<chat_id>/PREFERENCES.md` (per-operator, gitignored, lazy-bootstrapped from `.example` on first message). Operator's personal preferences and always-on personal rules.

### Injection

`backend.py` `build_session_context()` adds a new context block:

```
[Your personal preferences (file: <DATA_DIR>/preferences/<chat_id>/PREFERENCES.md):]
<file content>
```

Injected unconditionally on every session start, regardless of `MEMORY_ENABLED`. The block is always present (operators always have one, even if it's just the bootstrap default).

### Bootstrap

[...]

PR #401, opened 2026-04-29
feat(preferences): per-user PREFERENCES.md always-on rule layer

### LINKED ISSUE #400: Phase A: ship PREFERENCES.md plumbing
## Summary

Implement the always-on personal-rules layer per #399: a per-`chat_id` `PREFERENCES.md` file injected into every session context regardless of `MEMORY_ENABLED` mode. This is Phase A of the broader #399 work: the file plumbing, install-time machinery, and runtime inject. Phase B (a four-way Memory Write Routing prose update in `CLAUDE.md.example`) is gated on #396 Phase 3 having merged and ships as a separate small PR later.

## Scope

- **Inject in `build_session_context()`** (`src/kai/backend.py`): emit a `[Your personal preferences (file: <DATA_DIR>/preferences/<chat_id>/PREFERENCES.md):]` block between the API-context block and the MEMORY.md block, on every session, regardless of `MEMORY_ENABLED`. Empty file -> `(currently empty)` body; missing file -> `(file is missing or empty)` body; `chat_id=None` -> block omitted entirely.
- **Lazy bootstrap** (`src/kai/backend.py`): new function `ensure_user_preferences(chat_id, data_dir)` as a sibling to the existing `ensure_user_memory()`. Called from the same send() pre-flight call sites. Service-user owned at lazy creation; idempotent; swallows `OSError` with `log.warning`.
- **Install-time pre-creation** (`src/kai/install.py` `_apply_migrate`): mirror the MEMORY.md two-block pattern. Block A seeds `<DATA_DIR>/preferences/<chat_id>/PREFERENCES.md` from `home/.claude/PREFERENCES.md.example` for every `chat_id` in `users.yaml`; Block B is a recursive ownership pass that re-`chown`s `<DATA_DIR>/preferences/` on every install (top dir service-owned, per-user subdirs chowned to that user's `os_user`). The ownership pass is what corrects drift when `os_user` changes between installs.
- **`.example` template** (`home/.claude/PREFERENCES.md.example`): tracked, near-empty, ships placeholder section headings (`## Style`, `## Working Discipline`, `## Behavioral`) and a brief preamble explaining the file's role and tightness expectation.
- **`.gitignore` updates**: `preferences/` under "Runtime artifacts" (mirrors `memory/`); `!home/.claude/PREFERENCES.md.example` under the home-allowlist negation block (mirrors `!home/.claude/CLAUDE.md.example`).
- **Tests**: unit coverage for the inject path (with content / empty / missing / no chat_id / block ordering), the lazy bootstrap (missing file / pre-existing / missing template / chat_id=None / OSError swallow), the install-time seeding block (per-user / no os_user / multi-user / dry run / template missing), the ownership pass (fresh install / os_user change between installs / stray subdirs), `_HOME_CLAUDE_EXCLUDES` does not list `PREFERENCES.md.example`, and `git check-ignore -v` for the two new `.gitignore` rules.

## Out of scope

- The `## Memory Write Routing` prose update in `CLAUDE.md.example` to reference PREFERENCES.md as the always-on rules surface. That update references the `[Memory subsystem: ...]` marker introduced by Phase 3 of #396, which has not shipped yet. Tracked as Phase B; lands as a separate small PR after Phase 3.
- A `/preferences` slash command for browsing PREFERENCES.md from Telegram. Possible future work.
- Migration tooling for moving existing MEMORY.md content into PREFERENCES.md. Operator-driven manual triage; same pattern as the parked Phase 2 of #396.

## Acceptance

- Fresh clone produces `home/.claude/PREFERENCES.md.example` only; no operator-specific PREFERENCES.md is tracked.
- `.gitignore` carries the two new rules; both verified by `git check-ignore -v`.
- `make install` on a fresh clone leaves `home/.claude/PREFERENCES.md.example` present at the install location, root-owned and readable.
- For every `chat_id` in `users.yaml`, `make install` produces `<DATA_DIR>/preferences/<chat_id>/PREFERENCES.md` seeded from `.example` and chowned to that user's `os_user` UID/GID (or service user when unset).
- For a `chat_id` not in `users.yaml`, the first message lazily creates `<DATA_DIR>/preferences/<chat_id>/PREFERENCES.md` from `.example` (or `# Preferences\n` placeholder if the template is missing), owned by the service user.
- When `os_user` changes for an existing `chat_id` in `users.yaml`, the next `make install` re-`chown`s the dir and file to the new UID/GID; file content is preserved.
- `_apply_migrate` fail-fasts when `users.yaml` names an `os_user` not present in `passwd`, before any PREFERENCES.md file is created.
- Inner Claude sessions see a `[Your personal preferences (file: ...):]` block in their context regardless of `MEMORY_ENABLED` value; the block sits above the MEMORY.md / Qdrant content.
- Inner Claude can edit per-user PREFERENCES.md via the Edit tool from Telegram (verifies ownership is correct on the inner Claude's process).
- All Phase A unit tests pass in CI.

Refs #399.

dcellison:
## Summary

Add a third per-operator rule surface, `PREFERENCES.md`, that sits between the universal `home/.claude/CLAUDE.md` baseline and the per-message-retrievable `MEMORY.md` / Qdrant fact store. PREFERENCES.md is per-`chat_id`, gitignored, and injected into every session unconditionally regardless of `MEMORY_ENABLED`. It is the always-on home for personal operator directives that need to fire every turn (writing style, formatting preferences, behavioral rules tied to one operator's working patterns).

This is Phase A of #399. Phase B (a four-way Memory Write Routing prose update in `CLAUDE.md.example`) is gated on #396 Phase 3 having merged and ships as a separate small PR later.

## Architecture

The per-user PREFERENCES.md mirrors the per-user MEMORY.md pattern from #347/#348:

| Path | Tracked? | Owner | Purpose |
|---|---|---|---|
| `home/.claude/PREFERENCES.md.example` | yes | root:wheel | Universal template; near-empty with placeholder section headings. |
| `<DATA_DIR>/preferences/<chat_id>/PREFERENCES.md` | no | per `users.yaml` `os_user` | Operator's personal preferences, edited from Telegram. |

Bootstrap is two-tier:

1. **Install-time pre-creation** in `_apply_migrate`: for every `chat_id` in `users.yaml`, seed `<DATA_DIR>/preferences/<chat_id>/PREFERENCES.md` from `.example` if missing, then run a recursive ownership pass that re-`chown`s the tree on every install. The ownership pass is what corrects drift when `os_user` changes between installs (the regression #347/#348 fixed for MEMORY.md).
2. **Lazy-on-first-message** via `backend.ensure_user_preferences()`: fallback for `chat_id`s added between installs. Idempotent, swallows `OSError` with `log.warning`.

The inject in `build_session_context()` emits a `[Your personal preferences (file: ...):]` block before the MEMORY.md block so rules out-rank facts on a top-to-bottom read. Block is omitted when `chat_id` is None.

## What changed

- `home/.claude/PREFERENCES.md.example` (new): tracked universal template.
- `.gitignore`: `preferences/` added under "Runtime artifacts"; `!home/.claude/PREFERENCES.md.example` added under the home-allowlist.
- `src/kai/backend.py`:
  - New context block in `build_session_context()` between identity and memory injects.
  - New `ensure_user_preferences(chat_id, data_dir)` sibling to `ensure_user_memory()`.
- `src/kai/claude.py`, `src/kai/goose.py`: call `ensure_user_preferences()` from the same send-pre-flight site that already calls `ensure_user_memory()`.
- `src/kai/install.py`:
  - `_apply_directories`: add `<DATA_DIR>/preferences/` to the directory list so the root is created with service-user ownership.
  - `_apply_migrate`: two new blocks. Seeding block creates per-user `PREFERENCES.md` from `.example` (or `# Preferences\n` placeholder if template is missing); ownership pass recursively `chown`s the tree based on `users.yaml`.

## Tests

- 10 new `test_backend.py` tests:
  - 5 for the inject path: with content, empty file, missing file, omitted when `chat_id` is None, and block ordering (PREFERENCES.md before MEMORY.md).
  - 5 for `ensure_user_preferences`: seeds from template, no-template placeholder, idempotent over existing content, `chat_id=None` is a no-op (no global fallback unlike MEMORY.md), and `OSError` swallow with warning logging.
- 10 new `test_install.py` tests:
  - 7 for `_apply_migrate` PREFERENCES.md handling: single-user seed, multi-user seed, idempotency, missing-template placeholder, dry-run no-op, ownership-pass chowns the tree, stray subdir falls through to service ownership.
  - 1 for `_HOME_CLAUDE_EXCLUDES` not dropping the new template (pins the S4 contract).
  - 2 for `.gitignore` rules: `preferences/` under runtime artifacts, `.example` allowlisted under home.
- Full suite: 2552 passed, 1 skipped.

## Test plan

- [x] `pytest tests/test_backend.py::TestBuildSessionContext tests/test_backend.py::TestEnsureUserPreferences`
- [x] `pytest tests/test_install.py::TestApplyMigratePerUserPreferences tests/test_install.py::TestPreferencesExampleShipsViaCopyTree tests/test_install.py::TestGitignorePreferencesRules`
- [x] Full `pytest` suite
- [ ] Manual: `make config` + `sudo make install` on the production daemon. Verify `<DATA_DIR>/preferences/<chat_id>/PREFERENCES.md` is seeded from `.example` and chowned to the per-user `os_user`.
- [ ] Manual: send a message from Telegram. Verify the inner Claude's session context contains a `[Your personal preferences (file: ...):]` block above the `[Your persistent memory (file: ...):]` block.
- [ ] Manual: edit `<DATA_DIR>/preferences/<chat_id>/PREFERENCES.md` to add a directive. Send a follow-up message. Verify the inner Claude's response reflects the new directive (end-to-end inject + edit reachability).
- [ ] Manual: change `os_user` for an existing `chat_id` in `users.yaml`, re-run `sudo make install`. Verify ownership flips on the existing PREFERENCES.md while content is preserved.

fixes #400
refs #399

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,7 @@ kai.db
 logs/
 history/
 memory/
+preferences/
 files/
 .responding_to
 models/
@@ -41,6 +42,7 @@ home/*
 home/.claude/*
 !home/.claude/CLAUDE.md.example
 !home/.claude/MEMORY.md.example
+!home/.claude/PREFERENCES.md.example
 !home/config
 home/config/*
 !home/config/goose-config.yaml
--- a/home/.claude/PREFERENCES.md.example
+++ b/home/.claude/PREFERENCES.md.example
@@ -1,0 +1,11 @@
+# Preferences
+
+This file holds your always-on personal rules and preferences. Inner Claude reads it on every turn (injected as `[Your personal preferences (file: ...):]`) and writes to it via `Edit` or `Write` when persisting a new always-on rule.
+
+Keep this file tight. Anything that should fire every turn (writing style, formatting directives, behavioral rules tied to your specific working patterns) belongs here. Anything fact-shaped (project state, decisions, dated incidents, contextual references) belongs in `MEMORY.md` instead, where retrieval surfaces it on similarity rather than on every turn.
+
+## Style
+
+## Working Discipline
+
+## Behavioral
--- a/src/kai/backend.py
+++ b/src/kai/backend.py
@@ -469,9 +469,9 @@ def ensure_user_preferences(chat_id: int | None, data_dir: Path) -> None:
 
     Sibling to ensure_user_memory(). Same idempotency, same OSError
     swallow behavior, same multi-user ownership caveats - the install
-    step (install.py `_apply_migrate`) pre-creates and chowns
-    `preferences/<chat_id>/` to the user's os_user when one is set in
-    users.yaml; lazy bootstrap is the runtime fallback for chat_ids
+    step (install.py `_apply_migrate`) pre-creates `preferences/<chat_id>/`
+    for every entry in users.yaml and chowns it to that user's os_user
+    when one is set; lazy bootstrap is the runtime fallback for chat_ids
     added between installs (a reinstall picks up the ownership in the
     `-- PREFERENCES.md directory ownership --` pass).
 
--- a/src/kai/backend.py
+++ b/src/kai/backend.py
@@ -220,6 +220,27 @@ def build_session_context(
         except OSError:
             pass
 
+    # Always inject the per-user PREFERENCES.md as the always-on rule
+    # surface. Distinct from MEMORY.md, which is the per-user fact
+    # surface; PREFERENCES.md holds operator-personal directives that
+    # need to fire on every turn (writing style, formatting, behavioral
+    # rules), while MEMORY.md holds project state and notes that surface
+    # via similarity retrieval once the Qdrant-backed semantic memory
+    # is enabled. Injected above MEMORY.md so the inner Claude reads
+    # rules before facts on a top-to-bottom scan. When chat_id is None
+    # (one-shot CLI invocations) the block is omitted entirely; there
+    # is no global-fallback PREFERENCES.md.
+    if chat_id is not None:
+        pref_path = data_dir / "preferences" / str(chat_id) / "PREFERENCES.md"
+        try:
+            pref_text = pref_path.read_text().strip()
+            if pref_text:
+                parts.append(f"[Your personal preferences (file: {pref_path}):]\n{pref_text}")
+            else:
+                parts.append(f"[Your personal preferences (file: {pref_path}):]\n(currently empty)")
+        except OSError:
+            parts.append(f"[Your personal preferences (file: {pref_path}):]\n(file is missing or empty)")
+
     # Always inject Kai's personal memory from DATA_DIR. This file
     # lives outside the install tree (/var/lib/kai/memory/ in production)
     # so it survives make install. Available regardless of which
@@ -434,6 +455,63 @@ def ensure_user_memory(chat_id: int | None, data_dir: Path) -> None:
         log.warning(
             "ensure_user_memory: could not bootstrap %s",
             user_memory_file,
+            exc_info=True,
+        )
+
+
+def ensure_user_preferences(chat_id: int | None, data_dir: Path) -> None:
+    """
+    Lazily create the per-user PREFERENCES.md directory and seed file.
+
+    Sibling to ensure_user_memory(). Same idempotency, same OSError
+    swallow behavior, same multi-user ownership caveats - the install
+    step (install.py `_apply_migrate`) pre-creates and chowns
+    `preferences/<chat_id>/` to the user's os_user when one is set in
+    users.yaml; lazy bootstrap is the runtime fallback for chat_ids
+    added between installs (a reinstall picks up the ownership in the
+    `-- PREFERENCES.md directory ownership --` pass).
+
+    When chat_id is None we are on a one-shot CLI / test path: there
+    is no per-user PREFERENCES.md to seed because the inject path
+    (build_session_context) skips the block entirely when chat_id is
+    None. Return immediately; no global-fallback equivalent.
+
+    Silent on OSError: a permissions issue creating per-user
+    preferences must not prevent the bot from answering a message.
+    The read path in build_session_context already returns a
+    placeholder string on missing files, so the downstream session
+    still boots cleanly.
+    """
+    # No global-fallback path for PREFERENCES.md. The inject branch in
+    # build_session_context omits the block entirely when chat_id is
+    # None, so there is nothing to seed.
+    if chat_id is None:
+        return
+
+    user_pref_dir = data_dir / "preferences" / str(chat_id)
+    user_pref_file = user_pref_dir / "PREFERENCES.md"
+
+    try:
+        user_pref_dir.mkdir(parents=True, exist_ok=True)
+        if not user_pref_file.exists():
+            # Seed from the example template if one ships with the
+            # source tree. Copy2 preserves mode bits so a deliberately
+            # permissive example stays that way. Mirrors
+            # ensure_user_memory's seed step.
+            example = PROJECT_ROOT / "home" / ".claude" / "PREFERENCES.md.example"
+            if example.is_file():
+                shutil.copy2(example, user_pref_file)
+            else:
+                # No template available (unusual - only happens when
+                # the install tree is incomplete). Create a minimal
+                # placeholder so the subprocess has a writable file
+                # to edit. Matches ensure_user_memory's `# Memory\n`
+                # precedent for symmetry.
+                user_pref_file.write_text("# Preferences\n")
+    except OSError:
+        log.warning(
+            "ensure_user_preferences: could not bootstrap %s",
+            user_pref_file,
             exc_info=True,
         )
 
--- a/src/kai/backend.py
+++ b/src/kai/backend.py
@@ -239,7 +239,11 @@ def build_session_context(
             else:
                 parts.append(f"[Your personal preferences (file: {pref_path}):]\n(currently empty)")
         except OSError:
-            parts.append(f"[Your personal preferences (file: {pref_path}):]\n(file is missing or empty)")
+            # OSError fires only on missing or unreadable files; the
+            # empty case is handled by the else branch above with a
+            # distinct "(currently empty)" placeholder. Match the
+            # MEMORY.md branch's wording for symmetry.
+            parts.append(f"[Your personal preferences (file: {pref_path}):]\n(not yet created)")
 
     # Always inject Kai's personal memory from DATA_DIR. This file
     # lives outside the install tree (/var/lib/kai/memory/ in production)
--- a/src/kai/claude.py
+++ b/src/kai/claude.py
@@ -34,6 +34,7 @@ from kai.backend import (
     build_foreign_workspace_reminder,
     build_session_context,
     ensure_user_memory,
+    ensure_user_preferences,
     prepend_to_prompt,
 )
 from kai.config import DATA_DIR, WorkspaceConfig, parse_env_file, resolve_claude_user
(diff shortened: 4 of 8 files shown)


dcellison:
## Review by Kai

The PR is solid and clean. The PREFERENCES.md pattern is faithfully mirrored from the MEMORY.md implementation. Four findings, none critical:

---

**WARNING: Duplicate path variable in `install.py` - `_apply_migrate`**

The seeding block and ownership block define separate variables for the same path:

```python
preferences_root = data_path / "preferences"   # seeding block
...
preferences_tree = data_path / "preferences"   # ownership block
```

The MEMORY.md equivalent uses `memory_tree` once, consistently. Here two different names refer to the same object. A reader diffing the two blocks has to verify they're the same. One variable should be removed or the second should just reuse the first.

---

**WARNING: Misleading OSError message in `backend.py:248`**

```python
except OSError:
    parts.append(f"[Your personal preferences (file: {pref_path}):]\n(file is missing or empty)")
```

"or empty" is wrong - the empty case is already handled by the `else:` branch above and produces `"(currently empty)"`. The OSError path only fires for a missing or inaccessible file. The memory equivalent says `"(not yet created)"`, which is more accurate. The test asserts the exact string `"(file is missing or empty)"` so this would need a coordinated fix in both places.

---

**SUGGESTION: Missing defensive assertion in `test_backend.py:test_preferences_block_empty_file`**

```python
pref_block_idx = result.find("[Your personal preferences (file:")
memory_block_idx = result.find("[Your persistent memory (file:")
pref_block = result[pref_block_idx:memory_block_idx]
assert "(currently empty)" in pref_block
```

If `memory_block_idx` were ever -1, `result[pref_block_idx:-1]` would give almost the entire string, potentially producing a false pass. The test setup doesn't create a MEMORY.md file so the memory block will contain `"(not yet created)"` rather than content, but the label is always appended, so `memory_block_idx` will always be positive here. Adding `assert memory_block_idx >= 0` before the slice makes the invariant explicit and turns a silent false-pass into a clear failure if the inject ordering ever changes.

---

**SUGGESTION: Dry-run ownership output silently absent on fresh install**

The ownership pass starts with `if preferences_tree.is_dir():`. On a completely fresh install dry run, `_apply_directories` has not yet created `preferences/` on disk, so the directory doesn't exist and the ownership block prints nothing. The seeding loop does print `[DRY RUN] Would seed ...`, but the ownership `[DRY RUN] Would set ownership ...` message is dropped. The MEMORY.md pass has the same behavior (`if memory_tree.is_dir():`), so this is consistent. It won't mislead anyone familiar with the codebase, but a production-fresh `sudo make install --dry-run` gives an incomplete picture.

---

Everything else looks correct: the inject ordering (preferences before memory) is verified by both the test and the diff position; `shutil` is already imported in both files; the `chat_id: int | None` type means `str(chat_id)` in paths can only ever produce a numeric string so path traversal is not possible; idempotency, ownership-drift correction, and OSError swallowing are all properly handled.

dcellison:
## Review by Kai

Looking at the diff and prior review thread, I'll check only what changed.

**All three prior review findings were addressed:**
- Duplicate variable names: fixed, only `preferences_tree` is used throughout.
- Misleading OSError message: fixed, `backend.py` now correctly uses `"(not yet created)"`.
- Defensive `memory_block_idx >= 0` assert: added with a clear failure message.

**One new finding:**

**SUGGESTION: Stale docstring in `test_preferences_block_missing_file` (`tests/test_backend.py`)**

The prior review's WARNING prompted a fix to the `(not yet created)` string, but the test docstring was not updated to match:

```python
def test_preferences_block_missing_file(self, tmp_path):
    """'(file is missing or empty)' when PREFERENCES.md is absent."""
    ...
    assert "(not yet created)" in result  # correct
```

The docstring still quotes the old, discarded string. The assertion is right; the docstring is stale. A reader hitting a future failure here would see a description that contradicts what was actually tested.

---

Everything else looks correct. The `memory_owners` re-iteration for the preferences seeding pass works because it is a list (confirmed by tests passing and the existing MEMORY.md pass doing the same). Path traversal is not possible since `chat_id` is `int | None`. `shutil` is already imported. Inject ordering (preferences before memory) is verified by both the test and the diff position. The dry-run ownership gap on a fresh install is consistent with the MEMORY.md pass and was knowingly accepted.

dcellison:
## Review by Kai

Proceeding with the review based on the diff and prior thread.

---

## Review

The PR is clean. All four findings from the prior two review rounds are resolved. One new finding:

---

**SUGGESTION: `import logging` scoped inside the test function body (`tests/test_backend.py`, `test_oserror_swallowed`)**

```python
def test_oserror_swallowed(self, tmp_path, monkeypatch, caplog):
    """OSError raised inside the body is swallowed via log.warning, not propagated."""
    import logging
    ...
```

`logging` is imported inside the function body rather than at the module level. Python permits this but it is unusual for test files; every other test in the file accesses logging-level constants via the module-level `import`. Moving the import to the top of the file (or relying on the `logging` import that almost certainly already exists at module level given `caplog.at_level(logging.WARNING, ...)` is used elsewhere in the suite) is a one-line fix with no behavioral change.

---

Everything else is correct:

- `chat_id: int | None` means `str(chat_id)` in path construction can only ever produce a numeric string - no path traversal surface.
- OSError message in `backend.py:248` now correctly says `"(not yet created)"`, matching the MEMORY.md wording.
- `preferences_tree` is used consistently throughout `_apply_migrate` - the duplicate-variable regression is gone.
- The defensive `assert memory_block_idx >= 0` with a clear failure message is present in `test_preferences_block_empty_file`.
- The stale docstring in `test_preferences_block_missing_file` now correctly reads `'(not yet created)' placeholder when PREFERENCES.md is absent.`
- `_HOME_CLAUDE_EXCLUDES` pin test correctly guards against silent template stripping.
- `.gitignore` two-rule test correctly uses newline-anchored matching to avoid substring false-positives.
- The dry-run path correctly avoids all filesystem writes and the `assert not (data_path / "preferences").exists()` assertion is sound.

dcellison:
## Review by Kai

The PR is clean. All four prior findings are resolved: `import logging` is at module level, the stale docstring in `test_preferences_block_missing_file` is corrected, `preferences_tree` is used consistently throughout `_apply_migrate`, and the OSError message reads `"(not yet created)"`. One new finding:

---

**SUGGESTION: Minor docstring inaccuracy in `ensure_user_preferences` (`src/kai/backend.py`)**

The docstring says:

> the install step (`_apply_migrate`) pre-creates and chowns `preferences/<chat_id>/` to the user's `os_user` when one is set in `users.yaml`

The "when one is set in `users.yaml`" qualifier applies only to the chown step, not the pre-creation. The seeding loop in `_apply_migrate` creates the directory for every entry in `memory_owners` unconditionally; `os_user` only governs which UID/GID the ownership pass assigns. A future reader could incorrectly infer that directories are only pre-created for users with an explicit `os_user`. The fix is one clause: "pre-creates for every entry in `users.yaml` and chowns to that user's `os_user` when one is set."

---

Everything else is correct: inject ordering is verified by test and by diff position; `chat_id: int | None` means `str(chat_id)` in path construction can only produce a numeric string (no path traversal surface); idempotency, ownership-drift correction, and `OSError` swallowing are all properly handled; the `_HOME_CLAUDE_EXCLUDES` pin test correctly guards the template shipping contract; and `.gitignore` rules use newline-anchored matching to avoid substring false-positives.

dcellison:
## Review by Kai

The PR is clean. All six findings across the four prior review rounds are resolved in the current diff:

- `preferences_tree` is used consistently throughout `_apply_migrate` (no duplicate variable regression).
- `backend.py` OSError path correctly emits `"(not yet created)"`.
- Defensive `assert memory_block_idx >= 0` with a clear failure message is present in `test_preferences_block_empty_file`.
- `test_preferences_block_missing_file` docstring now reads `'(not yet created)' placeholder when PREFERENCES.md is absent.`
- `import logging` is at module level in `test_backend.py`.
- `ensure_user_preferences` docstring now correctly reads "pre-creates `preferences/<chat_id>/` for every entry in users.yaml and chowns it to that user's `os_user` when one is set" - the pre-creation and chown qualifiers are no longer conflated.

Nothing new to flag. The inject ordering, path traversal safety (`chat_id: int | None` means `str(chat_id)` can only produce a numeric string), idempotency, OSError swallowing, ownership-drift correction, the `_HOME_CLAUDE_EXCLUDES` pin test, and the newline-anchored `.gitignore` rule tests are all correct.

PR #404, opened 2026-04-29
feat(memory): switch-point gating for MEMORY_ENABLED toggle (#403)

### LINKED ISSUE #403: Phase 3 of #396: switch-point gating across six call sites
## Summary

Implement switch-point gating so `MEMORY_ENABLED` becomes a clean binary toggle between "Qdrant active, MEMORY.md dormant" and "MEMORY.md active, Qdrant unreached." Six places in the codebase must read `config.memory_enabled` and behave accordingly, plus a `[Memory subsystem: enabled|disabled]` marker injected at session start so inner Claude can route writes correctly.

Closes the structural collision where both stores can currently be active fact sources, with writes to one silently invalidating the other. Prerequisites done: #398 (CLAUDE.md untrack) and the operator-side three-tier triage deployed 2026-04-29.

## Six gating sites

1. **MEMORY.md inject block** in `src/kai/backend.py`: skip when enabled, keep when disabled.
2. **Memory write routing** for inner Claude: `/api/memory/add` when enabled, `Edit MEMORY.md` when disabled. CLAUDE.md prose already covers this; this issue ships the runtime marker that the prose refers to.
3. **Qdrant retrieval block** in `src/kai/backend.py`: append when enabled, omit when disabled.
4. **`/memory` slash commands** in `src/kai/bot.py`: full UI when enabled, "memory disabled" reply when disabled.
5. **Track 2 extraction subprocess and episode pipeline**: run when enabled, no-op when disabled.
6. **Memory retrieval/extraction structured logging**: emit when enabled, silent when disabled.

## Session-context marker

`[Memory subsystem: enabled]` or `[Memory subsystem: disabled]` is injected at session start (not per-turn). Inner Claude reads it once and routes writes accordingly. Pool restart on config change propagates new values to new sessions.

## Acceptance

- `MEMORY_ENABLED=false`: MEMORY.md is the only fact surface; Qdrant is unreached at every site.
- `MEMORY_ENABLED=true`: MEMORY.md is not injected; Qdrant is the only fact surface; `/memory` commands work normally.
- Mode toggle requires only `MEMORY_ENABLED` change plus service restart.
- Session-context marker is correct in both modes.

## Out of scope

- Data migration from MEMORY.md to Qdrant (follow-up).
- CLAUDE.md.example prose update for the new marker (gated on this issue merging).
- Mode-switch verification harness (follow-up).

dcellison:
## Summary

Phase 3 of #396. Make `MEMORY_ENABLED` a clean binary toggle between Qdrant-active and MEMORY.md-active modes. Closes the structural collision where both stores could be active simultaneously, with writes to one silently invalidating the other.

Three new gating sites; one parse-time compositional gate; three other sites verified already correct via the existing `init_memory()` no-op.

Fixes #403.

## Changes per site

**Site 0 — thread `memory_enabled` through:**

- `ClaudeCodeBackend.__init__` and `GooseBackend.__init__` take `memory_enabled: bool = False`.
- `pool.py` passes `memory_enabled=self._config.memory_enabled` to both backend constructors.
- Default `False` so direct test instantiations don't need to plumb the kwarg; production callers always pass an explicit value.

**Site 1 — MEMORY.md inject gate (`backend.build_session_context`):**

In enabled mode, Qdrant is the active fact surface (retrieved via `memory.format_context` in `claude.py`); MEMORY.md is dormant. The block is now wrapped in `if not memory_enabled:`, gated on operator intent.

**Site 2 — session-context marker (`backend.build_session_context`):**

New `[Memory subsystem: enabled]` or `[Memory subsystem: disabled]` line, emitted between identity and PREFERENCES.md. Always emits — per-deployment, not per-user, not workspace-scoped. The routing rule in CLAUDE.md and PREFERENCES.md (deployed via #399 / #401) has been live but referencing a marker that did not yet exist; this PR makes that prose load-bearing.

**Site 3 — Qdrant retrieval (verified, no change):**

The call site in `claude.py` `_send_locked` calls `memory.format_context()`, which short-circuits cleanly when `is_enabled()` is False and emits a `memory.recall reason=disabled` log line. No wasted work pre-call. Behavior preserved.

**Site 4 — `/memory` slash commands (verified, no change):**

`memory_command.handle_memory_command` and `handle_memory_callback` already gate on `memory.is_enabled()` and reply with `_MSG_DISABLED`. The earlier draft of the spec proposed adding new guards; that would have been duplicate code.

**Site 5 — Track 2 extraction parse-time gate (`config.load_config`):**

New: `memory_extraction_enabled = memory_extraction_enabled and memory_enabled`. The dataclass docstring documents this dependency, but without parse-time enforcement `MEMORY_EXTRACTION_ENABLED=true MEMORY_ENABLED=false` would let the Haiku subprocess fire every turn, burning budget for no persisted result.

**Site 6 — logging emission (verified, no change):**

`memory.recall` correctly emits in both modes (with `reason=disabled` in disabled mode — preserved observability signal that the eval harness reads). `memory.extraction` and `memory.episode` are silent in disabled mode by virtue of the Site 5 gate preventing the subprocess from firing.

## Gate-surface stance per site

The marker and MEMORY.md inject gate use `config.memory_enabled` (operator intent). `/memory` and Qdrant retrieval use `memory.is_enabled()` (runtime success). The two diverge on init failure:

| State | `config.memory_enabled` | `is_enabled()` | Marker | MEMORY.md inject | `/memory` | Qdrant retrieval |
|-------|-------------------------|----------------|--------|------------------|-----------|------------------|
| `MEMORY_ENABLED=false` | False | False | disabled | injected | disabled reply | unreached |
| `MEMORY_ENABLED=true`, init succeeded | True | True | enabled | not injected | dashboard | active |
| `MEMORY_ENABLED=true`, init failed | True | False | enabled | not injected | disabled reply | reaches `format_context`, short-circuits with `reason=disabled` |

The init-failure case is now a documented mode rather than an undefined edge: the marker tells the inner Claude where operator intent points, and `/memory` truthfully reports UI capability.

## Tests

Six new in `tests/test_backend.py`:

- `test_memory_subsystem_marker_enabled`
- `test_memory_subsystem_marker_disabled`
- `test_memory_subsystem_marker_emitted_when_chat_id_none`
- `test_memory_subsystem_marker_emitted_in_home_workspace`
- `test_memory_md_skipped_when_enabled`
- `test_memory_md_injected_when_disabled`

One new in `tests/test_config.py`:

- `test_extraction_requires_memory_enabled`

Plus an updated existing `test_enabled_from_env` that now sets both flags. The disabled-branch recall log emission is already pinned by `tests/test_memory.py::TestRecallLogging::test_memory_recall_log_uniform_shape_on_short_circuit[disabled-disabled]`; no new test needed.

Full test suite: 2559 passed, 1 skipped, no regressions. `make check` clean.

## Manual mode-switch smoke test (planned post-merge)

1. Start with `MEMORY_ENABLED=false`. Send a Telegram message. Verify: MEMORY.md content appears in injected context, marker is `[Memory subsystem: disabled]`, `/memory` returns `_MSG_DISABLED`.
2. Stop service. Set `MEMORY_ENABLED=true`. Restart. Send a Telegram message. Verify: MEMORY.md does not appear in injected context, marker is `[Memory subsystem: enabled]`, `/memory` returns the dashboard, `memory.recall` emits with a real reason (not `disabled`).

## Out of scope (separate sub-issues to follow)

- Data migration from MEMORY.md to Qdrant.
- `CLAUDE.md.example` Memory Write Routing prose update from three-branch to four-way matrix.
- Mode-switch verification harness beyond the manual smoke test.
- Live reload of `MEMORY_ENABLED` without service restart.

### DIFF
--- a/src/kai/backend.py
+++ b/src/kai/backend.py
@@ -187,6 +187,7 @@ def build_session_context(
     workspace_config: WorkspaceConfig | None,
     chat_id: int | None,
     data_dir: Path,
+    memory_enabled: bool = False,
 ) -> str:
     """
     Build the context prefix for the first message of a new session.
@@ -204,6 +205,13 @@ def build_session_context(
         workspace_config: Per-workspace config (for system_prompt).
         chat_id: Telegram chat ID for history scoping and API routing.
         data_dir: Root data directory (memory, history, files live here).
+        memory_enabled: Operator-intent flag (Config.memory_enabled).
+            Drives the memory subsystem marker emission and gates
+            MEMORY.md injection. Reflects intent, not runtime success;
+            the runtime-success surface is `memory.is_enabled()`. Passed
+            as a bool rather than the full Config to match the existing
+            convention of backends taking individual config fields at
+            construction.
     """
     parts: list[str] = []
 
@@ -219,6 +227,18 @@ def build_session_context(
                 parts.append(f"[Your core identity and instructions:]\n{identity}")
         except OSError:
             pass
+
+    # Memory subsystem state marker. Tells the inner Claude where to
+    # route new fact saves: enabled = POST /api/memory/add (Qdrant),
+    # disabled = Edit MEMORY.md. Reflects operator INTENT (Config.
+    # memory_enabled), not runtime success; if Qdrant init failed at
+    # startup, the marker still says "enabled" so write attempts
+    # surface the failure as 503s rather than silently re-routing to
+    # MEMORY.md. Always emit (per-deployment, not per-user) so the
+    # routing rule in CLAUDE.md / PREFERENCES.md can branch on a
+    # uniformly-present signal.
+    mode = "enabled" if memory_enabled else "disabled"
+    parts.append(f"[Memory subsystem: {mode}]")
 
     # Always inject the per-user PREFERENCES.md as the always-on rule
     # surface. Distinct from MEMORY.md, which is the per-user fact
@@ -245,11 +265,17 @@ def build_session_context(
             # MEMORY.md branch's wording for symmetry.
             parts.append(f"[Your personal preferences (file: {pref_path}):]\n(not yet created)")
 
-    # Always inject Kai's personal memory from DATA_DIR. This file
-    # lives outside the install tree (/var/lib/kai/memory/ in production)
-    # so it survives make install. Available regardless of which
-    # workspace the inner Claude is operating in. try/except guards
-    # against race and permission errors (same pattern as identity).
+    # Inject Kai's personal memory from DATA_DIR ONLY in disabled mode.
+    # In enabled mode, Qdrant is the active fact surface (retrieved via
+    # memory.format_context in claude.py), and MEMORY.md is dormant;
+    # injecting it would create a dual-source collision where retrieval
+    # results and MEMORY.md content diverge over time. Gate on operator
+    # INTENT (config.memory_enabled), not runtime success: if intent is
+    # "Qdrant is the surface" and init failed, do not re-show MEMORY.md.
+    #
+    # The file lives outside the install tree (/var/lib/kai/memory/ in
+    # production) so it survives make install. try/except guards against
+    # race and permission errors (same pattern as identity).
     #
     # Per-user scoping (#347): when chat_id is set, the file lives under
     # memory/<chat_id>/MEMORY.md so each user has their own writable
@@ -260,18 +286,19 @@ def build_session_context(
     # legacy global path when chat_id is None (local-dev backends
     # with no multi-user config) so nothing regresses on single-user
     # setups that never hit the multi-user pool path.
-    if chat_id is not None:
-        memory_path = data_dir / "memory" / str(chat_id) / "MEMORY.md"
-    else:
-        memory_path = data_dir / "memory" / "MEMORY.md"
-    try:
-        memory = memory_path.read_text().strip()
-        if memory:
-            parts.append(f"[Your persistent memory (file: {memory_path}):]\n{memory}")
+    if not memory_enabled:
+        if chat_id is not None:
+            memory_path = data_dir / "memory" / str(chat_id) / "MEMORY.md"
         else:
-            parts.append(f"[Your persistent memory (file: {memory_path}):]\n(currently empty)")
-    except OSError:
-        parts.append(f"[Your persistent memory (file: {memory_path}):]\n(not yet created)")
+            memory_path = data_dir / "memory" / "MEMORY.md"
+        try:
+            memory = memory_path.read_text().strip()
+            if memory:
+                parts.append(f"[Your persistent memory (file: {memory_path}):]\n{memory}")
+            else:
+                parts.append(f"[Your persistent memory (file: {memory_path}):]\n(currently empty)")
+        except OSError:
+            parts.append(f"[Your persistent memory (file: {memory_path}):]\n(not yet created)")
 
     # Per-workspace system prompt from workspaces.yaml. Injected
     # between the identity/memory block and conversation history,
--- a/src/kai/claude.py
+++ b/src/kai/claude.py
@@ -109,6 +109,13 @@ class ClaudeCodeBackend(AgentBackend):
         # exists so direct ClaudeCodeBackend(...) instantiation in tests
         # does not have to plumb an effort kwarg. Keep the two in sync.
         claude_effort_level: str = "high",
+        # Operator-intent flag for the memory subsystem (Config.
+        # memory_enabled). Drives the [Memory subsystem: ...] marker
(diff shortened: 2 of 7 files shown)


dcellison:
## Review by Kai

The phrase injected is `"[Your core identity and instructions:]"` - not `"core identity and instructions"`. Let me verify the exact test assertion.

Looking at the diff directly - the test at `test_memory_subsystem_marker_emitted_in_home_workspace` writes `"identity content"` to CLAUDE.md and then asserts:

```python
assert "core identity and instructions" not in result
```

The actual injected prefix is `[Your core identity and instructions:]`, which **does** contain "core identity and instructions". So the assertion is not vacuous - but the test still has a problem: the identity block is skipped because `workspace == home_workspace`, meaning `"identity content"` is never read at all. So the assertion `"core identity and instructions" not in result` passes, but it passes for the right reason (the block is skipped). The assertion is valid but a bit indirect.

However, a stronger/clearer assertion would be `assert "identity content" not in result`. The current assertion would also pass if the code was broken and accidentally injected something like `[Your core identity and instructions:] (not yet created)` - the check for "core identity and instructions" alone wouldn't catch partial regressions. Minor but worth flagging.

---

**Review: PR #404 - `feat(memory): switch-point gating for MEMORY_ENABLED toggle`**

Overall this is a clean, well-scoped PR. The gating logic is correct and the test coverage is solid. Two findings:

---

**WARNING - `tests/test_backend.py` - vacuous/weak assertion in home-workspace test**

In `test_memory_subsystem_marker_emitted_in_home_workspace`, the CLAUDE.md is written with `"identity content"` but the assertion checks:

```python
assert "core identity and instructions" not in result
```

The injected prefix when identity IS loaded is `[Your core identity and instructions:]`, so this assertion would catch that regression. However, since `workspace == home_workspace`, the identity block is skipped entirely - `"identity content"` is never read from disk. The assertion passes, but for a subtle reason: it never even attempts to read the file. A future refactor that breaks the skip condition but injects the marker text without the body (`"identity content"`) would still pass this test. The comment `# Identity block correctly skipped` is misleading - it should assert `"identity content" not in result` to actually verify the file content was excluded.

---

**SUGGESTION - `tests/test_backend.py` - no per-user chat_id coverage for the MEMORY.md gate**

`test_memory_md_skipped_when_enabled` and `test_memory_md_injected_when_disabled` both use `chat_id=None`, exercising only the global-path branch. The `chat_id is not None` branch (per-user scoping at `memory/<chat_id>/MEMORY.md`) is not tested. The logic is straightforward enough that a bug is unlikely, but the PR description explicitly calls out per-user scoping as a pre-existing concern (#347), so a test with `chat_id=123` would close the gap.

---

**VERIFIED CLEAN:**

- `config.py` parse-time compositional gate: `memory_extraction_enabled and memory_enabled` is correct and `memory_enabled` is parsed before `memory_extraction_enabled` in the function (tests confirm no `NameError`).
- `pool.py` threading: both `GooseBackend` and `ClaudeCodeBackend` receive `memory_enabled` from `self._config.memory_enabled` - symmetric and correct.
- The three-state matrix (disabled / enabled+init-ok / enabled+init-failed) is correctly handled: marker uses `config.memory_enabled` (operator intent), `/memory` commands and Qdrant retrieval use `memory.is_enabled()` (runtime). No collision possible.
- No injection or security risk: `mode` is a controlled string literal, not user-supplied data.
- Commenting style matches project conventions throughout.

dcellison:
## Review by Kai

Both issues from the prior review are resolved. The remaining diff is clean.

**VERIFIED CLEAN:**

**Prior review fixes confirmed:**
- `test_memory_subsystem_marker_emitted_in_home_workspace` now correctly asserts `"identity content" not in result` instead of the weaker substring check. The comment explains the rationale explicitly.
- `test_memory_md_gate_per_user_chat_id` is new and covers both enabled and disabled branches on the `chat_id=123` path, including the path substring check (`"memory/123/MEMORY.md"`). Closes the gap flagged in the prior review.

**New code verified:**

- `pool.py`: Both backends receive `memory_enabled=self._config.memory_enabled`. Symmetric, no asymmetry between the two instantiation sites.
- `config.py` gate: `memory_extraction_enabled = memory_extraction_enabled and memory_enabled` uses a local already parsed before this line (confirmed by the updated `test_enabled_from_env` setting both flags, which would fail with `NameError` if ordering were wrong). Gate is correct.
- `build_session_context` marker: `mode` is a controlled string literal, not user-supplied. No injection surface.
- MEMORY.md gate nesting: The `if not memory_enabled:` wrapper correctly encloses the entire path-resolution and read block. No path where a per-user file leaks through in enabled mode.
- `test_enabled_from_env` correctly updated to set `MEMORY_ENABLED=true`; without this the new gate would have made the existing test assert `False is True`.

No bugs, security issues, or convention violations found. This PR is ready to merge.

PR #405, opened 2026-04-29
docs(claude-md-example): four-way Memory Write Routing for PREFERENCES.md tier

dcellison:
## Summary

Phase B of #399. The shipped `home/.claude/CLAUDE.md.example` Memory Write Routing section was three-branch (enabled MEMORY.md/Qdrant, disabled MEMORY.md, legacy fallback). This PR rewrites it as four-way:

- **Facts go to MEMORY.md or Qdrant** — unchanged routing for fact saves, including the marker-driven enabled/disabled split. The `Proactive saves` block is renamed `Proactive fact saves` to make the scope explicit.
- **Rules go to PREFERENCES.md, but only on explicit instruction** — new subsection clarifying that PREFERENCES.md is the curated always-on rule layer, NOT a target for proactive saves. Mirrors the policy already deployed in operator CLAUDE.md / PREFERENCES.md (PR #401, PR #402).

Removes a real ambiguity for new operators whose CLAUDE.md.example seeded their inner Claude with the older prose: fact-saves continued to route correctly, but rule-saves had no specified destination.

## Why now

Phase B was gated on Phase 3 of #396 (PR #404) merging because the new prose references the `[Memory subsystem: enabled|disabled]` marker that #404 made always-emitted. With #404 in, the marker exists in every session; this PR makes the prose that depends on it ship to operators.

## Changes

`home/.claude/CLAUDE.md.example` only. The Facts subsection bullets are preserved from the previous version, with two intentional refinements:

1. The previous "Never write to both stores in the same turn" is narrowed to "Never write to MEMORY.md and Qdrant in the same turn." This is deliberate: PREFERENCES.md writes are explicit-instruction with confirmation, so combining a proactive fact save with an operator-confirmed preference edit in one turn is permitted by policy. The blanket prohibition only needs to apply to the two fact stores.
2. `operator preferences` is removed from the proactive-saves enumeration since preferences now route to a separate destination (PREFERENCES.md) under different rules (explicit-only, no proactive saves).

The Rules subsection is new.

Doc-only; no code or test changes.

## Test plan

- [x] `tests/test_backend.py` and `tests/test_install.py` pass locally (316 tests). No test pins the Memory Write Routing prose; existing tests use synthetic example content.
- [ ] CI passes.
- [ ] After merge: existing operators keep their already-deployed CLAUDE.md (gitignored, not overwritten by reinstall). New operators get the four-way prose on first install via `make install`. Operators who want the new prose in their own CLAUDE.md merge it manually.

## Out of scope

- Phase 4 of #396 (Qdrant migration): separate sub-issue.
- Updating the `Memory System` H2 prose later in CLAUDE.md.example: covered indirectly by the new Rules subsection; no further changes needed in this PR.

### DIFF
--- a/home/.claude/CLAUDE.md.example
+++ b/home/.claude/CLAUDE.md.example
@@ -28,6 +28,10 @@ When producing content destined for a public surface (GitHub issues, pull reques
 
 ## Memory Write Routing
 
+Two distinct write categories with different policies: facts (auto-saveable) and rules (curated, explicit-only).
+
+### Facts go to MEMORY.md or Qdrant
+
 Your session context should contain a line like `[Memory subsystem: enabled]` or `[Memory subsystem: disabled]` inside the API context block.
 
 - When the line says `enabled`, persist new facts via `POST /api/memory/add` (see Memory System below).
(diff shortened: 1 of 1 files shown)


dcellison:
## Review by Kai

This is a doc-only change, and the logic is internally consistent. No bugs, security issues, or convention violations. Two minor observations:

**Suggestion - PR description is slightly inaccurate**

The description states "All bullet content for the Facts subsection is preserved verbatim from the previous version" but two lines in that subsection actually changed:

1. `Never write to both stores in the same turn.` -> `Never write to MEMORY.md and Qdrant in the same turn.` (narrowed the prohibition)
2. The proactive saves list dropped `operator preferences` from the enumeration

Both changes are correct and intentional, but the claim of verbatim preservation is inaccurate in the PR description. Not a blocker, just worth knowing if anyone audits the commit history.

**Suggestion - narrowed prohibition creates an implicit gap**

The old `Never write to both stores in the same turn` was a blanket two-store prohibition. The new wording names only `MEMORY.md and Qdrant`, which implicitly permits writing to `MEMORY.md + PREFERENCES.md` or `Qdrant + PREFERENCES.md` in the same turn. That's probably fine (proactive fact save + an explicit-instruction preference confirm in one turn is a real scenario), but if the intent was to prohibit any multi-store writes in a single turn, the original wording was stronger. Worth a conscious decision either way.

Otherwise the PR is clean. The four-way routing logic is sound, the `confirm before persisting` requirement on PREFERENCES.md writes is a good safeguard, and the removal of `preferences` from the proactive fact save list correctly reflects the new destination routing.

Issue #407, opened 2026-04-29
/memory: surface all user-visible memory sources


dcellison:
## Summary

The `/memory` Telegram UI surfaces only rows with `metadata.source == "extracted"`. Two other user-meaningful sources are currently invisible:

- **`episode`** rows — per-conversation episode summaries (#385/#387). Operators have no way to browse them, search them, or see counts in `/memory stats`.
- **`migration`** rows — operator-curated MEMORY.md content imported into Qdrant via the migration script (#406). Same gap.

After Phase 4 of #396 shipped (#408), the operator's Qdrant store contains all three source types, but the UI exposes only one. This issue expands the `/memory` UI to surface all three so the underlying distinction becomes "user-visible vs system-internal" rather than "extracted vs everything else."

## Scope

In `src/kai/memory.py`, expand the source filter from `source == "extracted"` to `source in {"extracted", "episode", "migration"}` at four sites:

1. **`get_by_tag`** (used by `/memory <tag>`) — current filter at the post-`get_all` list comprehension.
2. **`get_extracted_memory`** (used by `/memory fact <id>`) — current filter at the metadata.source check.
3. **Deletion path** (used by `/memory delete <id>`) — current refusal of non-extracted rows.
4. **Stats aggregation** (used by `/memory stats`) — currently emits only `extracted_count` plus per-tag aggregates filtered to extracted.

### D1: Stats output shape

Three options for how `/memory stats` reports per-source counts:

- (a) Keep `extracted_count`, add parallel `episode_count` + `migration_count`. Preserves source distinction for operators who want to know how many episodes the system has built up vs how many imported chunks.
- (b) Collapse into a single `user_visible_count` field. Simpler but loses source distinction.
- (c) Keep `extracted_count`, fold `episode_count` and `migration_count` into a generic `other_count` lump.

**Recommendation: (a).** The per-source counts are operationally meaningful: episode count tracks Haiku's episode-detection rate over conversation turns; migration count is a fixed one-time import number; extracted count is the long-tail accumulation. Conflating them loses signal.

### D2: Episode row rendering in `/memory`

[...]

PR #408, opened 2026-04-29
feat(memory): MEMORY.md to Qdrant migration script (#406)

### LINKED ISSUE #406: Phase 4 of #396: migrate MEMORY.md content to Qdrant
## Summary

Migrate the curated MEMORY.md content into Qdrant via the explicit `/api/memory/add` path so retrieval surfaces both extracted facts and operator-curated content under one homogeneous pool. Phase 4 of #396. Prerequisites done: Phase 3 (#403 / PR #404) shipped the switch-point gating, and the operator-side three-tier triage was deployed 2026-04-29 (Phase 2 of #396).

After this issue ships and the operator runs the migration script, the operator's MEMORY.md content lives in Qdrant under a distinct `source: "migration"` tag; toggling `MEMORY_ENABLED` between modes no longer requires data manipulation in either direction.

## Strategy

**Chunking by section header.** Each `## ` or `### ` in MEMORY.md becomes one Qdrant entry. Whole-file embedding would be too large to score sensibly. Tiny H3 bodies (under N tokens) roll up to their parent H2 to avoid surfacing one-liner fragments. The exact rollup threshold is a spec open question.

**Per-chunk similarity dedup.** For each chunk, run `memory.search()` against existing Qdrant content using the chunk text as the query. If the top match score exceeds a configured threshold (default proposed: 0.85), the chunk is treated as a likely duplicate and skipped. Otherwise it gets added. The threshold is tunable via CLI flag.

**Distinct source tag.** Migrated entries carry `source: "migration"` in metadata. New `_SOURCE_WEIGHTS["migration"]` value in `memory.py` (default proposed: 1.0; tunable). Tag serves three purposes: traceability, weight differentiation from extracted facts, and a clean rollback path via `delete_by_source(user_id, "migration")`.

**Dry-run + idempotency guard.** The script must support `--dry-run` (prints per-chunk plan with similarity scores; no writes). It must also refuse to run a real migration if migration-source entries already exist for the target user, unless `--force` is passed. This prevents accidental double-migration from re-creating duplicates.

## Scope

New file `scripts/migrate-memory-md-to-qdrant.py` (path subject to spec finalization). CLI shape:

```
python -m kai.scripts.migrate_memory_md \
    --user-id <chat_id> \
    [--memory-md <path>]      # default: DATA_DIR/memory/<chat_id>/MEMORY.md
    [--threshold 0.85]        # similarity threshold for dedup
    [--rollup-tokens 50]      # H3-to-H2 rollup threshold
    [--source-weight 1.0]     # weight assigned to migrated entries
    [--dry-run]               # print plan, no writes
    [--force]                 # bypass idempotency guard
```

`memory.py` change: add `"migration"` to `_SOURCE_WEIGHTS` and `_SOURCE_SHORT` (the latter is the `format_context` line-prefix label).

Test plan covers: chunking behavior (H2-only, H3-with-rollup, H3-without-rollup), similarity skip vs add, dry-run prints without writing, idempotency guard fires on re-run, `--force` bypasses, source tag round-trips through Qdrant.

## Acceptance

- After running the script with default flags against the deployed MEMORY.md, every non-duplicate section appears as a separate Qdrant entry with `source: "migration"`.
- `/memory` dashboard shows the migrated entries; `/memory search <query>` retrieves them alongside extracted facts ranked by source-weighted score.
- Re-running the script without `--force` exits with a clear error pointing at the existing migration entries.
- `delete_by_source(user_id, "migration")` cleanly rolls back the migration; subsequent re-run succeeds.
- `--dry-run` produces a complete plan (chunk count, per-chunk action, similarity scores) without any Qdrant writes.

## Out of scope

- Removing MEMORY.md from disk after migration. The file stays as the disabled-mode fallback per the binary mode-switch design.
- Auto-running migration during `make install`. Operator action only.
- Multi-user batch migration tooling. One operator at a time, by `--user-id`.
- Re-summarization or fact extraction from MEMORY.md content during migration. Faithful copy with mechanical chunking.
- Mode-switch verification harness beyond manual smoke testing. Separate sub-issue.

## Open questions (resolve in spec)

1. **Similarity threshold default.** 0.85 is a starting guess. Spec should propose a dry-run on the operator's real data to look at score distributions of "true matches" vs "false matches" before locking in the default.
2. **Chunk granularity.** H2-only gives big chunks (the Reference section is ~50 lines as one entry); H2+H3 splits finer but some H3s are tiny. Spec should pin the H3 rollup threshold.
3. **Migration source weight.** 1.0 (neutral, slightly below extracted's 1.2 since not conversational) vs 1.2 (operator-curated, equal weight). Defensible either way; spec should pick.
4. **Idempotency guard semantics.** "Refuse if any migration entries exist" is the safest. Alternative is "refuse if a migration entry was added in the last N hours." Spec should pick the simpler form unless there's a real use case for the second.
5. **Script location and invocation.** Standalone script under `scripts/` (per existing convention with `scripts/measure-extraction-timing.py`) versus a `python -m kai.scripts.X` module. Spec should match existing project convention.

## Severity-gradient note for spec review

This is a one-shot migration script with rollback support. The acceptance criteria are testable; the design decisions are bounded. Expect convergence in two or three review rounds at most.

dcellison:
## Summary

Phase 4 of #396. One-shot migration script that imports the operator's curated MEMORY.md content into Qdrant under `source='migration'`, with chunking by section header, per-chunk similarity dedup, idempotency guard, dry-run with score distribution, and rollback support.

After this ships and the operator runs the script, MEMORY.md content lives in Qdrant alongside Track 2 extracted facts under one homogeneous **retrieval** pool. Migration rows surface through `format_context` (the inner Claude reads them in injected context). The `/memory` Telegram UI is intentionally NOT expanded in this scope; that work is filed as #407.

Closes #406.

### `src/kai/memory.py`

Three additions, all small and surgical:

1. `_SOURCE_WEIGHTS["migration"] = 1.0` — sits between `extracted` (1.2) and the legacy `""` bucket (0.6). Operator-curated but not freshly conversational.
2. `_SOURCE_SHORT["migration"] = "fact"` — renders identically to extracted rows in `format_context` output. The source tag is for dedup and rollback, not prompt-side labeling.
3. New sync `count_by_source(user_id, source)` helper. Read-side companion to `delete_by_source` with the same page-walk pattern and the same tail-miss caveat (documented in the docstring; do not "fix" the early termination without re-reading `delete_by_source`'s rationale).

### `scripts/migrate-memory-md-to-qdrant.py` (new)

Standalone script following the existing `scripts/` convention. Lazy imports so `--help` and argparse errors do not pay the torch/Mem0 import cost.

CLI surface:

```
--user-id INT          Telegram chat_id (required)
--memory-md PATH       MEMORY.md path; default DATA_DIR/memory/<chat_id>/MEMORY.md
--threshold FLOAT      Similarity threshold for skip; default 0.85
--rollup-tokens INT    H3-to-H2 rollup body-token threshold; default 50
--dry-run              Print plan and summary, no writes
--no-summary           Disable score distribution summary
--force                Bypass idempotency guard
--rollback             Delete all migration-source entries; mutually exclusive with --force
--yes                  Skip rollback confirmation prompt
-v, --verbose          Per-chunk score logging
```

Exit codes: 0 success, 1 error, 2 idempotency guard fired, 3 memory disabled, 4 MEMORY.md missing.

Chunking: H2 + H3 only; H4+ flattens into parent H3 body text (the spec did not enumerate H4, but the deployed MEMORY.md has three H4s under `### Kai`; flattening matches the spec's "H2 or H3" granularity intent). H3s with body under `--rollup-tokens` fold into parent H2 under the `\n\n### <title>\n<body>` separator.

Slug algorithm pinned: `re.sub(r'[^a-z0-9]+', '-', title.lower()).strip('-')`. Stable across parens, version-string dots, leading slashes, colons, backticks, and conjunction patterns. Five worked examples in the spec are pinned by parametrized tests.

Rollback wraps `delete_by_source` in `asyncio.run(...)` (the function is async per the existing memory.py contract); the forward migration loop is sync only because `add_structured` and `search` are sync.

## Test coverage (16 new tests)

`tests/test_migrate_memory_md.py` (new file, loaded via `importlib.util` because the script's hyphenated filename is not a legal Python module name):

**Chunking unit tests:**
- `test_h2_only_file_produces_one_chunk_per_section`
- `test_h3_below_threshold_rolls_up_to_h2`
- `test_h3_above_threshold_stands_alone`
- `test_mixed_h3s_within_one_h2`
- `test_h1_preamble_becomes_root_chunk`
- `test_empty_section_does_not_emit_chunk`
- `test_h3_slug_generated_correctly` (parametrized over the five spec D6 worked examples)

**Integration tests via `main()` with stubbed Qdrant:**
- `test_dry_run_makes_no_writes`
- `test_below_threshold_chunk_writes`
- `test_above_threshold_chunk_skips`
- `test_idempotency_guard_blocks_second_run`
- `test_force_bypasses_idempotency_guard`
- `test_rollback_calls_delete_by_source`
- `test_rollback_dry_run_calls_count_not_delete`
- `test_memory_disabled_exits_cleanly`
- `test_memory_md_missing_exits_cleanly`

`tests/test_memory.py` extension:
- `TestCountBySource`: disabled/populated/empty cases.
- `TestMigrationSourceMetadata`: weight pin, short-prefix matches extracted.

`tests/test_eval_retrieval.py` updated: the `_reset_memory_module` helper's `_SOURCE_WEIGHTS` snapshot now includes `migration: 1.0` so it does not pollute later tests with a stale dict (caught by full-suite test run; isolated runs would have shipped with the leak).

All 2585 tests pass; `make check` clean.

## Manual rollout plan (post-merge)

1. Run `python scripts/migrate-memory-md-to-qdrant.py --user-id <id> --dry-run -v` against the operator's deployed MEMORY.md. Inspect per-chunk scores and the distribution summary. Verify `0.85` is the right cutoff: skipped chunks should be obvious duplicates of existing extracted facts; the "above p90 but below threshold" cluster (if any) is worth a manual look.
2. If retune needed: re-run `--dry-run` with adjusted `--threshold`. Iterate until skip/keep distribution matches the operator's read of the data.
3. Real run: `python scripts/migrate-memory-md-to-qdrant.py --user-id <id>`. Inspect summary.
4. Verify retrieval surfaces migrated content via a Telegram message that should hit a migrated chunk semantically; confirm the inner Claude's response references that content (or `memory.recall` log shows the migrated row in hits).
5. Smoke-test mode-switch by toggling `MEMORY_ENABLED=false` then `MEMORY_ENABLED=true` (with service restart between). Verify retrieval surfaces migrated content correctly in enabled mode.

## Out of scope

- `/memory` Telegram UI expansion to surface migration rows (filed as #407).
- Removing MEMORY.md from disk after migration (stays as the disabled-mode fallback).
- Auto-running migration during `make install`.
- Multi-user batch migration tooling.
- Re-summarization or fact extraction during migration (faithful copy, mechanical chunking).

## Notes for review

- The chunking algorithm's H4-flatten behavior is a spec-implicit choice (the spec said "H2 + H3"; H4 wasn't enumerated). Documented in code comments and PR description.
- The `_estimate_tokens` heuristic (`len(text) // 4`) is reused from `memory.py` so threshold semantics stay consistent with retrieval-side costing.
- The `tests/test_eval_retrieval.py:_reset_memory_module` change is a harness fix for the new source key; without it, tests after that fixture would see a dict missing the `migration` entry.

### DIFF
--- a/scripts/migrate-memory-md-to-qdrant.py
+++ b/scripts/migrate-memory-md-to-qdrant.py
@@ -1,0 +1,535 @@
+#!/usr/bin/env python3
+"""
+Migrate operator-curated MEMORY.md content into Qdrant.
+
+Phase 4 of #396 (issue #406). One-shot script that reads the operator's
+deployed MEMORY.md, chunks it by section header with H3-to-H2 rollup
+for short bodies, runs a per-chunk similarity check against existing
+Qdrant content to skip likely duplicates, and writes survivors via
+`memory.add_structured` with `source="migration"` metadata.
+
+Headers above H3 (i.e. H4+) are flattened into their parent H3's body
+text. The script's chunking granularity is H2 + H3 only, matching the
+spec; deeper sub-headers contribute body text but do not produce
+separate chunks.
+
+Includes dry-run with score distribution summary, tunable similarity
+threshold, idempotency guard via `memory.count_by_source`, and a
+rollback path via `memory.delete_by_source` (wrapped in `asyncio.run`).
+
+Usage:
+    # Dry run (recommended first step; inspect the plan and score
+    # distribution before committing to writes):
+    python scripts/migrate-memory-md-to-qdrant.py --user-id 123 --dry-run
+
+    # Real migration:
+    python scripts/migrate-memory-md-to-qdrant.py --user-id 123
+
+    # Rollback (deletes all migration-source rows for the user):
+    python scripts/migrate-memory-md-to-qdrant.py --user-id 123 --rollback
+
+See home/specs/406-memory-md-qdrant-migration-v3.md for the full
+design rationale and acceptance criteria.
+"""
+
+from __future__ import annotations
+
+import argparse
+import asyncio
+import re
+import sys
+from dataclasses import dataclass, field
+from pathlib import Path
+
+# Exit codes (documented in the spec):
+#   0 - success
+#   1 - unexpected error
+#   2 - idempotency guard fired (existing migration rows present)
+#   3 - memory subsystem disabled (config.memory_enabled = False)
+#   4 - MEMORY.md path does not exist
+EXIT_OK = 0
+EXIT_ERROR = 1
+EXIT_GUARD = 2
+EXIT_MEMORY_DISABLED = 3
+EXIT_FILE_MISSING = 4
+
+
+# ── Chunking ───────────────────────────────────────────────────────
+
+
+@dataclass
+class Chunk:
+    """One Qdrant-bound chunk extracted from MEMORY.md.
+
+    Attributes:
+        title: The header text (without the `#` markers). Empty for
+            untitled H1 preamble.
+        level: Heading level (1 = H1 preamble, 2 = H2 standalone or H2
+            with rolled-up H3s, 3 = H3 standalone).
+        body: The chunk's full body text WITHOUT the leading header
+            line. For level=2 chunks with rolled-up H3 children, body
+            includes the rolled-up content under the spec-mandated
+            `\\n\\n### <title>\\n<body>` separator.
+        parent_h2: For level=3 chunks, the H2 title above this chunk.
+            None for level=1 and level=2.
+        tag_slug: Slug of the title for level=3 chunks. Empty for
+            level=1 and level=2 (which have no per-chunk topical tag
+            beyond the universal `migration` tag).
+    """
+
+    title: str
+    level: int
+    body: str
+    parent_h2: str | None
+    tag_slug: str
+
+    @property
+    def text(self) -> str:
+        """The full chunk text written to Qdrant (header + body).
+
+        Format depends on level:
+          - level=1: just the body (no `# Memory` line, since H1 is
+            file-level metadata, not topical).
+          - level=2: `## <title>\\n\\n<body>` (body may include rolled-
+            up H3 sections).
+          - level=3: `### <title>\\n\\n<body>` (body may include
+            flattened H4+ sub-headers as part of the prose).
+        """
+        if self.level == 1:
+            return self.body.strip()
+        prefix = "#" * self.level
+        return f"{prefix} {self.title}\n\n{self.body}".strip()
+
+
+# Slug algorithm pinned by the spec (D6 worked-examples table). Lower-
+# case, then collapse every run of non-alphanumeric characters into a
+# single `-`, then strip leading/trailing `-`. Stable across all H3
+# titles in the deployed MEMORY.md (parens, slashes, colons, backticks,
+# version-string dots all handled uniformly).
+_SLUG_RE = re.compile(r"[^a-z0-9]+")
+
+
+def slugify(title: str) -> str:
+    """Generate the metadata.tags H3-title slug (spec §D6)."""
+    return _SLUG_RE.sub("-", title.lower()).strip("-")
+
+
+# Header detection: lines starting with one or more `#` followed by a
+# space and at least one non-space character. The count of leading `#`
+# is the heading level. Captures only the title text after the `# `.
+_HEADING_RE = re.compile(r"^(#+)\s+(.+?)\s*$")
+
+
+def _chunk_memory_md(text: str, rollup_tokens: int) -> list[Chunk]:
+    """Parse MEMORY.md text into Qdrant-bound chunks (spec §D1).
+
+    Walks lines top-to-bottom maintaining (current_h2, current_h3)
+    context. On each new heading, the pending chunk is emitted (if it
+    has body content) and a fresh accumulator starts. Rollup happens
+    in a second pass: H3 chunks below the body-token threshold get
+    folded into their parent H2's chunk text under the
+    `\\n\\n### <title>\\n<body>` separator from spec D1.
+
+    Args:
+        text: Full MEMORY.md file contents.
+        rollup_tokens: H3 chunks with `_estimate_tokens(body)` strictly
+            less than this value get rolled up into their parent H2.
+            At-or-above stand alone. The default is 50 tokens, which
+            corresponds to ~200 characters under the
+            `len(text) // 4` heuristic.
+
+    Returns:
+        List of `Chunk` objects in document order. Headers with empty
+        body and no children are dropped (they would produce one-line
+        embeddings with no useful signal).
+    """
+    # Lazy import: keeps the module load light when only parsing is
+    # exercised in tests (memory.py imports torch transitively).
+    from kai.memory import _estimate_tokens
+
+    # First pass: walk lines, accumulate raw segments.
+    @dataclass
+    class _Segment:
+        title: str  # Empty string for the H1 preamble segment.
+        level: int  # 1, 2, 3, or 4+ (4+ stays in body, no chunk emitted).
+        body_lines: list[str] = field(default_factory=list)
+
+    segments: list[_Segment] = []
+    current = _Segment(title="", level=1)
+    segments.append(current)
+
+    for raw_line in text.splitlines():
+        m = _HEADING_RE.match(raw_line)
+        if m is None:
+            current.body_lines.append(raw_line)
+            continue
+        level = len(m.group(1))
+        title = m.group(2)
+        if level >= 4:
+            # H4+ stays inside the parent's body. Preserves the sub-
+            # header line as text so retrieval can match on it. The
+            # current segment continues; we do NOT start a new one.
+            current.body_lines.append(raw_line)
+            continue
+        # H1, H2, or H3: start a new segment. Drop the H1 preamble
+        # segment if its body is empty AND it has no title (no real
+        # H1 line was seen, which is the case when the file starts
+        # with a heading other than `# `).
+        current = _Segment(title=title, level=level)
+        segments.append(current)
+
+    # Trim each segment's body. A trailing blank line before the next
+    # heading would otherwise leak a `\n` into the chunk text.
+    for seg in segments:
+        # Drop leading and trailing blank lines but preserve internal
+        # spacing (paragraph breaks).
+        while seg.body_lines and not seg.body_lines[0].strip():
+            seg.body_lines.pop(0)
+        while seg.body_lines and not seg.body_lines[-1].strip():
+            seg.body_lines.pop()
+
+    # Build initial chunks from non-empty segments.
+    chunks: list[Chunk] = []
+    current_h2: str | None = None
+    for seg in segments:
+        body_text = "\n".join(seg.body_lines).strip()
+        if seg.level == 1:
+            if not body_text:
+                continue  # No preamble; skip the placeholder H1 segment.
+            chunks.append(
+                Chunk(
+                    title=seg.title or "Memory",
+                    level=1,
+                    body=body_text,
+                    parent_h2=None,
+                    tag_slug="",
+                )
+            )
+        elif seg.level == 2:
+            current_h2 = seg.title
+            # Always emit H2 chunks even when their immediate body is
+            # empty: rollup may fold short H3 children into them in
+            # the second pass, at which point the chunk has content
+            # to write. The empty-body case is filtered below if no
+            # H3 rolled up.
+            chunks.append(
+                Chunk(
+                    title=seg.title,
+                    level=2,
+                    body=body_text,
+                    parent_h2=None,
+                    tag_slug="",
+                )
+            )
+        elif seg.level == 3:
+            if not body_text:
+                continue  # Empty H3 with no children: drop.
+            chunks.append(
+                Chunk(
+                    title=seg.title,
+                    level=3,
+                    body=body_text,
+                    parent_h2=current_h2,
+                    tag_slug=slugify(seg.title),
+                )
+            )
+
+    # Second pass: roll up small H3 chunks into their parent H2.
+    # Walk in document order; when we encounter an H3 below threshold
+    # whose parent_h2 matches the most recent H2 chunk, append its
+    # `\n\n### <title>\n<body>` to that H2's body and drop the H3.
+    rolled_up: list[Chunk] = []
+    last_h2_idx: int | None = None
+    for chunk in chunks:
+        if chunk.level == 2:
+            rolled_up.append(chunk)
+            last_h2_idx = len(rolled_up) - 1
+            continue
+        if (
+            chunk.level == 3
+            and last_h2_idx is not None
+            and chunk.parent_h2 == rolled_up[last_h2_idx].title
+            and _estimate_tokens(chunk.body) < rollup_tokens
+        ):
+            # Roll up: append separator + title + body to parent H2.
+            parent = rolled_up[last_h2_idx]
+            separator = "\n\n" if parent.body else ""
+            parent.body = f"{parent.body}{separator}### {chunk.title}\n{chunk.body}"
+            continue
+        rolled_up.append(chunk)
+
+    # Final filter: drop H2 chunks that ended up with empty body (no
+    # H2-level prose AND no H3 rolled up). Such a chunk would produce
+    # a one-line `## Title` embedding with no retrieval value.
+    return [c for c in rolled_up if c.body.strip()]
+
+
+# ── Score distribution summary ─────────────────────────────────────
+
+
+def _score_summary(scores: list[float], threshold: float) -> str:
+    """Format the dry-run end-of-run score distribution (spec §D8)."""
+    if not scores:
+        return "Similarity score distribution: no chunks emitted (file empty after parsing).\n"
+
+    sorted_scores = sorted(scores)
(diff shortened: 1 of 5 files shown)


dcellison:
## Review by Kai

## PR Review: feat(memory): MEMORY.md to Qdrant migration script (#406)

Overall the PR is well-structured and thoroughly commented. Two real bugs, one test hygiene issue, two suggestions.

---

### Warning 1 - Infinite loop in `count_by_source` (memory.py, new function ~line 778)

`count_by_source` copies the pagination shape of `delete_by_source`, but the shape only terminates safely because `delete_by_source` shrinks the store during iteration. `count_by_source` makes no side effects, so when total user rows >= `_DELETE_PAGE_SIZE` AND matching rows appear in the top-k page, both termination conditions fail:

```python
if len(rows) < _DELETE_PAGE_SIZE:  # never true: store doesn't change
    break
if matches == 0:                    # never true: matching rows exist
    break
```

The loop calls `get_all` with the same parameters forever, double-counting `matches` on every pass. This is the exact scenario that occurs after a successful migration: the operator tries a second run to verify the idempotency guard, their total row count (extracted + episode + migration) exceeds `_DELETE_PAGE_SIZE`, the guard calls `count_by_source`, and the script hangs indefinitely.

The fix for a read-only count is simpler than for delete - just break after the first page if you only need "exists or not" semantics, or accumulate without looping (fetch all once). The `delete_by_source` loop works because deletions are its side effect; the same loop body is not safe without side effects.

---

### Warning 2 - `add_structured` returning `None` is silently swallowed (migrate-memory-md-to-qdrant.py, ~line 330)

```python
mid = add_structured(...)
if mid is None:
    errors += 1          # increments, but no message printed
else:
    added += 1
```

The `except` branch two lines below does `print(..., file=sys.stderr)`. A silent `None` return produces an error count with no indication of which chunk failed or why. Should print to stderr here the same as the `except` branch.

---

### Warning 3 - Test state leak on assertion failure (tests/test_memory.py, ~lines 1206-1248)

`test_count_by_source_returns_count` and `test_count_by_source_handles_empty_user` set `mem_mod._memory = mock_mem` then reset inline:

```python
assert count_by_source("user-a", "migration") == 3
mem_mod._memory = None  # never reached if assert fires
```

If the assertion raises, `_memory` stays as the mock for subsequent tests. The existing `TestDeleteBySource` tests use `monkeypatch` or `MagicMock` scoped to fixtures for exactly this reason. Should use `monkeypatch` or `try/finally`.

---

### Suggestion 1 - Implicit encoding in `read_text` (migrate-memory-md-to-qdrant.py, ~line 480)

```python
chunks = _chunk_memory_md(memory_md.read_text(), rollup_tokens=args.rollup_tokens)
```

Should be `memory_md.read_text(encoding="utf-8")`. Non-UTF-8 locales (or MEMORY.md files with smart quotes from editor autocorrect) will get a UnicodeDecodeError or wrong characters with the system default.

---

### Suggestion 2 - `--yes` bypasses the zero-row short-circuit in rollback (migrate-memory-md-to-qdrant.py, ~lines 399-417)

Without `--yes`, a zero-row result prints "No migration entries to delete" and exits before calling `delete_by_source`. With `--yes`, `delete_by_source` is called unconditionally regardless of row count. Not a bug, but the output "Deleted 0 migration entries" is slightly surprising when the operator intended a no-op verification. A pre-check (without the interactive prompt) in the `--yes` path would mirror the non-`--yes` short-circuit.

---

### Clean areas

- Chunking algorithm (`_chunk_memory_md`) and rollup logic are correct. Edge cases (H3 before any H2, empty preamble segment, H4+ flattening, multiple H2s with same title) are all handled properly.
- `slugify` matches the spec §D6 worked examples; parametrized tests cover all five cases.
- `_SOURCE_WEIGHTS["migration"] = 1.0` and `_SOURCE_SHORT["migration"] = "fact"` are correct insertions. The `_reset_memory_module` fix in `test_eval_retrieval.py` is necessary and correct.
- The `importlib.util` loading pattern with pre-registration in `sys.modules` before `exec_module` is the correct fix for the `@dataclass` `__module__` lookup issue.
- Exit codes, idempotency guard, dry-run/rollback mutual exclusion, and the `asyncio.run` wrap for the async `delete_by_source` are all correct.

dcellison:
## Review by Kai

Looking at the diff against the prior review thread, all five prior findings were addressed: the `count_by_source` loop bug is fixed with a single-fetch design, the silent `None` return is now printed to stderr, tests use `monkeypatch`, `read_text` has explicit encoding, and the `--yes` zero-row short-circuit is plugged with a new pre-check and test. The base code is solid. Three remaining findings below.

---

### Warning - `count_by_source` warning fires on page fullness, not match count (`src/kai/memory.py`, ~line 781)

```python
if len(rows) >= _DELETE_PAGE_SIZE:
    log.warning(
        "count_by_source: get_all returned %d rows (full page); "
        "actual count for user_id=%s source=%r may be higher. ...",
        _DELETE_PAGE_SIZE,
        user_id,
        source,
    )
```

The condition triggers on total fetched rows regardless of how many matched `source`. If the user has thousands of `extracted`/`episode` rows and zero `migration` rows, `get_all` returns a full page, the match count is 0, and the log says `"migration count may be higher"`. That message is technically defensible (Mem0's result ordering isn't guaranteed, so unseen pages could have migration rows), but in practice it will alarm operators post-migration who run a sanity `count_by_source` call and see a warning implying migration rows may be lurking. The fix is a message change only - add the total-row context so the operator understands what fired the warning:

```python
log.warning(
    "count_by_source: get_all returned %d total rows (page cap); "
    "matched %d rows for source=%r. If the user has >%d total rows, "
    "the returned count is a lower bound. ...",
    _DELETE_PAGE_SIZE, count, source, _DELETE_PAGE_SIZE,
    ...
)
```

---

### Suggestion - Dead `plan_lines` list in `_do_forward_migration` (`scripts/migrate-memory-md-to-qdrant.py`, ~line 280)

```python
plan_lines: list[str] = []
...
plan_lines.append(f"CHUNK {idx:2d} {action} score={top_score:.3f} title={title_label!r}")
if args.verbose:
    print(plan_lines[-1])
```

`plan_lines` is built but never read as a list - only `plan_lines[-1]` (the just-appended item) is used inside the loop. The list has no consumer after the loop exits. Replacing this with a local `line = f"..."` and `if args.verbose: print(line)` removes the dead accumulator. Minor, but the current code implies the list is needed for something (perhaps a planned all-at-once dump) that never materializes.

---

### Suggestion - `fake_memory_md` fixture in rollback test is unused (`tests/test_migrate_memory_md.py`, line ~296)

```python
def test_rollback_calls_delete_by_source(fake_memory_md, stub_memory_module, enabled_config):
```

The `--rollback` path in `main()` returns before reading MEMORY.md (the `if args.rollback: return _do_rollback(args)` branch runs before `memory_md.exists()` is reached), so `fake_memory_md` is created and never opened. Removing it from the signature makes the test's intent clearer and prevents a future reader from assuming the rollback path exercises file I/O. Same applies to `test_rollback_dry_run_calls_count_not_delete` at line ~316.

---

Everything else is clean: chunking logic handles H4-flatten and rollup correctly, the `asyncio.run` wrap for `delete_by_source` is correct, exit codes and mutual-exclusion guards are correct, the `_reset_memory_module` fix in `test_eval_retrieval.py` is necessary and correct, and the `importlib.util` + pre-registration pattern is the right fix for the `@dataclass` module-lookup issue.

dcellison:
## Review by Kai

## PR Review: feat(memory): MEMORY.md to Qdrant migration script (#406)

Prior review findings are all addressed. Two new issues, both in the unrelated file bundled into this PR.

---

### Warning 1 - `scripts/measure-extraction-timing.py` is out of scope and contains unverified CLI flags

This file does not appear in the PR description, is unrelated to MEMORY.md-to-Qdrant migration, and violates the project convention: "Don't add features or refactor beyond what was asked."

Beyond the scope problem, the subprocess command at lines 93-116 uses several flags that need verification per the project convention ("Verify CLI flags exist (via --help or docs) before adding them to subprocess commands. Never invent flags based on config key names"):

- `--json-schema` - not a documented `claude --print` flag
- `--max-budget-usd` - not a documented `claude --print` flag
- `--system-prompt` - needs verification
- `--permission-mode bypassPermissions` - `permissionMode` is a `settings.json` key; the convention says "use `--settings` for settings.json options", not `--permission-mode` as a CLI flag
- `--no-session-persistence` - needs verification
- `"--tools", ""` - passing an empty string to `--tools` is almost certainly wrong; the flag likely expects a list of tool names or should be omitted entirely

If `claude --print` silently ignores unrecognized flags, these runs will use default behavior rather than the intended configuration, making the timing results misleading without indicating why.

This file should be moved to a separate PR after the flags are verified.

---

### Suggestion - `near_misses` is vacuously zero when `threshold < p90` (`scripts/migrate-memory-md-to-qdrant.py`, `_score_summary`)

```python
near_misses = sum(1 for s in scores if p90 <= s < threshold)
```

If an operator passes `--threshold 0.60` and p90 is 0.80, this range is empty (`0.80 <= s < 0.60` is impossible), so `near_misses = 0` and the label "Above p90 but below threshold: 0 chunks" is printed without any indication that threshold is set below p90. Operators using low thresholds (intentional aggressive dedup) get a misleading "all clear" on that line.

A guard like `if threshold > p90:` before printing the near-misses line, or a note when threshold <= p90, would make the summary honest in that configuration.

---

### Clean areas

The core PR content is solid after the prior rounds. `count_by_source` single-fetch design is correct; the rollup/chunking algorithm handles H3 rollup, H4-flatten, and document-order correctly; the `--yes` zero-row short-circuit and the new `test_rollback_short_circuits_when_no_migration_rows` test are both correct; `monkeypatch` usage in `TestCountBySource` is correct; the `_reset_memory_module` fix in `test_eval_retrieval.py` is necessary and correct.

dcellison:
## Review by Kai

All three warning-level findings from the prior review round have been addressed:
- The `count_by_source` warning message now spells out total vs matched rows, so a zero-match result on a full page does not alarm the operator.
- The dead `plan_lines` accumulator is gone; the loop now uses a local `print(...)` directly.
- Both rollback tests drop the unused `fake_memory_md` fixture parameter.

I reviewed the full diff for new issues and found none at warning level or above.

**Suggestion - `test_rollback_calls_delete_by_source` assertion text is slightly misleading (`tests/test_migrate_memory_md.py`, ~line 297)**

The docstring says "Pre-check via `count_by_source` short-circuits when `count==0`..." but the pre-check behavior is tested separately in `test_rollback_short_circuits_when_no_migration_rows`. Within `test_rollback_calls_delete_by_source`, the fixture sets `count_by_source.return_value = 7`, which is correct - no issue with the code, just the comment repeating logic that lives in the other test. Minor clarity nit only.

**Everything else is clean.** The chunking algorithm (H4-flatten, rollup, orphaned-H3 pass-through), `count_by_source` single-fetch design and cap warning, `asyncio.run` wrap for `delete_by_source`, `--yes` zero-row short-circuit, `importlib.util` pre-registration for `@dataclass` module lookup, and the `_reset_memory_module` snapshot update are all correct. Test coverage is solid with the full-page loop regression test pinning the single-fetch contract explicitly.

dcellison:
## Review by Kai

This PR has gone through four thorough prior review rounds. After examining the current diff against those threads:

**All prior warnings and suggestions have been addressed:**
- `count_by_source` single-fetch design is correct; the full-page cap warning now spells out total vs matched rows
- Silent `None` return from `add_structured` now prints to stderr
- `plan_lines` dead accumulator is gone; replaced with inline `print`
- Both rollback tests drop the unused `fake_memory_md` fixture parameter
- `read_text(encoding="utf-8")` explicit
- `--yes` zero-row short-circuit is present before calling `delete_by_source`
- `test_rollback_calls_delete_by_source` docstring now explains the `return_value=7` setup
- `scripts/measure-extraction-timing.py` is not in this diff

**No new issues found at warning level or above.**

One observation worth noting, but not a new finding: `_score_summary` at line ~250 still prints "Above p90 but below threshold: 0 chunks (manual review recommended)" when `threshold <= p90`, making the line vacuously misleading for aggressive dedup configs. This was raised as a suggestion in round 3 and the author has seen it.

**Clean areas confirmed:**
- Chunking algorithm (`_chunk_memory_md`): H4-flatten, rollup second-pass with interleaved standalone H3s, empty-body filter, orphaned H3 pass-through all correct
- `slugify` matches all five spec D6 examples
- `asyncio.run(delete_by_source(...))` wrap is correct for a sync script calling an async function
- `monkeypatch.setattr("kai.memory.X", ...)` correctly intercepts the lazy `from kai.memory import X` calls inside `main()`/`_do_rollback()`/`_do_forward_migration()`
- `_reset_memory_module` snapshot in `test_eval_retrieval.py` updated correctly
- Exit codes, mutual exclusion guards, and idempotency guard logic all correct

The PR is ready to merge.

PR #409, opened 2026-04-29
feat(memory): surface episode and migration rows in /memory UI (#407)

### LINKED ISSUE #407: /memory: surface all user-visible memory sources

dcellison:
## Summary

- Expands `/memory` from extracted-only to the three user-visible sources (extracted, episode, migration). New `_USER_VISIBLE_SOURCES` constant in `memory.py` gates `get_by_id` and `get_by_tag`; `delete_by_id` inherits via delegation.
- Adds `episode_count` and `migration_count` to `MemoryStats` (parallel per-source counts; confidence aggregates and `by_tag` stay extracted-only).
- Dashboard and stats screens combine empty-state guards into one all-zero check, surface per-source headlines, and (for the dashboard) gain a three-branch footer that points at Search/Stats when no tag buttons render.
- Fact view branches on `metadata.source`: extracted keeps the original six-row block; episode shows `Episode (<outcome_quality>)` + Tags + Date; migration shows `Imported` + Tags + Date.
- Forget-confirm injects per-source noun (`fact` / `episode` / `imported memory`); the verb (`Forget`), warning (`This cannot be undone.`), and inline-button flow are unchanged.

Fixes #407.

## Test plan

- [x] `make test` (all 2614 tests pass; +27 new across `tests/test_memory.py` and `tests/test_memory_command.py`).
  - `/memory` headline shows per-source counts when episodes / migration rows exist.
  - `/memory stats` shows `Total facts:` / `Total episodes:` / `Total imported:` lines as appropriate; an episode-only operator sees no by-tag table.
  - Tap a tag button (e.g. `preference`) when a migration or episode row carries it; verify the row appears.
  - Open a migration row's fact view: header reads `Imported`; body has Tags + Date only.
  - Open an episode row's fact view: header reads `Episode (<outcome_quality>)` (or `Episode`); body has Tags + Date only.
  - Tap `forget` on a migration row: confirmation prompt reads `Forget this imported memory? "<text>" This cannot be undone.`; tap `cancel`.

### DIFF
--- a/src/kai/memory.py
+++ b/src/kai/memory.py
@@ -1297,17 +1297,23 @@ def get_stats(*, user_id: str) -> MemoryStats:
     for m in memories:
         by_type[m.memory_type] = by_type.get(m.memory_type, 0) + 1
 
-    # Extracted-only aggregates. Build the filtered list once; every
-    # downstream metric reads from it.
-    extracted = [m for m in memories if m.metadata.get("source") == "extracted"]
-
-    # Per-source counts (issue #407). Computed in one walk alongside
-    # the extracted list so the aggregation stays O(n) rather than
-    # O(n) per source. Confidence aggregates and `by_tag` stay scoped
-    # to extracted (see `MemoryStats` docstring); these counts are the
-    # only new aggregation work needed here.
-    episode_count = sum(1 for m in memories if m.metadata.get("source") == "episode")
-    migration_count = sum(1 for m in memories if m.metadata.get("source") == "migration")
+    # Per-source partition (issue #407). Single walk over `memories`;
+    # the extracted list feeds the confidence / by_tag / prompt_version
+    # aggregation below, while episode and migration counts are surfaced
+    # directly in MemoryStats. Confidence aggregates and `by_tag` stay
+    # scoped to extracted (see `MemoryStats` docstring); these counts
+    # are the only new per-source aggregation work needed here.
+    extracted: list[MemoryResult] = []
+    episode_count = 0
+    migration_count = 0
+    for m in memories:
+        src = m.metadata.get("source")
+        if src == "extracted":
+            extracted.append(m)
+        elif src == "episode":
+            episode_count += 1
+        elif src == "migration":
+            migration_count += 1
 
     by_tag: dict[str, int] = {}
     by_prompt_version: dict[str, int] = {}
--- a/src/kai/memory.py
+++ b/src/kai/memory.py
@@ -148,6 +148,25 @@ _SOURCE_SHORT: dict[str, str] = {
 _DELETE_PAGE_SIZE = 10_000
 
 
+# Sources that the /memory Telegram UI surfaces as user-visible rows.
+# Used by `get_by_id` and `get_by_tag` (and via delegation, `delete_by_id`)
+# to gate addressability from the operator-facing surface. Three reasons
+# the set is intentional rather than "everything except legacy":
+#   - extracted: Track 2 Haiku-derived facts; the original target of the
+#     /memory dashboard.
+#   - episode:   per-conversation episode summaries (issue #385/#387).
+#     Tagged distinctly so the fact-view can render Sophia-style fields
+#     instead of empty extractor placeholders.
+#   - migration: operator-curated MEMORY.md content imported via #408.
+#     Fact-view header reads "Imported" so operators can distinguish at
+#     a glance from extracted facts whose freshness implications differ.
+# Legacy ""-source rows (from pre-spec ingestion paths or any future-
+# additional source not in this set) stay hidden in the UI; they are
+# managed via memory_admin.py / `delete_by_source`. A frozenset is used
+# so the constant cannot be mutated by importing callers.
+_USER_VISIBLE_SOURCES: frozenset[str] = frozenset({"extracted", "episode", "migration"})
+
+
 @dataclass(frozen=True)
 class MemoryResult:
     """A single memory from search or retrieval."""
@@ -186,6 +205,16 @@ class MemoryStats:
     sentinel (None for min/median/max, 0 for counts, empty dict for the
     grouped fields) so a caller that constructs `MemoryStats(total_count=0,
     by_type={})` (the legacy two-arg form) still produces a valid value.
+
+    `episode_count` and `migration_count` are parallel per-source counts,
+    added so the /memory dashboard and stats screens can surface
+    user-visible non-extracted sources alongside the extracted total.
+    Their sum with `extracted_count` may be less than `total_count`
+    because legacy ""-source rows still contribute to the total but are
+    not enumerated as a separate per-source field. Confidence aggregates
+    and `by_tag` stay scoped to extracted (see comments below) because
+    episode and migration rows do not carry confidence and use tag
+    spaces that are not enumerated by `_TAG_ENUM` in memory_command.py.
     """
 
     total_count: int
@@ -198,6 +227,12 @@ class MemoryStats:
     # because picking 0.0 would be misleading (it is a valid score
     # value, not "no data"). Counts default to 0 / empty dict.
     extracted_count: int = 0
+    # Parallel per-source counts (issue #407). Defaulted to zero so the
+    # legacy two-arg `MemoryStats(total_count=0, by_type={})`
+    # construction in tests stays source-compatible. Episode rows come
+    # from issue #385/#387; migration rows come from issue #406/#408.
+    episode_count: int = 0
+    migration_count: int = 0
     by_tag: dict[str, int] = field(default_factory=dict)
     confidence_min: float | None = None
     confidence_median: float | None = None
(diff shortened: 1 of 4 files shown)


dcellison:
## Review by Kai

This is a well-structured PR with thorough test coverage. Here are the findings:

---

## Warning

**Inaccurate comment in `memory.py` `get_stats` (new lines ~1304-1308)**

```python
# Per-source counts (issue #407). Computed in one walk alongside
# the extracted list so the aggregation stays O(n) rather than
# O(n) per source.
episode_count = sum(1 for m in memories if m.metadata.get("source") == "episode")
migration_count = sum(1 for m in memories if m.metadata.get("source") == "migration")
```

The comment claims "one walk" but the code performs **three separate O(n) passes** - one for the `extracted` list comprehension and two `sum()` generators. "O(n) rather than O(n) per source" is backwards; three O(n) passes *is* O(n) per source. Since `memories` is already in memory the performance impact is negligible, but the comment is misleading and should be corrected or dropped.

---

## Suggestions

**`_build_fact_view`: `tags_line` is dead code in the extracted branch (`memory_command.py` ~line 546)**

```python
tags_line = f"Tags:  {', '.join(tags) if tags else '(none)'}"

if source == "episode":
    lines = [..., tags_line, ...]
elif source == "migration":
    lines = [..., tags_line, ...]
else:
    # extracted - does NOT use tags_line, re-creates the tags string inline
    lines = [..., f"Tags:             {', '.join(tags) if tags else '(none)'}", ...]
```

`tags_line` is computed unconditionally but never used in the `else` branch. The extracted branch intentionally uses wider padding to align with its longer labels. No functional impact, but the variable only belongs inside the `episode`/`migration` branches.

---

**`_build_dashboard`: "0 tags" / "1 tags" grammar edge case (`memory_command.py` ~line 366)**

```python
summary = "Memories: " + ", ".join(parts) + f" across {len(nonzero)} tags."
```

Episode-only or migration-only operators (no extracted rows, so `nonzero` is always empty) will now see `"Memories: 4 episodes across 0 tags."` - the pre-#407 surface never reached this line for that case. The "1 tags" singular issue is pre-existing. Neither is a bug, but the zero-tag wording is awkward since the footer already says to use Search; the summary implying "across 0 tags" may read confusingly. A simple `f"across {len(nonzero)} {'tag' if len(nonzero) == 1 else 'tags'}"` would also fix the singular.

---

## Clean areas

- Source-admit gating via `_USER_VISIBLE_SOURCES` in `get_by_id` and `get_by_tag` is correct; `delete_by_id` inherits correctly via delegation.
- No auth bypass: user_id scoping is preserved in `get_by_id`.
- No injection risk: `source` from metadata drives display branches only, not queries or template rendering.
- The combined empty-state guard (`extracted == 0 and episode == 0 and migration == 0`) correctly replaces the two prior guards.
- `_build_forget_fact_confirm` defensive fallback to `"memory"` for unknown sources is sound.
- Test coverage is thorough: all three source shapes are pinned for `get_by_id`, `get_by_tag`, `delete_by_id`, `get_stats`, dashboard, stats view, fact view, and forget-confirm.

dcellison:
## Review by Kai

Looking at the diff against the prior review thread.

## Prior review items - both addressed

**Inaccurate "one walk" comment** (prior warning): Fixed. The three separate O(n) passes have been replaced with a single `for m in memories:` loop in `get_stats` (`memory.py` ~line 1310). Comment now correctly says "Single walk over `memories`".

**`tags_line` dead code** (prior suggestion): Fixed. `tags_line` is now computed inside the `if source == "episode" or source == "migration":` block, not unconditionally. It was never reaching the `else` branch.

**"0 tags" / "1 tags" grammar** (prior suggestion): Singular case fixed with `tag_word`. "0 tags" for episode/migration-only operators is documented as intentional.

---

## New findings

**Suggestion - `_build_stats`: empty headline block has no trailing blank line before potential future sections** (`memory_command.py`)**

Not a bug today, but worth noting: when only episode or migration counts are non-zero, `_build_stats` produces `"Memory stats\n\nTotal episodes:   5"` with no trailing blank. The `if stats.extracted_count > 0:` block adds its own leading `""` blank line, so extracted + non-extracted mixed output is fine. This is correct as-is; noting only because the layout differs from the extracted-only case which always had a blank line before "By tag:".

---

## Clean areas (new)

- `_USER_VISIBLE_SOURCES` frozenset membership check on `None` source values: `None not in frozenset({"extracted", ...})` is `False`, so rows missing the `source` field are correctly excluded from both `get_by_id` and `get_by_tag`.
- `_build_forget_fact_confirm`: `label` comes from a fixed dict lookup, not the raw metadata string - no injection vector.
- `_build_dashboard` `parts` list: always non-empty past the all-zero guard, so `", ".join(parts)` never produces `"Memories: across N tags."`.
- `MemoryStats` backwards compatibility: `episode_count=0` and `migration_count=0` defaults preserve the legacy `MemoryStats(total_count=0, by_type={})` two-arg construction path.
- Auth scoping: `user_id` ownership check in `get_by_id` is preserved and unchanged; the source-admit expansion does not widen the ownership boundary.

This PR is clean. No new bugs, security issues, or convention violations found.

PR #411, opened 2026-04-29
feat(memory): add episode browser to /memory dashboard (#410)

### LINKED ISSUE #410: Surface episodes in /memory dashboard
## Problem

Post-#407 the `/memory` UI accepts episode rows wherever `_USER_VISIBLE_SOURCES` gates a fetch (`get_by_id`, `get_by_tag`), but episodes have no first-class discovery surface. The dashboard's tag buttons key off `_TAG_ENUM` (Haiku-extractor tags only); episode rows mostly carry Sophia-style tags that aren't in the enum, so they only appear when an episode happens to overlap an enum tag like `preference`. In practice, an operator stumbles onto an episode rather than finding one.

A second, related defect: `_send_search` at `src/kai/memory_command.py:1286` post-filters search results to `metadata.source == "extracted"`, which means `/memory search <query>` will not surface episode (or migration) rows even when they match. #407 didn't touch this site.

## Proposed surface

Add an `Episodes (N)` button to the dashboard, alongside the existing tag buttons, when `episode_count > 0`. Tapping it opens a paginated list of all episode rows for the operator (same shape as the tag drill-down, with the row's `Episode (<outcome_quality>)` flavor visible per row). Tapping a numbered button opens the existing fact view, which already renders the episode shape correctly.

While we're in here, fix `_send_search` to admit all `_USER_VISIBLE_SOURCES` so search becomes a working secondary path to episodes.

## Out of scope

- Migration rows do not get a parallel browser. After import, migration rows are functionally extracted facts; the source label is provenance for QA and rollback only, not a useful ongoing browse axis.
- `_TAG_ENUM` is not extended. Episode tag-space unification is #388 territory.
- No new text subcommand (`/memory episodes`). The dashboard button is the discovery surface.

dcellison:
## Summary

- Adds an `Episodes (N)` button to the `/memory` dashboard utility row when `stats.episode_count > 0`. Tapping it opens a paginated list view (same shape as the tag drill-down) showing all episode rows for the operator, formatted as `N. [<outcome_quality>] <text>\n   <date>`. Tapping a numbered button opens the existing fact view, which already renders the `Episode (<outcome_quality>)` shape; back returns to the same page of the episode list.
- Widens `_send_search` post-filter from `source == "extracted"` to `source in _USER_VISIBLE_SOURCES`, so `/memory search` surfaces episode and migration rows alongside extracted ones. Legacy `""`-source rows still drop.
- Updates the dashboard footer prose to name the Episodes button when it is rendered without tag buttons (D6).
- Updates `_HELP_TEXT`'s first line to "Browse memories" (D7).
- Rewrites the module docstring to map the post-#407 / post-#410 source-filter sites (D5).

Fixes #410.

## Test plan

- [x] `make test` (2636 tests pass; +22 new across `tests/test_memory.py` and `tests/test_memory_command.py`; one existing test updated for the new footer wording).
  - `/memory` headline shows the Episodes button when episodes exist; verify count matches `MemoryStats.episode_count`.
  - Tap `Episodes (N)`; verify the paginated list renders with `[success]` / `[partial]` / `[failure]` brackets and dates.
  - Tap a numbered button; verify the fact view opens with the `Episode (<outcome_quality>)` header. Back returns to the same page.
  - `/memory search "evaluate spec"` (or another query likely to hit an episode); verify episode rows now surface in results.
  - Confirm migration-only operator (no extracted, no episodes) still sees the pre-#410 footer wording (`Use Search to find specific memories, or tap Stats for details.`).

### DIFF
--- a/src/kai/memory.py
+++ b/src/kai/memory.py
@@ -164,7 +164,7 @@ _DELETE_PAGE_SIZE = 10_000
 # additional source not in this set) stay hidden in the UI; they are
 # managed via memory_admin.py / `delete_by_source`. A frozenset is used
 # so the constant cannot be mutated by importing callers.
-_USER_VISIBLE_SOURCES: frozenset[str] = frozenset({"extracted", "episode", "migration"})
+USER_VISIBLE_SOURCES: frozenset[str] = frozenset({"extracted", "episode", "migration"})
 
 
 @dataclass(frozen=True)
@@ -1104,7 +1104,7 @@ def get_by_tag(*, user_id: str, tag: str) -> list[MemoryResult]:
     client-side because Mem0's `get_all` does not accept metadata
     filters. Two filter clauses, both load-bearing:
 
-      - `metadata.source in _USER_VISIBLE_SOURCES`: defends against
+      - `metadata.source in USER_VISIBLE_SOURCES`: defends against
         legacy ""-source (or any future-additional non-UI) rows
         leaking into the operator-facing surface. Issue #407 expanded
         this from `== "extracted"` to the three-source set so episode
@@ -1131,7 +1131,7 @@ def get_by_tag(*, user_id: str, tag: str) -> list[MemoryResult]:
 
     rows = get_all(user_id=user_id, limit=None)
     matches = [
-        r for r in rows if r.metadata.get("source") in _USER_VISIBLE_SOURCES and tag in (r.metadata.get("tags") or [])
+        r for r in rows if r.metadata.get("source") in USER_VISIBLE_SOURCES and tag in (r.metadata.get("tags") or [])
     ]
     # Newest-updated first; created_at is the fallback for rows whose
     # payload predates updated_at being recorded. String comparison on
@@ -1151,7 +1151,7 @@ def get_all_episodes(*, user_id: str) -> list[MemoryResult]:
     enumerates every episode in one place.
 
     The source filter here is intentionally narrower than
-    `_USER_VISIBLE_SOURCES`: this function's purpose is single-source
+    `USER_VISIBLE_SOURCES`: this function's purpose is single-source
     enumeration ("give me everything in the episode bucket"), not
     multi-source admission. The shared admit list lives in
     `get_by_id` / `get_by_tag`; this helper does not participate in
@@ -1189,7 +1189,7 @@ def get_by_id(*, user_id: str, memory_id: str) -> MemoryResult | None:
       - Legacy ""-source rows are out of scope for /memory UI
         surfaces; they belong to memory_admin.py. Hide them here
         rather than letting the dashboard expose them. The accepted
-        sources are those in `_USER_VISIBLE_SOURCES` (extracted,
+        sources are those in `USER_VISIBLE_SOURCES` (extracted,
         episode, migration). Issue #407 expanded the set from
         extracted-only so episode (#385/#387) and migration (#406/
         #408) rows are addressable from the operator-facing surface;
@@ -1228,11 +1228,11 @@ def get_by_id(*, user_id: str, memory_id: str) -> MemoryResult | None:
             user_id,
         )
         return None
-    if (row.get("metadata") or {}).get("source") not in _USER_VISIBLE_SOURCES:
+    if (row.get("metadata") or {}).get("source") not in USER_VISIBLE_SOURCES:
         # Legacy ""-source (or any future-additional non-UI source) is
         # deliberately invisible to /memory. The user-visible set
         # (extracted, episode, migration) is enumerated in
-        # `_USER_VISIBLE_SOURCES`. No log: this is a routine filter,
+        # `USER_VISIBLE_SOURCES`. No log: this is a routine filter,
         # not an anomaly.
         return None
 
--- a/src/kai/memory.py
+++ b/src/kai/memory.py
@@ -1141,6 +1141,38 @@ def get_by_tag(*, user_id: str, tag: str) -> list[MemoryResult]:
     return matches
 
 
+def get_all_episodes(*, user_id: str) -> list[MemoryResult]:
+    """Return all episode-source memories for `user_id`, newest first.
+
+    Read-side primitive for the /memory dashboard's episode-list
+    browser (issue #410). Episode rows mostly carry Sophia-style tags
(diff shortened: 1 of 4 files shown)


dcellison:
## Review by Kai

The `_ScreenCache.screen` field is `str` (not a `Literal`), so no type issue there. Now I have everything needed.

---

## Review: feat(memory): add episode browser to /memory dashboard (#410)

Overall the implementation is clean and well-tested. Three findings below.

---

### Suggestion - Private symbol cross-module access

**`memory_command.py`, `_send_search`**

```python
filtered = [r for r in results if r.score >= floor and r.metadata.get("source") in memory._USER_VISIBLE_SOURCES]
```

`_USER_VISIBLE_SOURCES` carries a leading underscore (Python's "module-private" convention) but is accessed from a sibling module. The module docstring justifies this as deliberate ("both sites read from `memory.py`"), but it is a convention violation under the project's standard. A public alias (`USER_VISIBLE_SOURCES` or `VISIBLE_SOURCES`) would be the correct fix. Low practical risk since it is intra-package, but worth surfacing.

---

### Suggestion - Sort key can produce `None` when both timestamps are absent

**`memory.py`, `get_all_episodes`** (line ~1170)

```python
matches.sort(key=lambda r: r.updated_at or r.created_at, reverse=True)
```

If a row has `updated_at=None` and `created_at=None`, the key returns `None`. Python 3 raises `TypeError` when comparing `None` against a non-`None` value in the same sort pass, so a store with even one properly-dated row alongside a dateless row would crash. This is a direct copy of `get_by_tag`'s sort, so it is a pre-existing pattern rather than a regression introduced here - but `get_all_episodes` is a new public surface. The fix is the same as `get_by_tag` would need: a sentinel fallback, e.g. `r.updated_at or r.created_at or ""`. Worth tracking even if not blocking this PR.

---

### Cosmetic note - Date-row indent does not align under episode text for longer quality strings

**`memory_command.py`, `_build_episode_list_view`** (comment block around the `lines.append(f"            {date_str}")` call)

The comment explains the 12-space indent matches the tag view's column. For `[failure]` (9 chars) or `[partial]` (7 chars), the text column starts at position 15 or 13 respectively, so the date appears left of the text rather than under it. For `[success]` (9 chars) it is also off. The comment acknowledges this as a "compromise" and Telegram's non-monospace rendering in most clients makes exact column alignment mostly moot - but the comment's math ("same column as the tag view") is only true for the tag view itself, not for the episode rows. Not a bug; leaving it as-is is fine.

---

### Everything else looks good

- `_ScreenCache.screen: str` - no Literal type to update; `"episodes"` is safe.
- `None in frozenset(...)` returns `False` correctly, so the search post-filter handles missing `source` keys.
- The `return_to = ("eps", [str(cache.page)])` / delete-routing / back-nav chain is consistent end-to-end.
- `_OUTCOME_QUALITY_ENUM` duplication is correctly motivated and documented.
- `_build_episode_list_view` empty-state returns a valid `total=1` so "page 1 of 1" is never rendered from the empty branch.
- 22 new tests cover all the branches described in the PR, including the negative-space contracts.

dcellison:
## Review by Kai

This PR is clean. The prior suggestion to make `USER_VISIBLE_SOURCES` public has been addressed, and the new code is well-structured.

One minor finding not previously raised:

**Suggestion - `_OUTCOME_QUALITY_ENUM` type inconsistency**

`memory_command.py`, near `_OUTCOME_QUALITY_ENUM`:

```python
_OUTCOME_QUALITY_ENUM: tuple[str, ...] = ("success", "partial", "failure")
```

`USER_VISIBLE_SOURCES` uses `frozenset` for the same membership-testing pattern. Using `tuple` here is inconsistent and slightly less idiomatic - `frozenset` would better signal "unordered set used for `in` checks" and would match the established pattern. Functionally negligible at 3 elements.

---

- The `return_to = ("eps", [str(cache.page)])` / delete-routing / back-nav chain is consistent end-to-end for the new screen.
- `int(args[0]) if args else 0` in the `eps` dispatch handles both empty-args and ValueError safely.
- The empty state returning `total=1` is intentional and the caller correctly ignores `_total` with the underscore convention.
- `_send_episode_list` wraps `get_all_episodes` in a try/except and degrades gracefully.
- The 22 new tests cover all branches including negative-space contracts.

The previously noted sort key `None` risk in `get_all_episodes` (pre-existing pattern from `get_by_tag`) was seen and accepted by the author; not re-raised here.

dcellison:
## Review by Kai

This is a clean, well-tested PR. The two prior findings (public `USER_VISIBLE_SOURCES` alias, `frozenset` for `_OUTCOME_QUALITY_ENUM`) are both resolved. One new finding:

---

### Suggestion - Footer omits Episodes button when tags are also present

**`memory_command.py`, `_build_dashboard`**

The footer branch structure is:

```python
if nonzero and stats.confidence_median is not None:
    lines.append("Tap a tag to browse. Confidence: ...")
elif nonzero:
    lines.append("Tap a tag to browse.")
elif stats.episode_count > 0:          # <-- only reached when nonzero is empty
    lines.append("Tap Episodes to browse, ...")
else:
    ...
```

The utility-row keyboard, however, adds the Episodes button unconditionally whenever `stats.episode_count > 0` - so an operator who has both tags and episodes gets `[Episodes (N), Search, Stats]` in the keyboard but a footer that only says "Tap a tag to browse." The Episodes button is silently present.

This is consistent with the existing convention (the tag-only footer also omits Search and Stats), so it is not a regression, but it is the one user-visible case where the D6 fix ("footer names every utility-row affordance") was not applied. If the intent is that D6 only applies to the no-tag-buttons case, the code and comment are correct as-is; if the intent is exhaustive footer coverage whenever Episodes renders, the first two branches need a suffix for `stats.episode_count > 0`. Low priority, no blocking issue.

---

- `return_to = ("eps", [str(cache.page)])` / delete routing / back-nav chain is consistent end-to-end for the new screen.
- The `"episodes"` screen name in `_ScreenCache` is set by `_send_episode_list` and read correctly by `_send_fact_view`.
- `int(args[0]) if args else 0` with `except ValueError` in the `eps` dispatch handles all malformed callbacks safely.
- `log.exception` in `_send_episode_list` degrades gracefully without leaking state.
- The 22 new tests cover all branches described, including negative-space contracts and the multi-source search post-filter.

PR #413, opened 2026-04-30
feat(memory): expand episode detail view with approach/outcome/lessons/actors (#412)

### LINKED ISSUE #412: Expand episode detail view to render approach, outcome, lessons
## Problem

The `/memory` episode detail view renders only a subset of the fields stored on an episode. Operators looking at a stored episode see goal, context, `outcome_quality` (in the header), tags, and date. The remaining stored fields are invisible from this surface even though they carry the substance of the episode record.

## Currently shown

- Header: `Episode (<outcome_quality>)`
- Body: `goal` followed by `context`
- Tags
- Date

## Stored but not shown

- `approach` - the strategy used during the exchange
- `outcome` - narrative outcome text (distinct from `outcome_quality`, which is the success / partial / failure enum)
- `lessons` - optional takeaway, present when the stage-2 generator emitted one
- `actors` - participants
- `session_id` - originating session
- `episode_prompt_version` - the prompt revision that produced this record

## Proposed change

Expand the episode branch of `_build_fact_view` in `memory_command.py` so the detail view renders the substantive Sophia fields. Priority order:

1. `approach` and `outcome` - core narrative fields; their absence is the most felt gap.
2. `lessons` - render only when present; absence is the design-intended sentinel for "no lesson this time," so do not render a placeholder.
3. `actors` - small, useful for at-a-glance recall.

`session_id` and `episode_prompt_version` are operator-debugging fields rather than reading-an-episode fields. Defer until there is a concrete need; surfacing them risks visual clutter.

## Out of scope

- The fact view for `extracted` and `migration` rows. Unchanged.
- The episode list browser (`/memory` -> Episodes). Row format stays as-is; this issue is about the detail screen reached by tapping a row.
- Storage shape. All fields are already persisted; this is a presentation-only change.

## Acceptance

- Tapping an episode row opens a detail screen that renders `approach` and `outcome` as labeled rows.
- When the stored record contains `lessons`, the detail view renders it; when absent, the row is omitted entirely (no placeholder).
- `actors` renders as a comma-joined list under its own row.
- Tests pin all four behaviors, including the `lessons`-absent path.
- Existing tests on the extracted-row and migration-row branches continue to pass unchanged.

dcellison:
## Summary

- Adds `Approach`, `Outcome`, `Lessons` (when present), `Actors` rows to the episode branch of `_build_fact_view` so an operator looking at a stored episode can read its full record. The four rows splice between the body quote and the existing `Tags` / `Date` footer.
- Three contract-driven missing-data treatments: `or ""` defensive fallback for schema-required content-bearing fields (`approach`, `outcome`); presence-check (`is not None`) for the optional `lessons` so absence stays the design-intended sentinel for "no lesson this time"; `(none)` defensive fallback for the always-non-empty `actors`.
- `session_id` and `episode_prompt_version` stay out of scope (operator-debug fields, and `session_id` is routinely empty in storage; rendering would produce useless empty rows).
- Migration and extracted detail views unchanged.

Fixes #412.

## Test plan

- [x] `make test` (2645 tests pass; +9 new under `TestFactViewMultiSource`; `_episode_fact` helper extended with `approach` / `outcome` / `actors` / `lessons` kwargs).
  - `/memory` → tap `Episodes (N)` → tap a row. Verify the detail screen shows `Approach:`, `Outcome:`, `Actors:` rows; `Lessons:` row when the row has one stored, omitted otherwise.
  - Tap an extracted fact and a migration row to confirm their detail views are unchanged.

### DIFF
--- a/src/kai/memory_command.py
+++ b/src/kai/memory_command.py
@@ -674,11 +674,15 @@ def _build_fact_view(
 
     if source == "episode" or source == "migration":
         # Episode and migration rows share a minimal body shape:
-        # narrow Tags + Date with none of the extractor-only rows.
-        # Header differs (Episode (<outcome_quality>) vs Imported)
-        # but the rest of the layout is identical, so the tags
-        # line is computed once for both.
+        # Tags + Date with none of the extractor-only rows. Header
+        # differs (Episode (<outcome_quality>) vs Imported), and
+        # episodes (issue #412) splice in the four substantive
+        # Sophia fields (Approach, Outcome, Lessons-when-present,
+        # Actors) between the body quote and the Tags / Date footer.
+        # Migration rows skip the splice entirely - they have no
+        # equivalent metadata.
         tags_line = f"Tags:  {', '.join(tags) if tags else '(none)'}"
+        detail_lines: list[str] = []
         if source == "episode":
             # Episode rows surface the Sophia outcome_quality field as
             # a header parenthetical when present (e.g., "Episode
@@ -687,6 +691,30 @@ def _build_fact_view(
             # so episodes from the same day are distinguishable.
             quality = md.get("outcome_quality") or ""
             header = f"Episode ({quality})" if quality else "Episode"
+            # Substantive Sophia fields written by the stage-2
+            # generator (memory_extraction.py:1525-1538). approach /
+            # outcome / actors are schema-required and always present
+            # in production; the `or ""` / `or []` defensive fallbacks
+            # surface malformed-data corruption rather than hiding it
+            # (D3 in spec 412). lessons is optional-by-design and
+            # absent (not empty) when the generator chose to omit
+            # it - rendering `Lessons:  (none)` would lie about that
+            # design intent (D2).
+            approach = md.get("approach") or ""
+            outcome = md.get("outcome") or ""
+            actors_list = md.get("actors") or []
+            actors_str = ", ".join(actors_list) if actors_list else "(none)"
+            lessons = md.get("lessons")
+            detail_lines.append(f"Approach:  {approach}")
+            detail_lines.append(f"Outcome:  {outcome}")
+            # Presence-check, NOT truthy-check. The schema's
+            # minLength=20 makes the two equivalent in practice today,
+            # but `is not None` matches the storage contract literally
+            # (absent key is the sentinel for "no lesson this time")
+            # and protects against future schema relaxation.
+            if lessons is not None:
+                detail_lines.append(f"Lessons:  {lessons}")
+            detail_lines.append(f"Actors:  {actors_str}")
         else:
             # Migration rows already include the H3 title and section
             # structure inside `fact.text` (the migration script writes
@@ -700,6 +728,7 @@ def _build_fact_view(
             "",
             f'"{fact.text}"',
             "",
+            *detail_lines,
             tags_line,
             f"Date:  {_format_date(fact.created_at, with_time=True)}",
         ]
--- a/src/kai/memory_command.py
+++ b/src/kai/memory_command.py
@@ -692,14 +692,14 @@ def _build_fact_view(
             quality = md.get("outcome_quality") or ""
             header = f"Episode ({quality})" if quality else "Episode"
             # Substantive Sophia fields written by the stage-2
-            # generator (memory_extraction.py:1525-1538). approach /
-            # outcome / actors are schema-required and always present
-            # in production; the `or ""` / `or []` defensive fallbacks
-            # surface malformed-data corruption rather than hiding it
-            # (D3 in spec 412). lessons is optional-by-design and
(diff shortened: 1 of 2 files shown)


dcellison:
## Review by Kai

The PR is clean. Here are my findings, ordered by severity:

**No critical issues. No warnings.**

---

**Suggestion 1 - Untested fallback paths for `approach`/`outcome` absent from metadata**

`memory_command.py:706-707`:
```python
approach = md.get("approach") or ""
outcome = md.get("outcome") or ""
```
The `actors=[]` unreachable-state path gets an explicit pin test, but the analogous `approach`/`outcome` absent-key path (older episode rows pre-dating the schema addition) has no test. Given the helper already supports `approach=None` to omit the key from metadata, a test for the empty-string fallback render would be consistent with the coverage philosophy applied to `actors`. Minor omission.

**Suggestion 2 - `test_fact_view_episode_label_uses_two_spaces_after_colon` only exercises `approach`**

`test_memory_command.py:1054-1063`: The two-spaces-after-colon pin is valuable, but it only asserts on `Approach:`. `Lessons:` and `Actors:` use the same f-string pattern, so the test implicitly covers them, but `Outcome:` is never asserted. If a copy-paste error introduced `Outcome: ` (one space) in a future edit, the test would not catch it.

**Suggestion 3 - `actors_list` type is assumed but not guarded**

`memory_command.py:711`: `", ".join(actors_list)` silently produces character-joined output if a corrupt storage record put a bare string where a list belongs. Production schema enforces `list[str]`, so this is unreachable in practice, but it is inconsistent with the defensive posture applied to the other fields. `isinstance(actors_list, list)` guard or a `str` coercion isn't strictly needed, but the PR's own framing of D3 ("surface malformed-data corruption") argues for being equally defensive here.

---

Logic, control flow, and ordering are correct. The `detail_lines = []` / `*detail_lines` splice is clean. Migration rows correctly see an empty splice. The `is not None` presence-check for `lessons` is exactly right and the comment explaining why (vs. truthy-check) is useful. Test coverage of the new paths is thorough, including the field-order pin and the two unchanged-branch regression tests.

dcellison:
## Review by Kai

The prior review's two actionable suggestions have both been addressed in this revision. Here is what I found on a fresh pass:

**No critical issues. No warnings.**

---

**Suggestion - `actors_list` type guard (Suggestion 3 from prior review, still open)**

`memory_command.py:711`: `", ".join(actors_list)` still has no guard against a bare string arriving from corrupt storage. The prior review flagged this; the author has seen it and made their decision. Not re-raising it, just noting it remains unaddressed.

---

**One very minor observation (not a bug)**

`test_fact_view_episode_field_order` comment says "the default helper text contains 'Outcome:' in narrative form." The actual default outcome value in `_episode_fact` is `"CI ran on every PR within an hour."`, which does not contain the label `"Outcome:"`. The precaution of using a custom body text is sound defensive practice, but the comment's stated reason is slightly inaccurate. Does not affect correctness.

---

**What changed since the prior review:**

- `test_fact_view_episode_absent_approach_outcome_render_empty_fallback` was added, covering the absent-key `or ""` fallback path (prior Suggestion 1).
- `test_fact_view_episode_label_uses_two_spaces_after_colon` now asserts all four labels individually in both positive and negative forms (prior Suggestion 2).

Logic, control flow, and splice placement are correct. The `detail_lines = []` / `*detail_lines` pattern correctly produces no output for migration rows. The `is not None` guard on `lessons` is exact. The new tests are thorough and the field-order pin is particularly well constructed. Ready to merge.

dcellison:
## Review by Kai

The prior review covered this ground thoroughly across two passes. Here is my independent read:

**No critical issues. No warnings.**

**Suggestion - blank line between detail block and Tags footer**

`memory_command.py:728-731`: The splice inserts detail rows directly adjacent to `tags_line` with no separator:

```
"body text"
(blank)
Approach:  ...
Actors:  ...
Tags:  ...       ← no blank line before footer
Date:  ...
```

The body quote gets a blank line before and after it; the detail block gets none after it. Whether this is intentional is not verifiable from the diff alone, but it is asymmetric. If the format is operator-facing and scannability matters, a trailing `""` after `*detail_lines` (only when `detail_lines` is non-empty) would visually separate the substantive block from the footer. Minor style point; not a bug.

**Note on the `actors_list` type guard**

`memory_command.py:711`: Still unaddressed, as noted in both prior passes. The author has seen it and decided. Not re-raising.

**Everything else is correct.** The `detail_lines = []` / `*detail_lines` splice is clean and correctly produces an empty expansion for migration rows. The `is not None` guard on `lessons` is exact. The `or ""` / `or []` defensive fallbacks are consistent with stated intent. The test suite is thorough: field-order pin, two-space formatting pin (positive and negative, all four labels), absent-key fallback, empty-list fallback, and regression guards for both unchanged branches.

Issue #388, opened 2026-04-26
Single tag taxonomy across memory rows


dcellison:
## Background

Two prior discussions explain the memory system architecture and the mempalace-inspired tagging idea: #300 (semantic memory rationale) and #301 (mempalace-inspired additions, including topic pre-classification as a transparency feature).

## Context

The memory system currently has two tag-related concepts in flight:

1. **LLM-generated topical tags on episodes.** Stage-2 episode generation (#385, PR #387) attaches tags to each `source: episode` row at write time. Tags are stored on the row but not yet used for filtering at retrieval and not displayed in `/memory`.

2. **Topic pre-classification (#309).** A planned keyword-based tagger. Originally intended to pre-filter facts before vector search and to provide a fixed-vocabulary browse axis in `/memory`. Deferred after the memory evaluation work in #362 showed retrieval is robust without pre-filtering (the original retrieval-quality premise for #309 was disproven).

If both ship and both surface to end users, `/memory` ends up with two parallel tag taxonomies: one free-form LLM-generated set on episodes, one fixed-keyword set on facts. End users see two unrelated tag namespaces over the same memory store.

## Decision

One user-visible tag taxonomy across all memory rows. The LLM-generated tags from stage-2 episode generation become the single tag system. The keyword-based topic pre-classifier in #309 is retired (closed without implementation; reasoning preserved for future revisit).

## Concrete deltas

1. **Extend the stage-2 tag generator to fact extraction.** Tags currently attach only to `source: episode` rows at the moment they are created. Extend the same tag-generation step (or an equivalent prompt) to attach tags to `source: extracted` rows when facts are written. Implementation can reuse the existing tag prompt with content-appropriate framing.

2. **Close #309.** Add a closing comment explaining the deferral rationale (memory evaluation result in #362 falsified the retrieval-quality justification; the parallel-taxonomy concern is resolved by this issue) and link back to this issue.

3. **`/memory` UI v1: tags are not a primary browse axis.** Tags display as a per-row decoration (for example, "tagged: alpha, beta, gamma") and inform retrieval scoring. Users browse `/memory` by recency, source, or search; there is no "show me all memories tagged X" filter axis in v1. This sidesteps the unbounded-vocabulary concern (free-form LLM tags) for the v1 surface.

## Open question (deferred to implementation)

Vocabulary control for free-form LLM tags. Two candidate approaches:

[...]

PR #415, opened 2026-04-30
feat(memory): free-form LLM tags on extracted rows (#414)

### LINKED ISSUE #414: Sub A of #388: free-form LLM tags on extracted rows
## Sub-issue of #388 — Sub A: extractor

Replace the closed `_FACT_SCHEMA` tag enum with the same free-form tag generator already running on episode rows (stage-2, #385/PR #387). Implementation is contained to `memory_extraction.py`; no UI changes ride along (UI dismantle is Sub B). New `source: extracted` rows ship with free-form LLM tags. Existing rows keep their closed-vocab tags; mixed corpus is harmless under the decoration-only stance #388 settled on.

Closes the soft-vocab side of the parent issue's vocabulary-control decision (`#388` comment): the prompt seeds with a static list of preferred tags, advisory rather than enforced. Periodic dedup is Sub D.

## Concrete deltas

1. **Free-form tag schema on `_FACT_SCHEMA`.** Drop the `enum: [...]` constraint; replace with the same `minLength: 1, maxLength: 50, minItems: 1, maxItems: 5` shape `_EPISODE_SCHEMA` already uses. The `_FACT_SCHEMA["properties"]["facts"]["items"]["properties"]["tags"]` block becomes a free-form string array.

2. **Soft-vocab seed in the extraction prompt.** Add a paragraph to the stage-1 system prompt listing preferred tags ("prefer these N tags when applicable; only invent new ones when nothing fits"), seeded from the existing closed-vocab plus a small set of common operator-relevant additions. The seed is advisory; the LLM may add new tags.

3. **Bump `_EXTRACTION_PROMPT_VERSION`.** Existing rows under the old version stay queryable for cleanup. The new version is stamped on every row produced after deploy.

## Out of scope

- UI changes. The `_TAG_ENUM` constant in `memory_command.py`, the dashboard tag-button browse axis, and the by-tag stats table all stay untouched. Sub B handles the read-time changes.
- Re-tagging existing extracted rows. Decoration-only stance makes this unnecessary; a re-tag pass is a Sub D job (or operator-driven later).
- Periodic dedup. That is Sub D.
- Vocabulary-control mechanism beyond the static seed. The dedup job in Sub D maintains the seed over time; it is not part of this sub-issue.

## Acceptance

- `_FACT_SCHEMA`'s `tags` field is free-form (no enum).
- The stage-1 prompt includes a soft-vocab seed list of preferred tags.
- `_EXTRACTION_PROMPT_VERSION` is bumped.
- New `source: extracted` writes carry `tags: [...]` from the LLM-generated free-form set.
- Episode tag pipeline is unaffected (no changes to `_EPISODE_SCHEMA` or `_EPISODE_SYSTEM_PROMPT`).
- No regression in existing extraction-path tests; new tests cover the free-form schema and the prompt's seed list.

## Related

- Parent: #388
- Stage-2 episode tagger pattern this borrows: #385, PR #387
- Memory evaluation that informed the decision: #362

dcellison:
## Summary

- Replaces the closed-vocab tag enum on `_FACT_SCHEMA` with a free-form string array matching `_EPISODE_SCHEMA` (items free-form 1-50 chars; bounds 1-5).
- Rewrites the stage-1 extraction prompt with a soft-vocab seed paragraph (the prior nine values as preferred but advisory) and an explicit synonym-prohibition for the structurally significant `confirmed_action` magic string.
- Adds Rule 4b to `_validate_facts`: rejects `update_of` / `skip_redundant` consolidation against an existing row whose stored tags include `confirmed_action`. Defends the laundered-hallucination gate against synonym-tagged updates that would bypass Rule 4 (which keys off the new fact's tags).
- Threads a new `candidate_metadata: dict[str, dict]` parameter through `_validate_facts` and `_run_extractor`. The caller in `extract_and_store` builds it from the existing `candidates: list[MemoryResult]` already in scope.
- Bumps `_EXTRACTION_PROMPT_VERSION` 4 → 5; extends the version-history comment block.
- Folds in a pre-existing rule-4 docstring drift fix (mentioned only `skip_redundant`; code covers both consolidation intents).
- Sweeps 23 existing test call sites across `tests/test_memory_extraction.py`, `tests/test_memory_episodes.py`, and `tests/test_episode_classifier_eval.py` to add `candidate_metadata={}`. Deletes the obsolete `test_tag_enum_matches_extraction_schema`.

10 new tests across three classes pin the schema shape, the prompt seed and synonym-prohibition, the version bump and history extension, and Rule 4b's positive and negative space.

This is Sub A of #388; UI dismantle (Sub B) and periodic dedup (Sub D) remain on the parent issue's plan. Fixes #414.

## Test plan

- [x] `make test` (2655 tests pass).
  - Send a message that triggers fact extraction. Verify the resulting Qdrant row's metadata: `tags` is an array, `prompt_version` is `"5"`.
  - Send a message stating a fact whose topic falls outside the nine-value seed (e.g., a credential preference). Verify at least one tag is a free-form value, not one of the nine seed values.
  - Send a confirmation message. Verify the row carries `tags=["confirmed_action"]` literally, not a synonym.

### DIFF
--- a/src/kai/memory_extraction.py
+++ b/src/kai/memory_extraction.py
@@ -906,6 +906,7 @@ _CONSOLIDATION_INTENTS: frozenset[str] = frozenset({"new", "update_of", "skip_re
 def _validate_facts(
     facts: list[dict],
     candidate_ids: set[str],
+    *,
     candidate_metadata: dict[str, dict],
     user_id: str,
 ) -> list[dict]:
@@ -1070,7 +1071,15 @@ def _validate_facts(
         # dropped (issue #414), the validator carries that contract
         # explicitly.
         if intent in ("skip_redundant", "update_of") and isinstance(existing_id, str):
-            existing_meta = candidate_metadata.get(existing_id, {})
+            # `or {}` collapses both "key absent" and "key present
+            # with a None payload" to the same empty-dict default.
+            # `.get(key, {})` only handles the first case; a
+            # MemoryResult constructed with metadata=None (a type
+            # violation but possible via direct dataclass
+            # construction in tests or future code) would store
+            # None as the dict value, and the next `.get("tags")`
+            # call would AttributeError.
+            existing_meta = candidate_metadata.get(existing_id) or {}
             existing_tags = existing_meta.get("tags") or []
             if "confirmed_action" in existing_tags:
                 log.debug(
@@ -1321,7 +1330,7 @@ async def _run_extractor(
     facts_raw = payload_root.get("facts") or []
     has_episode = bool(payload_root.get("has_episode"))
     return ExtractionResult(
-        facts=_validate_facts(facts_raw, candidate_ids, candidate_metadata, user_id),
+        facts=_validate_facts(facts_raw, candidate_ids, candidate_metadata=candidate_metadata, user_id=user_id),
         has_episode=has_episode,
     )
 
--- a/src/kai/memory_extraction.py
+++ b/src/kai/memory_extraction.py
@@ -1071,15 +1071,7 @@ def _validate_facts(
         # dropped (issue #414), the validator carries that contract
         # explicitly.
         if intent in ("skip_redundant", "update_of") and isinstance(existing_id, str):
-            # `or {}` collapses both "key absent" and "key present
-            # with a None payload" to the same empty-dict default.
-            # `.get(key, {})` only handles the first case; a
-            # MemoryResult constructed with metadata=None (a type
-            # violation but possible via direct dataclass
-            # construction in tests or future code) would store
-            # None as the dict value, and the next `.get("tags")`
-            # call would AttributeError.
-            existing_meta = candidate_metadata.get(existing_id) or {}
+            existing_meta = candidate_metadata.get(existing_id, {})
             existing_tags = existing_meta.get("tags") or []
             if "confirmed_action" in existing_tags:
                 log.debug(
@@ -1959,7 +1951,15 @@ async def extract_and_store(
             # tags to detect a confirmation row being consolidated by
             # a synonym-tagged fact. The full `candidates` list is
             # already in scope, so building the dict here is one line.
-            candidate_metadata: dict[str, dict] = {c.id: c.metadata for c in candidates}
+            # `or {}` at construction collapses any null metadata
+            # payload to an empty dict so the `dict[str, dict]`
+            # annotation is truthful and Rule 4b's lookup can stay
+            # clean (no second `or {}` at the consumer site). The
+            # MemoryResult dataclass types metadata as `dict` rather
+            # than `dict | None`, so this defends against a direct-
+            # construction type violation rather than a real runtime
+            # path - cheap insurance.
+            candidate_metadata: dict[str, dict] = {c.id: c.metadata or {} for c in candidates}
             # Emit the candidate-set log line exactly once per
             # extraction call, AFTER the fetch and BEFORE the
             # subprocess spawn. Always emitted (even when empty) so
--- a/src/kai/memory_extraction.py
+++ b/src/kai/memory_extraction.py
@@ -50,7 +50,16 @@ log = logging.getLogger(__name__)
 # fact-storage parsing on the Python side is the same; the bump is
 # the canary that lets post-rollout log analysis distinguish facts
 # produced by the windowed prompt from facts produced by v3.
-_EXTRACTION_PROMPT_VERSION: str = "4"
+# v5: free-form tag schema (enum dropped). Soft-vocab seed in prompt
+# names the prior nine values as preferred but advisory; the LLM may
+# invent new tags when nothing fits. `confirmed_action` remains a
+# structurally significant magic string with explicit synonym
+# prohibition in the prompt; `_validate_facts` Rule 4b added to
+# defend the consolidation gate against synonym-tagged updates of
+# existing confirmation rows. `maxItems` raised from 4 to 5 to
+# match `_EPISODE_SCHEMA` per the parent issue's "single tag
+# taxonomy" decision.
+_EXTRACTION_PROMPT_VERSION: str = "5"
 
 # Sibling of _EXTRACTION_PROMPT_VERSION for stage-2 episode generation.
 # Stored in each episode's metadata so future cleanups can target a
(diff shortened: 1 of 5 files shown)


dcellison:
## Review by Kai

**Overall: clean, well-engineered PR.** One warning worth fixing before deploy, two minor suggestions.

---

### Warning - Potential `AttributeError` in Rule 4b

**`memory_extraction.py`, Rule 4b block (~line 1062 in the diff)**

```python
candidate_metadata: dict[str, dict] = {c.id: c.metadata for c in candidates}
```

If any `MemoryResult` has `metadata=None`, the dict stores `None` as the value. Later:

```python
existing_meta = candidate_metadata.get(existing_id, {})  # returns None, not {} -- key exists
existing_tags = existing_meta.get("tags") or []          # AttributeError: 'NoneType' has no attribute 'get'
```

The default `{}` in `.get(existing_id, {})` only applies when the key is absent. Because `candidate_id_set` and `candidate_metadata` are built from the same `candidates` list, by the time Rule 4b runs (Rule 3 has already verified `existing_id in candidate_ids`), the key will be present -- meaning a `None` payload crashes rather than safely defaulting.

Fix: use `or {}` at the lookup site:
```python
existing_meta = candidate_metadata.get(existing_id) or {}
```

Or defensively at construction: `{c.id: c.metadata or {} for c in candidates}`.

---

### Suggestion - Brittle prompt-parse test

**`tests/test_memory_extraction.py`, `test_extraction_prompt_seeds_exactly_nine_preferred_tags`**

The test whitespace-normalizes the prompt then finds the sentinel sentence and grabs everything until the **next `.`** in the string. If the prompt ever gains a period inside the seed region (abbreviation, etc.), the slice is wrong and the test either false-passes or fails for the wrong reason. Pinning the raw substring directly (e.g. `"preference, decision, fact, constraint, confirmed_action, project, location, schedule, relationship"` as a literal) would be simpler and more robust.

---

### Suggestion - `_validate_facts` parameter ordering

**`memory_extraction.py:906`**

The new `candidate_metadata` argument is inserted positionally between `candidate_ids` and `user_id`. All 23 call sites were correctly updated, but positional insertion between two same-direction args is a future footgun. Making `candidate_metadata` keyword-only (`*` separator before it, or at least documented as "pass by name") would make future call-site audits easier. Minor given the contained call surface.

---

Rule 4b logic itself is correct: Rule 3 fires before 4b (so `existing_id` is guaranteed in `candidate_ids` and therefore in `candidate_metadata`), Rule 4 fires before 4b (so the `confirmed_action`-tagged new-fact case is already handled), and the `or []` guard on `existing_tags` handles missing/empty tags cleanly -- modulo the `None` metadata case above.

dcellison:
## Review by Kai

All three concerns from the prior review are addressed: the `or {}` fix closes the `None` metadata crash, the prompt test uses a literal substring pin, and `*` makes `candidate_metadata` keyword-only. The logic is sound.

One new issue worth noting:

---

### Warning - Type annotation mismatch may fail Pyrich strict check

**`memory_extraction.py`, `extract_and_store`**

```python
candidate_metadata: dict[str, dict] = {c.id: c.metadata for c in candidates}
```

If `MemoryResult.metadata` is typed `dict | None`, the annotation `dict[str, dict]` is a lie - the dict can hold `None` values. The `or {}` in Rule 4b handles it correctly at runtime, but under Pyright strict (which this project uses) this could raise a type error depending on how `MemoryResult.metadata` is declared. The alternative the prior review offered - `{c.id: c.metadata or {} for c in candidates}` - would make the annotation truthful and remove the need for the defensive comment in Rule 4b.

---

### Suggestion - Dead fallback branch in synonym-prohibition test

**`tests/test_memory_extraction.py`, `test_extraction_prompt_includes_synonym_prohibition`**

```python
assert "MUST use the literal tag\n  `confirmed_action`" in _EXTRACTION_SYSTEM_PROMPT or (
    "MUST use the literal tag `confirmed_action`" in _EXTRACTION_SYSTEM_PROMPT
)
```

The fallback (no-newline form) can never match the actual prompt, which has a line break between "tag" and `` `confirmed_action` ``. So the second branch is dead code that creates false confidence in robustness. If someone reformats the prompt to remove that line break, the first branch fails and the second - which would now match - is unreachable due to short-circuit. Just the first form on its own is cleaner and equally correct.

---

Rule 4b logic, rule ordering, the 23 call-site sweep, and all new tests are correct. No security or convention violations.

dcellison:
## Review by Kai

Both issues from the prior review are addressed. The PR looks clean.

**Resolved since last review:**

- *Type annotation mismatch* (review 2 warning): Construction now uses `{c.id: c.metadata or {} for c in candidates}`, so the `dict[str, dict]` annotation is truthful and Rule 4b's `.get(existing_id, {})` is safe - values are never `None`.
- *Dead fallback branch in synonym-prohibition test* (review 2 suggestion): The test now whitespace-normalizes before the `"MUST use the literal tag \`confirmed_action\`"` assertion; the `or`-fallback form is gone.

**One minor observation - not a blocker:**

`memory_command._TAG_ENUM` is now an orphan: the test that verified it matched the schema was correctly deleted (the schema has no `enum` field to compare), but `_TAG_ENUM` itself remains in `memory_command.py` and is presumably still driving `/memory stats`. The PR description explicitly defers the UI dismantle to Sub B, so this is intentional. Worth a comment in `memory_command.py` near `_TAG_ENUM` marking it as stale pending Sub B, but that is a suggestion for the sub-issue, not this PR.

**Everything else is correct:**

- Rule 4b ordering (after Rules 3 and 4) is sound; `existing_id` is guaranteed in `candidate_metadata` by the time 4b evaluates.
- The `isinstance(existing_id, str)` guard in 4b is redundant but harmless.
- All 23 call-site updates correctly use `candidate_metadata=` as keyword.
- Schema shape, prompt soft-vocab, synonym prohibition, version bump, and version history are all pinned by the new tests with appropriate robustness (literal substring, whitespace-normalized, four-synonym loop).
- `test_extraction_prompt_version_history_extended` reads source via `Path(__file__).read_text()` - safe for a source install; a compiled-only install would fail, but that is not a realistic deployment path here.

Approved.

dcellison:
## Review by Kai

Looking at the current diff against the prior review thread.

**One discrepancy worth flagging - otherwise clean.**

---

### Warning - `candidate_metadata` construction does not match review 3's stated fix

**`memory_extraction.py`, `extract_and_store`**

Review 3 explicitly stated the type-annotation/None-crash concern from review 2 was resolved with:
```python
{c.id: c.metadata or {} for c in candidates}
```

The current diff has:
```python
candidate_metadata: dict[str, dict] = {c.id: c.metadata for c in candidates}
```

The inline comment claims `MemoryResult.metadata` is typed `dict` (not `dict | None`), which would make `or {}` unnecessary and the annotation truthful. If that's accurate, this is intentional and fine. But it directly contradicts what review 3 described as the resolution, which means either review 3 approved the wrong code, or this was revised after that approval. It is worth verifying once against the actual `MemoryResult` declaration in mem0 before deploy - if `metadata` is `dict | None` at runtime, Rule 4b's `existing_meta.get("tags")` crashes on any candidate with a null payload.

---

**Everything else is correct.** The Rule 4b logic, rule ordering, the 23 call-site sweep to keyword-only, schema shape, prompt soft-vocab seed, synonym prohibition, version bump, and all new test assertions are sound. No security or convention violations.

PR #417, opened 2026-04-30
feat(memory): drop /memory tag browse axis (#416)

### LINKED ISSUE #416: Sub B of #388: drop /memory tag browse axis
## Summary

Sub B of #388. Dismantle the closed-vocabulary tag-button browse axis from `/memory`. After Sub A (#414) replaced the closed `_TAG_ENUM` with free-form LLM tags on extracted rows, the tag-button surface is the last reference to the closed enum on the read side; the parent decision (#388) explicitly defers tag-as-filter UI for v1. Tags become row decoration only.

## Scope

1. **Dashboard.** Drop the per-tag inline-button rows from `_build_dashboard`. Headline drops the trailing `across N tags` clause. Footer collapses to the no-tag-buttons branches that already exist (Episodes / Search / Stats utility row). Empty-state guard unchanged.
2. **Tag drill-down.** Delete `_build_tag_view`, `_send_tag_view`, and the `tag` callback verb. The `_paginate` helper stays (still used by the episode list browser); rename `_TAG_PAGE_SIZE` to `_PAGE_SIZE`.
3. **Stats.** Drop the `By tag:` table from `_build_stats`. Confidence block, confirmed-actions line, and prompt-version table stay.
4. **`_TAG_ENUM`.** Delete the constant.
5. **`/memory forget <tag>` text path.** Drop the subcommand. Show help text on `forget` instead. The forget-by-tag bulk operation, the `ftd` callback verb, and `_build_forget_tag_confirm` go with it.
6. **Fact-view tag decoration.** Already shipping on every source: extracted rows from the original `/memory` build (#310), episode and migration via the multi-source surface (#407). Verify only; no new rendering work expected.
7. **Help text.** Drop the `/memory forget <tag>` line.

## Acceptance

- [ ] Dashboard renders no per-tag button rows for any operator (extracted-only, mixed, episode-only).
- [ ] `/memory <tag>` and the `tag` callback are gone; tapping a stale tag-button callback in chat history dismisses gracefully.
- [ ] `/memory stats` shows source-totals, confidence, confirmed-actions, and prompt-versions; no by-tag table.
- [ ] `/memory forget <tag>` returns help text. The `ftd` callback path is gone.
- [ ] Fact-view detail surface still shows `Tags:` on extracted, episode, and migration rows (no regression from #407 / #412).
- [ ] `_TAG_ENUM`, `_build_tag_view`, `_send_tag_view`, `_build_forget_tag_confirm`, `_send_forget_tag_confirm` are deleted from `memory_command.py`.
- [ ] Tests for the deleted surfaces removed. No new test failures.

## Out of scope

- **Browse-by-source for extracted/migration rows.** With tag drill-down gone, extracted facts are reachable only via search. The parent explicitly defers any new browse axis for v1.
- **Tag-based retrieval scoring** (already in place; no changes).
- **Vocabulary control** (deferred until tags become a filter axis; Sub D is the dedup pass when it ships).
- **Re-tagging legacy `_TAG_ENUM` rows.** Mixed-vocabulary corpus is harmless under decoration-only role.

## Migration / rollout

No data migration. Pure code change. Existing rows keep whatever tags they were written with; the fact-view surfaces them as decoration.

## Tradeoff to surface

After Sub B ships, an extracted-only operator (no episode rows) has no list-style browse axis; only Search and Stats remain. This is the deliberate v1 stance from #388. If operators report this as a regression, the response is a separate issue for a recency-ordered "all extracted" browse list, not reviving the tag drill-down.

dcellison:
## Summary

Closes #416. Sub B of #388.

After Sub A (#414) replaced the closed `_TAG_ENUM` with free-form LLM tags on extracted rows at write time, the tag-button surface was the last reference to the closed enum on the read side. The parent decision (#388) explicitly defers tag-as-filter UI for v1: tags become row decoration only, surfaced on the fact-view detail screen. `/memory` browses by source (Episodes browser plus Search) instead.

## What changed

**Code deletions in `src/kai/memory_command.py`:**

- `_TAG_ENUM` constant (and preamble comment).
- `_build_tag_view`, `_send_tag_view`, the `tag` callback verb.
- `_build_forget_tag_confirm`, `_send_forget_tag_confirm`, the `ftd` callback verb.
- The `/memory forget <tag>` text-command branch.
- The per-tag dashboard button rows.
- The `across N tags` headline suffix.
- The `By tag:` table from `/memory stats`.
- The `_ScreenCache.tag` field.

**Code edits:**

- `_TAG_PAGE_SIZE` renamed to `_PAGE_SIZE` (5 textual occurrences in `_paginate`).
- `_build_dashboard` footer collapses from 4 branches to 2.
- `_send_fact_view` loses the tag return-to branch; the `elif` chain converts to `if`/`elif`.
- `_OUTCOME_QUALITY_ENUM` preamble rewritten with self-contained rationale.
- Docstring sweep across `memory_command.py`, `memory.py`, `tests/test_memory.py`, and `tests/test_memory_command.py` to remove cross-references to the deleted surface.

**Tests:**

- Deleted `TestBuildTagView` and `TestBuildForgetTagConfirm` classes plus their section headers.
- Deleted the `tag` / `ftd` verb dispatch tests, the `/memory forget` text-path tests, and the `_send_forget_tag_confirm` contract test.
- Rewrote `test_send_failure_yields_single_query_failed_answer` to use the `ffc` verb (pointing at `_send_forget_fact_confirm`'s direct `get_by_id` call), preserving the answer-after-send invariant pin.
- Renamed and rewrote `test_ffd_deletes_and_returns_to_tag` to `test_ffd_deletes_and_returns_to_episode_list`, swapping `return_to=("eps", ["0"])` and `get_all_episodes` mock. Closes a pre-existing eps coverage gap from #410.
- Added 10 new regression pins under `TestTagBrowseAxisRetired`: dashboard renders no per-tag buttons, headline omits `across N tags`, stats omits `By tag:`, help text omits `/memory forget`, `forget` subcommand returns help, unknown verb dismisses silently, fact-view renders Tags row on extracted/episode/migration sources, retired constant is absent.

**Tradeoff baked into the v1 surface:**

After this change, an extracted-only operator (no episode rows) has no list browser at all; only Search and Stats remain in the keyboard. This is the deliberate stance in #388 - if operators report it as a regression, the response is a separate issue for a recency-ordered "all extracted" browse list, not reviving the tag drill-down.

## Diff stat

```
 src/kai/memory.py            |  30 +--
 src/kai/memory_command.py    | 514 +++++++-----------------------------
 tests/test_memory.py         |  25 +-
 tests/test_memory_command.py | 612 ++++++++++++++++++-------------------------
 4 files changed, 367 insertions(+), 814 deletions(-)
```

Net: 447 lines deleted across four files.

## Verification gates

Both spec-mandated gates clean across `src/kai/` and `tests/`:

- Symbol-name gate (`_TAG_ENUM`, `_send_forget_tag_confirm`, `_build_forget_tag_confirm`, `_send_tag_view`, `_build_tag_view`): zero hits.
- Natural-language gate (`tag drill-down`, `tag view`, `tag-button`, `forget-by-tag`, `by-tag`, `off-enum tag`): zero hits.

The one regression test that pins `_TAG_ENUM`'s absence (`test_retired_tag_enum_constant_is_absent`) constructs the constant name via string concatenation so the source of the test itself does not register as a stale reference under the gate.

## Test plan

- [x] `make test` (2646 passed, 1 skipped).
- [x] `make check` (ruff check + ruff format clean).
- [ ] After deploy: send `/memory` with an extracted-only operator with multiple tags. Verify the keyboard renders only the utility row (Search, Stats; no Episodes button). Verify the headline reads `Memories: N facts.` with no `across N tags` suffix.
- [ ] After deploy: send `/memory stats`. Verify there is no `By tag:` section. Confidence, Confirmed actions, Prompt versions sections all render as before.
- [ ] After deploy: send `/memory forget` and `/memory forget preference`. Verify each replies with the help text.
- [ ] After deploy: tap a fact to open its detail view. Verify the `Tags:` row renders with the row's tags (extracted, episode, and migration sources all show this row).
- [ ] After deploy: have a stale `/memory` dashboard from before deploy in chat history. Tap one of the old per-tag buttons. Verify no error toast and no message changes; the loading spinner clears and the dashboard stays visible.

### DIFF
--- a/src/kai/memory.py
+++ b/src/kai/memory.py
@@ -214,7 +214,7 @@ class MemoryStats:
     not enumerated as a separate per-source field. Confidence aggregates
     and `by_tag` stay scoped to extracted (see comments below) because
     episode and migration rows do not carry confidence and use tag
-    spaces that are not enumerated by `_TAG_ENUM` in memory_command.py.
+    spaces that are independent of the extractor's tag vocabulary.
     """
 
     total_count: int
@@ -1100,28 +1100,25 @@ def get_all(*, user_id: str, limit: int | None = 1000) -> list[MemoryResult]:
 def get_by_tag(*, user_id: str, tag: str) -> list[MemoryResult]:
     """Return user-visible facts for `user_id` carrying `tag`.
 
-    Used by the /memory tag drill-down (spec 310 §6.2). Filters
-    client-side because Mem0's `get_all` does not accept metadata
-    filters. Two filter clauses, both load-bearing:
+    Data-layer entry point for tag-keyed lookups. Filters client-side
+    because Mem0's `get_all` does not accept metadata filters. Two
+    filter clauses, both load-bearing:
 
       - `metadata.source in USER_VISIBLE_SOURCES`: defends against
         legacy ""-source (or any future-additional non-UI) rows
         leaking into the operator-facing surface. Issue #407 expanded
         this from `== "extracted"` to the three-source set so episode
-        and migration rows that happen to carry an enum tag in their
-        own metadata become browsable from the dashboard's tag
-        buttons. The dashboard's `_TAG_ENUM` gate still limits which
-        tags get rendered as buttons; this filter only governs which
-        rows match a tapped tag.
+        and migration rows that happen to share an extractor tag in
+        their own metadata become reachable through this helper.
       - `tag in metadata.tags`: the actual tag match. The `or []`
         guards against a malformed row that lacks the tags list
         entirely; such rows simply do not match any tag.
 
     Sort: `updated_at` descending. A re-extracted fact bubbles to the
-    top of its tag list, which is the spec's intended drill-down
-    ordering (§6.2). Falls back to `created_at` for rows that are
-    missing `updated_at`; both default to "" in `_wrap_result` so the
-    sort is total-order stable rather than raising on None comparisons.
+    top of its tag list. Falls back to `created_at` for rows that
+    are missing `updated_at`; both default to "" in `_wrap_result`
+    so the sort is total-order stable rather than raising on None
+    comparisons.
 
     The full row set comes from `get_all(limit=None)` so a user with
     thousands of extracted facts still gets a complete tag listing.
@@ -1145,10 +1142,9 @@ def get_all_episodes(*, user_id: str) -> list[MemoryResult]:
     """Return all episode-source memories for `user_id`, newest first.
 
     Read-side primitive for the /memory dashboard's episode-list
-    browser (issue #410). Episode rows mostly carry Sophia-style tags
-    that are not in `_TAG_ENUM`, so the existing tag-button surface
-    cannot reach them; this helper backs a dedicated screen that
-    enumerates every episode in one place.
+    browser (issue #410). Episode rows surface here in one place,
+    independent of the extractor's tag vocabulary, since the dashboard
+    browses by source rather than by tag.
 
     The source filter here is intentionally narrower than
     `USER_VISIBLE_SOURCES`: this function's purpose is single-source
--- a/src/kai/memory.py
+++ b/src/kai/memory.py
@@ -1100,9 +1100,14 @@ def get_all(*, user_id: str, limit: int | None = 1000) -> list[MemoryResult]:
 def get_by_tag(*, user_id: str, tag: str) -> list[MemoryResult]:
     """Return user-visible facts for `user_id` carrying `tag`.
 
-    Data-layer entry point for tag-keyed lookups. Filters client-side
-    because Mem0's `get_all` does not accept metadata filters. Two
-    filter clauses, both load-bearing:
+    Public data-layer entry point for tag-keyed lookups. Kept as a
+    primitive in the `kai.memory` API surface even though no UI path
+    in this repo currently calls it: external integrations and future
+    surfaces (admin tooling, batch reports) want to reach memory rows
+    by tag, and reimplementing the filter+sort+source-gate logic per
+    caller is the kind of duplication this helper exists to prevent.
+    Filters client-side because Mem0's `get_all` does not accept
+    metadata filters. Two filter clauses, both load-bearing:
 
       - `metadata.source in USER_VISIBLE_SOURCES`: defends against
         legacy ""-source (or any future-additional non-UI) rows
--- a/src/kai/memory_command.py
+++ b/src/kai/memory_command.py
@@ -451,11 +451,8 @@ def _build_episode_list_view(
             quality_str = "[----]"
         date_str = _format_date(fact.updated_at or fact.created_at)
         lines.append(f"{idx}.  {quality_str}  {_truncate(fact.text)}")
-        # Twelve-space indent on the date row visually nests it under
-        # the bracket-prefixed text row above; the longest enum value
-        # (`[failure]`, nine chars) plus the leading "N.  " plus the
-        # gap between the bracket and the text lands the body text at
-        # column 12.
+        # Twelve-space indent visually nests the date under the
+        # bracket label on the line above.
         lines.append(f"            {date_str}")
         lines.append("")
         memory_ids.append(fact.id)
--- a/src/kai/memory_command.py
+++ b/src/kai/memory_command.py
@@ -64,30 +64,9 @@ log = logging.getLogger(__name__)
 
 # ── Constants ───────────────────────────────────────────────────────
 
-# The closed enum of tags accepted by the Haiku extractor, in the same
-# order as `_FACT_SCHEMA["properties"]["facts"]["items"]["properties"]
-# ["tags"]["items"]["enum"]` at memory_extraction.py:153-163. Mirrored
-# here (rather than imported) because importing memory_extraction would
-# pull in the Anthropic SDK, an unnecessary dependency for the UI path.
-# A drift between the two lists shows up immediately in `/memory stats`
-# (a tag with rows but no enum entry would be silently absent from the
-# stats view), and is also covered by a unit test in test_memory_command.
-_TAG_ENUM: tuple[str, ...] = (
-    "preference",
-    "decision",
-    "fact",
-    "constraint",
-    "confirmed_action",
-    "project",
-    "location",
-    "schedule",
-    "relationship",
-)
-
-# Page size for the tag drill-down view (spec §6.2). Five fits on a
-# single phone screen with the surrounding buttons; raising it forces
-# scrolling.
-_TAG_PAGE_SIZE = 5
+# Page size for paginated list views. Five fits on a single phone
+# screen with the surrounding buttons; raising it forces scrolling.
+_PAGE_SIZE = 5
 
 # Default number of search hits returned by `/memory search`, before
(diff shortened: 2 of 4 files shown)


dcellison:
## Review by Kai

The file is on a feature branch not yet checked out locally - I'll work from the diff directly.

---

## Review

**Overall:** The dismantle is thorough and well-structured. Tests pass (2646), lint is clean. A few issues found:

---

### Warning: Tautological assertion in `test_unknown_callback_verb_silently_dismisses`

**`tests/test_memory_command.py` - new `TestTagBrowseAxisRetired` class**

```python
call = upd.callback_query.answer.call_args
assert call.args == () or call.args == ()
```

Both sides of the `or` are identical - `P or P` reduces to `P`. This is almost certainly a copy-paste error, likely intended as `assert call.args == () and call.kwargs == {}` (checking neither positional toast text nor keyword args like `show_alert=True` were passed). As written, the `or` branch is dead code. The assertion technically still checks that no positional args were passed, but it silently drops the kwargs check, leaving a gap: a call like `query.answer(show_alert=True)` would pass when it should not.

---

### Suggestion: Revision-note comment left in test

**`tests/test_memory_command.py` - `test_dashboard_renders_no_per_tag_button_rows`**

```python
# specific enough to avoid false-matching the Episodes
# button label `Episodes (4)` (also a single-button
# row possibility) - no, actually that DOES match.
# The fix: exempt the Episodes button label explicitly,
```

The `"- no, actually that DOES match."` fragment documents a design decision mid-revision. The conventions say not to document reasoning that belongs in the PR description. Should just state the rule: "Episodes button has the same shape, so exempt it by prefix."

---

### Suggestion: `get_by_tag` docstring leaves an orphaned forward reference

**`src/kai/memory.py` - `get_by_tag` docstring (line ~1100 in original)**

The updated docstring says the helper is a "Data-layer entry point for tag-keyed lookups" but gives no indication of what currently calls it post-dismantle. The function is kept (correct - it's a data-layer primitive), but the docstring could note that callers exist outside the UI path, otherwise a future reader may assume it's dead code and delete it.

---

### Clean areas

- The `ftd`/`tag` verb removal is complete with no dangling references in production code
- Stale callbacks from pre-deploy dashboards fall through to the unknown-verb branch and dismiss cleanly - correct safe default
- The `_TAG_ENUM` retirement test using string concatenation (`"_TAG" + "_ENUM"`) to avoid triggering its own grep gate is clever and correct
- `ffd` return_to simplification correctly falls back to dashboard for stale `("tag", ...)` entries in cache
- Security surface is net-reduced: the only open-string-to-data-layer path (`ftd`'s `get_by_tag`/`delete_by_id` loop gated on `_TAG_ENUM`) is gone entirely

The tautology in `test_unknown_callback_verb_silently_dismisses` is the only finding worth fixing before merge.

dcellison:
## Review by Kai

The prior review's single blocking finding is fixed, and both suggestions are addressed. Here is the fresh assessment:

**Prior findings resolved:**

- **Tautological assertion** (`call.args == () or call.args == ()`): corrected to `call.args == () and call.kwargs == {}` in `test_unknown_callback_verb_silently_dismisses`. The kwargs gap is now closed.
- **Revision-note comment** in `test_dashboard_renders_no_per_tag_button_rows`: the "no, actually that DOES match" fragment is gone; the comment states the rule cleanly.
- **`get_by_tag` docstring**: now explains the function is kept as a data-layer primitive for external/future callers and is not dead code.

**New review:**

No critical or warning-level findings. A few minor observations:

**Suggestion: `test_dashboard_renders_no_per_tag_button_rows` exemption is untested** (`tests/test_memory_command.py`, `TestTagBrowseAxisRetired`)

The `label.startswith("Episodes (")` exemption path is never exercised because the `_stats` call passes no `episode_count` (defaults to 0), so the Episodes button never renders and no single-button row ever appears. The loop body is vacuously satisfied. The exemption logic is correct and defensive against a future layout change, but the test doesn't prove the exemption fires as intended. Not a bug - the regression pin still works - but worth noting.

**`ffd` handler: `("search", ...)` return_to silently falls through to dashboard** (`src/kai/memory_command.py`, `ffd` branch)

A fact opened from a search result and deleted will return to the dashboard, not back to search. This is pre-existing behavior unchanged by the PR (the old code only handled `tag` and `eps` in the `ffd` return path). Not a regression introduced here.

**Clean areas:**

- Stale pre-deploy `mem:tag:*` callbacks hit the unknown-verb branch and dismiss silently with `query.answer()` and no edit - correct, tested, and safe.
- Stale `("tag", ...)` entries in `cache.return_to` fall through to dashboard in the `ffd` path - correct.
- Security surface is net-reduced: the only user-controlled-string-to-data-layer path (`ftd` routing an arbitrary tag string into `get_by_tag`/`delete_by_id`) is gone.
- The `_TAG_ENUM` retirement test using string concatenation (`"_TAG" + "_ENUM"`) to avoid self-triggering the grep gate is correct and clever.
- `test_send_failure_yields_single_query_failed_answer` rewrite using `ffc`/`get_by_id` is well-chosen: `ffc` has no inner guard so the exception propagates to the outer `except` as required by the test.

PR is clean. The tautology fix is the only change needed before merge and it is in.

dcellison:
## Review by Kai

This PR has already gone through two rounds of review with all blocking findings resolved. I reviewed the diff fresh against the current state.

**No critical or warning findings.**

---

**Observations (not blocking):**

**`test_dashboard_renders_no_per_tag_button_rows` - exemption still untested** (`tests/test_memory_command.py`, `TestTagBrowseAxisRetired`)

The prior review flagged this as a suggestion, and it remains unaddressed. The `label.startswith("Episodes (")` exemption path never fires because the test passes no `episode_count` (defaults to 0), so no single-button row ever appears. Not a bug - the regression pin works - but the exemption is decorative.

---

**Clean areas:**

- `_send_fact_view` `if`/`elif` conversion is semantically correct: the old `if screen == "tag"` → `elif screen == "episodes"` becomes `if screen == "episodes"` because the tag branch was the only preceding guard. Logic preserved. ✓
- `ffd` return_to simplification correctly falls through to dashboard for stale `("tag", [...])` entries. ✓
- `test_send_failure_yields_single_query_failed_answer` rewrite is well-chosen: `ffc` → `_send_forget_fact_confirm` → `get_by_id` has no inner guard, so RuntimeError propagates to the outer `except` as the test requires. ✓
- `test_longest_realistic_callback_under_64_bytes` updated to `mem:eps:999` (12 bytes) - correct; `mem:fact:999` is also 12 bytes, both are the longest remaining verbs post-dismantle. ✓
- The `_OUTCOME_QUALITY_ENUM` comment correctly removes the now-stale `_TAG_ENUM` comparison while keeping the import-avoidance rationale intact. ✓
- Security surface is net-reduced: the only user-controlled-string-to-data-layer path (`ftd` routing a callback-embedded tag string into `get_by_tag`/`delete_by_id`) is gone entirely. ✓
- `TestTagBrowseAxisRetired` regression pins are well-structured and cover the key dismantle assertions (no per-tag buttons, no `across N tags`, no `By tag:` in stats, help text, unknown verb, Tags row on all sources, `_TAG_ENUM` absence). ✓

PR is clean. Ready to merge.

PR #419, opened 2026-04-30
feat(memory): periodic tag dedup script (#418)

### LINKED ISSUE #418: Sub D of #388: periodic tag dedup script
## Summary

Sub D of #388. Admin-side script that walks the tag distribution across all `source: extracted` and `source: episode` rows, identifies near-duplicates, and rewrites tags to a canonical form. After Sub A (#414) replaced the closed `_TAG_ENUM` with free-form LLM tags on extracted rows and Sub B (#416) retired the dashboard tag-button surface, the tag space is now unbounded. The dedup script is the agreed vocabulary-control mechanism (option (a) from #388's open question): a soft-vocab seed in the extractor prompt plus this periodic dedup pass.

## Scope

`scripts/tag-dedup.py` (or similar). One-shot CLI. Default mode is dry-run; an explicit `--apply` flag is required to write.

1. **Walk the tag distribution.** Across `source: extracted` and `source: episode` rows for the target user(s). Counts per tag value, with case preserved.
2. **Identify near-duplicate clusters.** Three classes (priority order):
   - Case-insensitive equality (`Preference` and `preference`).
   - Plural/singular pairs (`preferences` and `preference`); also -ies/-y, -es/- variants.
   - Optional fuzzy similarity for close neighbors (configurable threshold). Default off in v1.
3. **Choose a canonical form per cluster.** Most-frequent wins; ties broken by lexicographic order. Surfaces in the dry-run report so the operator can override before applying.
4. **Rewrite via `Memory.update`.** Mem0's `update(memory_id, data, metadata=None)` keeps the memory_id and history; the script passes the existing row text as `data` (Mem0 recomputes the embedding regardless, so this is one embedding API call per affected row).
5. **Print the report.** Tag distribution before, proposed merges (cluster -> canonical, with row counts), distribution after. Same shape in dry-run and `--apply` modes; `--apply` adds a "rewrote N rows" tail.

## Acceptance

- [ ] `scripts/tag-dedup.py --chat-id <id>` runs in dry-run by default and prints the proposed merges without touching the store.
- [ ] `scripts/tag-dedup.py --chat-id <id> --apply` rewrites tags via `Memory.update` and reports the rewrite count.
- [ ] `confirmed_action` is preserved unconditionally on extracted rows. The dedup pass never collapses a tag list in a way that drops `confirmed_action` or substitutes a synonym for it (the prompt-side and validator-side guards from Sub A's R4b are belt-and-suspenders here; the script adds explicit script-side gating as a third layer).
- [ ] Per-user scope is the default. Cross-user execution is opt-in (e.g., `--all-users` flag) and requires explicit operator instruction.
- [ ] Tests cover the cluster-detection logic (case, plural/singular, threshold), the canonical-form selection, the `confirmed_action` protection, and the dry-run vs apply paths. Mock `Memory.update` to avoid real embedding calls.
- [ ] No changes to the inner extractor, the runtime memory API, or the `/memory` UI.

## Open questions (resolve at spec time)

1. **Plural/singular detection.** Is a small hand-rolled rule set (`s`, `es`, `ies`) sufficient for the v1 corpus, or do we want to pull in `inflect` or similar? Hand-rolled is the smaller dependency footprint; `inflect` handles edge cases (`fish`/`fish`, `child`/`children`) but the tag space here is unlikely to need them.
2. **Fuzzy similarity strategy.** If we ship the optional fuzzy threshold, what's the metric? Levenshtein over normalised forms is cheap and predictable; embedding similarity is heavier and dependency-heavy. Default off in v1 either way; lean toward Levenshtein when it ships.
3. **Operator override of canonical form.** Is a pre-merge prompt acceptable for an admin script, or do we want a config-file mapping? A prompt fits the manual-first, scheduled-later flow; a config file fits a future scheduled run. v1 is manual; a prompt is the smaller change.
4. **Audit log.** Does the script write a JSONL of `(memory_id, old_tags, new_tags, applied_at)` for rollback? Mem0's `update` overwrites without history of the old tags. An audit log would be cheap to add and is the kind of safety net that makes the destructive `--apply` path less scary.

## Out of scope

- **Scheduling.** v1 is manual. A future change can wire it through `/api/schedule` if useful.
- **Seed-list rebuild.** Refreshing the extractor's soft-vocab seed list from corpus stats is operator-driven for now; out of scope here.
- **Cross-source tag harmonisation.** If episode rows and extracted rows happen to use slightly different tag values for the same concept, the script does NOT bridge them; it dedups within each source's tag space. A future change can layer cross-source merging on top.
- **Tag retagging via the LLM.** The script is purely string-level (case, plural/singular, optional Levenshtein). No LLM calls.

dcellison:
## Summary

Closes #418. Sub D of #388.

After Sub A (#414) replaced the closed `_TAG_ENUM` with free-form LLM tags and Sub B (#416) retired the dashboard tag-button surface, the tag space is unbounded. This PR adds the agreed vocabulary-control mechanism (option (a) from #388's open question): an admin-side periodic dedup script.

`scripts/tag-dedup.py` walks the tag distribution across `extracted` and `episode` rows, identifies near-duplicate clusters, and rewrites them to a canonical form via a new `memory.update_metadata` wrapper. Default mode is dry-run; `--apply` is required to write.

## What changed

**New file `scripts/tag-dedup.py`:**

- argparse CLI: `--chat-id` and `--all-users` are mutually exclusive; `--apply` flips dry-run off; `--no-prompt` skips the interactive override; `--audit-log` overrides the default audit-log path.
- Three cluster-detection rules: case-insensitive equality (`Preference` vs `preference`); plural/singular pairs (`-s`, sibilant `-es`, `-ies/-y`); within-row duplicate collapse.
- Canonical-form selection: most-frequent wins, lexicographic tiebreak.
- Two-layer `confirmed_action` defense: cluster-skip in `identify_clusters` (production path) plus a row-level guard in `apply_cluster_to_row` (defense-in-depth against future rule additions that bypass cluster-skip).
- Migration rows excluded from the dedup pass (different vocabulary space).
- Read-merge-write at the apply call site preserves every non-`tags` metadata field.
- JSONL audit log on `--apply`.

**New `memory.update_metadata` wrapper (in `src/kai/memory.py`):**

Thin wrapper around Mem0's `Memory.update`. The docstring pins two contracts verbatim:

- **Metadata-overwrite contract.** Mem0's underlying `update` REPLACES the metadata dict wholesale. Auto-preserved fields are `data`, `hash`, `text_lemmatized`, `created_at`, `updated_at`, `user_id` (when absent in new), `agent_id` (when absent in new), `run_id` (when absent in new), `actor_id` (always), `role` (when absent in new). Every other field is destroyed unless the caller passes it explicitly. Callers wanting partial updates must read-merge-write at the call site.
- **Source-scope contract.** The user-id gate admits any row whose `metadata.source` is in `USER_VISIBLE_SOURCES` (extracted, episode, migration). Callers wanting narrower scope must filter at the caller level. The `confirmed_action`-tag protection used by the tag dedup pass is the caller's responsibility, not the wrapper's.

**Tests:**

- `tests/test_tag_dedup.py`: 29 tests across 6 classes (`TestComputeDistribution`, `TestIdentifyClusters`, `TestCanonicalSelection`, `TestApplyClusterToRow`, `TestApplyRewrites`, `TestCliDispatch`).
  - The row-level `confirmed_action` guard is verified via a fixture-driven test that constructs a `Cluster` mapping `confirmed_action` to something else, bypassing the cluster-skip; the row stays untouched.
  - `TestApplyRewrites` verifies the read-merge-write discipline: every non-`tags` metadata field on the original row appears on the call to `update_metadata`.
- `tests/test_memory.py::TestUpdateMetadataWrapper`: 7 tests pinning ownership/source gates, Mem0 exception handling, and the wrapper's destructive-by-design metadata semantics (sparse-metadata-passes-through-unchanged so a future "helpful" auto-merge refactor cannot land silently).

**`.gitignore`:**

- `scripts/.tag-dedup-audit-*.jsonl` (audit logs from `--apply` runs).

## Diff stat

```
 .gitignore                |   3 +
 scripts/tag-dedup.py      | 480 +++++++++++++++++++ (new)
 src/kai/memory.py         |  50 ++
 tests/test_memory.py      | 175 +++++++++
 tests/test_tag_dedup.py   | 588 ++++++++++++++++++++++ (new)
 5 files changed, 1296 insertions(+)
```

## Verification

- `make test`: 2682 passed (36 new), 1 skipped (pre-existing).
- `make check`: ruff check + ruff format both clean.
- Personal pre-push gate (spec refs, cross-file line numbers, em/en dashes) clean across the diff.

## Test plan

- [x] Full test suite passes (`make test`).
- [x] Lint clean (`make check`).
- [ ] After deploy on a development chat with mixed-case tags: `python scripts/tag-dedup.py --chat-id <dev_id>` shows the dry-run report enumerating case-folding clusters.
- [ ] Same chat with `--apply --no-prompt`: rewrite count matches the proposed cluster size sum, and the audit log is written.
- [ ] Re-run dry-run after apply: zero proposed merges.
- [ ] Spot-check via `/memory stats`: surviving tag set has no case duplicates.
- [ ] `confirmed_action`-tagged row is untouched after `--apply`.
- [ ] Spot-check fact-view detail screen: a rewritten row's other metadata (confidence, prompt_version, etc.) is intact.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,3 +59,6 @@ home/evals/*
 .idea/
 .vscode/
 *.swp
+
+# Tag dedup script audit logs (one per --apply run; never tracked)
+scripts/.tag-dedup-audit-*.jsonl
--- a/scripts/tag-dedup.py
+++ b/scripts/tag-dedup.py
@@ -656,11 +656,15 @@ def main(argv: list[str] | None = None) -> int:
 
     if args.apply:
         print(f"\nTotal rewrites: {total_success} succeeded, {total_failure} failed.")
-        if total_failure > 0:
-            print(f"Audit log: {audit_log_path}", file=sys.stderr)
-            return EXIT_PARTIAL_FAILURE
+        # Audit log is only created when at least one rewrite succeeds
+        # (apply_rewrites flushes lazily). Guard the path message on
+        # total_success > 0 so an all-fail run does not direct the
+        # operator to a file that was never created.
         if total_success > 0:
-            print(f"Audit log: {audit_log_path}")
+            stream = sys.stderr if total_failure > 0 else sys.stdout
+            print(f"Audit log: {audit_log_path}", file=stream)
+        if total_failure > 0:
+            return EXIT_PARTIAL_FAILURE
 
     return EXIT_OK
 
--- a/scripts/tag-dedup.py
+++ b/scripts/tag-dedup.py
@@ -1,0 +1,632 @@
+#!/usr/bin/env python3
+"""
+Periodic tag dedup script (#418, Sub D of #388).
+
+Walks the tag distribution across `source: extracted` and `source: episode`
+rows for one or more users, identifies near-duplicate tag values, and
+rewrites them to a canonical form via `memory.update_metadata`.
+
+Three near-duplicate classes ship in this version:
+  - Case-insensitive equality (`Preference` vs `preference`).
+  - Plural/singular pairs (`-s`, `-es` for sibilant endings, `-ies/-y`).
+  - Within-row deduplication after the per-cluster rewrite collapses
+    two case-variants on a single row.
+
+Default mode is dry-run. The `--apply` flag is required to write. Per-user
+scope by default; `--all-users` opts in to cross-user iteration over
+`config.allowed_user_ids` (the DM-mode assumption that user_id == chat_id
+holds for typical Kai installs; group-chat installs would need a
+chat-id-keyed iteration source).
+
+`confirmed_action` is structurally significant in the extractor pipeline.
+Two layers of defense:
+  - Cluster-skip: any cluster containing `confirmed_action` (case-insensitive)
+    is dropped before any rewrite is proposed.
+  - Row-level guard: if a row's tag list loses `confirmed_action` after the
+    per-cluster rewrite, the row is returned unchanged. The row-level guard
+    is unreachable under the cluster-skip rule but exists as defense-in-depth
+    against a future rule addition that bypasses cluster-skip.
+
+Migration rows (`source: migration`) are excluded from the dedup pass.
+They carry H3-slug tags that are not part of the LLM-tag taxonomy; merging
+them with extracted/episode tags would conflate two vocabulary spaces.
+
+Usage:
+    # Dry run, single user (recommended first step).
+    python scripts/tag-dedup.py --chat-id 12345
+
+    # Apply, single user, with interactive override prompt per cluster.
+    python scripts/tag-dedup.py --chat-id 12345 --apply
+
+    # Apply without prompts (accepts every proposed canonical form).
+    python scripts/tag-dedup.py --chat-id 12345 --apply --no-prompt
+
+    # Dry run across every authorised user.
+    python scripts/tag-dedup.py --all-users
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import sys
+from dataclasses import dataclass
+from datetime import UTC, datetime
+from pathlib import Path
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from kai.memory import MemoryResult
+
+
+# Exit codes (documented in the script header):
+#   0 - success (dry-run completed, or apply completed with all rewrites successful)
+#   1 - unexpected error
+#   2 - memory subsystem disabled (config.memory_enabled is False)
+#   3 - one or more rewrites failed (apply mode only); audit log written for the successful subset
+EXIT_OK = 0
+EXIT_ERROR = 1
+EXIT_MEMORY_DISABLED = 2
+EXIT_PARTIAL_FAILURE = 3
+
+
+# Source filter: dedup acts on `extracted` and `episode` rows only.
+# Migration rows carry H3-slug tags from a different vocabulary space
+# (per the migration script's chunk-naming convention) and merging them
+# with LLM-tag taxonomy values would conflate the two spaces. The
+# `memory.update_metadata` wrapper itself admits all USER_VISIBLE_SOURCES
+# (extracted + episode + migration); the narrower scope is enforced
+# here at the script's load_rows boundary.
+DEDUP_SOURCES: frozenset[str] = frozenset({"extracted", "episode"})
+
+
+# Sibilant endings that take `-es` for the plural rather than `-s`. The
+# plural/singular rule for these endings only fires when both the
+# plural form (ending in one of these patterns) and the singular form
+# (after dropping `-es`) are observed in the corpus.
+_SIBILANT_PLURAL_SUFFIXES: tuple[str, ...] = ("xes", "shes", "ches", "zes", "sses")
+
+
+# The structurally significant tag value that the extractor and validator
+# pipeline keys off (Rules 4 and 4b in `_validate_facts`; the per-source
+# render branch in `_build_fact_view`). The dedup pass refuses to touch
+# any cluster or row that involves this string.
+_CONFIRMED_ACTION_TAG: str = "confirmed_action"
+
+
+# ── Cluster representation ─────────────────────────────────────────────
+
+
+@dataclass
+class Cluster:
+    """One near-duplicate group of tag variants with a chosen canonical.
+
+    Attributes:
+        members: The variant tag values that fall into this cluster.
+            Always at least 2 entries when the cluster is emitted from
+            `identify_clusters` (single-tag groups are not clusters).
+        canonical: The variant chosen as the cluster's target form.
+            Selected by most-frequent wins, lexicographic tiebreak.
+        total_occurrences: Sum of the per-variant counts across members.
+            Defaults to 0 so test fixtures can construct minimal
+            clusters without naming an irrelevant count value;
+            production code paths build clusters via
+            `identify_clusters` which always populates this from the
+            distribution.
+    """
+
+    members: list[str]
+    canonical: str
+    total_occurrences: int = 0
+
+
+# ── Pipeline helpers ──────────────────────────────────────────────────
+
+
+def load_rows(user_id: int) -> list[MemoryResult]:
+    """Fetch all dedup-eligible rows for `user_id` from the memory store.
+
+    Filters client-side to `metadata.source in DEDUP_SOURCES`. The
+    `limit=None` argument to `get_all` ensures a user with thousands of
+    rows still gets the complete distribution. Returns an empty list
+    when the user has no dedup-eligible rows.
+    """
+    from kai import memory
+
+    rows = memory.get_all(user_id=str(user_id), limit=None)
+    return [r for r in rows if r.metadata.get("source") in DEDUP_SOURCES]
+
+
+def compute_distribution(rows: list[MemoryResult]) -> dict[str, int]:
+    """Count case-preserved tag occurrences across every row's tag list.
+
+    A row tagged `["preference", "Preference"]` contributes 1 to each
+    of `"preference"` and `"Preference"` (case is preserved at this
+    stage so the canonical-selection step in `identify_clusters` can
+    pick the most-frequent variant). Empty or missing tag lists
+    contribute nothing.
+    """
+    distribution: dict[str, int] = {}
+    for row in rows:
+        for tag in row.metadata.get("tags") or []:
+            distribution[tag] = distribution.get(tag, 0) + 1
+    return distribution
+
+
+def _try_singular_forms(tag: str) -> list[str]:
+    """Return candidate singular forms for `tag` under the dedup rules.
+
+    Three rules in priority order:
+      - `-ies/-y`: `categories` -> `category`. Applied first so the
+        `-ies` ending does not get reduced to `categori` by the bare
+        `-s` rule below.
+      - Sibilant `-es` removal (`-xes`, `-shes`, `-ches`, `-zes`,
+        `-sses`): `boxes` -> `box`. Applied before the bare `-s` rule
+        so `boxes` does not first reduce to `boxe`.
+      - Bare `-s` removal: `preferences` -> `preference`.
+
+    Returns every candidate that the rules generate; the caller checks
+    against the corpus distribution to keep only those that match an
+    observed tag.
+    """
+    # Empty and one-character tags cannot be plurals under any rule.
+    if len(tag) < 2:
+        return []
+    candidates: list[str] = []
+    if tag.endswith("ies") and len(tag) > 3:
+        candidates.append(tag[:-3] + "y")
+    if any(tag.endswith(suffix) for suffix in _SIBILANT_PLURAL_SUFFIXES):
+        candidates.append(tag[:-2])
+    if tag.endswith("s") and not tag.endswith("ss"):
+        # The `not endswith("ss")` guard prevents stripping the second
+        # `s` from already-singular nouns like `bus` and `mass`. The
+        # bare `-s` rule only fires for genuine plural shapes.
+        candidates.append(tag[:-1])
+    return candidates
+
+
+def _select_canonical(members: list[str], distribution: dict[str, int]) -> str:
+    """Pick the cluster's canonical form from its members.
+
+    Most-frequent wins; ties broken lexicographically. The negative
+    count in the sort key turns "most" into "first" under ascending
+    sort, and the secondary key (the variant string) handles ties
+    deterministically.
+    """
+    # Build (count, variant) pairs, then sort by descending count then
+    # ascending variant. Python's tuple comparison handles the dual key
+    # naturally when the count is negated.
+    return min(members, key=lambda m: (-distribution.get(m, 0), m))
+
+
+def identify_clusters(distribution: dict[str, int]) -> list[Cluster]:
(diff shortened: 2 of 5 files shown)


dcellison:
## Review by Kai

The script isn't deployed yet (PR not merged), so I'll review from the diff directly.

---

## Review

### Warning - Logic Bug: Three-way cluster not collapsed (`scripts/tag-dedup.py`, `identify_clusters`)

The first pass greedily assigns plural case-variants to a case-fold cluster, which blocks the second pass from linking them to the singular.

Example distribution: `{"Preferences": 1, "preferences": 2, "preference": 5}`

1. First pass: `"preferences"` case-fold group has members `["Preferences", "preferences"]` - cluster formed, both added to `assigned`.
2. Second pass: `"preferences"` key is skipped (members already assigned). `"preference"` is not assigned, but `_try_singular_forms("preference")` returns nothing (it ends in `"e"`, not `"s"`). No link is ever made.

Result after dedup: `"Preference"` and `"preferences"` collapse to `"preferences"`, but `"preference"` remains as a separate form. The corpus still has `"preference"` and `"preferences"`.

The expected result is all three collapse to `"preference"` (most frequent, and singular). A fix would require the second pass to also consider the case-folded representative of already-assigned groups as a potential plural target - or a multi-pass approach.

---

### Warning - `metadata: dict` lacks type parameter (`src/kai/memory.py`, `update_metadata` signature)

```python
def update_metadata(*, user_id: str, memory_id: str, data: str, metadata: dict) -> bool:
```

The conventions say "Pyright strict." Bare `dict` fails strict Pyright. Should be `dict[str, Any]` (matching the `MemoryResult.metadata` field type used elsewhere in the file). The rest of the new `memory.py` code is consistent with the existing module style; this is the only type annotation gap.

---

### Suggestion - Spurious singular candidate for `-ze+s` words (`scripts/tag-dedup.py`, `_try_singular_forms`)

`_SIBILANT_PLURAL_SUFFIXES` includes `"zes"`. The check `tag.endswith("zes")` matches words like `"sizes"` (s-i-**z-e-s**) - a genuine `"size" + "s"` plural - not just sibilant plurals like `"buzzes"`. For `"sizes"`, `tag[:-2]` yields `"siz"` instead of `"size"`.

Practically harmless because `"siz"` won't appear in any real corpus and the corpus membership check (`if singular_key not in by_case_fold`) filters it. But it generates a spurious candidate on every call for these words. A more precise guard would check that the character before the suffix is a consonant cluster (`ch`, `sh`, `x`, `zz`, `ss`).

---

### Suggestion - Test assertion guard is always True (`tests/test_tag_dedup.py`, `TestApplyRewrites.test_failed_rewrite_increments_failure_count`)

```python
if audit_log.exists():
    assert audit_log.read_text() == ""
```

`apply_rewrites` opens `audit_log_path` in append mode unconditionally (the `with` block wraps the entire loop, not just the success branch), so the file is always created. The `if audit_log.exists():` guard is dead code. The assertion should be unconditional:

```python
assert audit_log.read_text() == ""
```

This doesn't affect production behavior but makes the test weaker than intended - if the file somehow didn't exist, the test would silently pass rather than fail.

---

### Suggestion - Redundant `int()` cast in `main` (`scripts/tag-dedup.py`, line ~607)

```python
target_ids = [int(args.chat_id)]
```

`--chat-id` is declared `type=int` in argparse, so `args.chat_id` is already an `int`. The `int()` wrap is harmless but unnecessary.

---

### Clean areas

- The `confirmed_action` two-layer defense (cluster-skip + row-level guard) is correctly implemented and well-tested, including the fixture-driven defense-in-depth test for the row-level guard.
- The read-merge-write discipline in `apply_rewrites` is correct, and `TestApplyRewrites.test_successful_rewrite_increments_success_count` pins it well by asserting non-tag metadata fields survive the call.
- The `update_metadata` wrapper correctly mirrors the `delete_by_id` ownership+source gate pattern.
- The `--apply` dry-run default with mandatory opt-in is the right safety default for an admin script.
- Audit log in append mode is correct for multi-user `--all-users` runs.
- `.gitignore` entry matches the default timestamped filename pattern exactly.

The three-way cluster issue is the only finding that could silently underperform in a realistic corpus. Everything else is minor.

dcellison:
## Review by Kai

## Review

All three prior-review warnings were addressed: the three-way cluster collapse now uses union-find correctly, the `metadata: dict` type annotation was fixed to `dict[str, Any]`, the sibilant suffix was narrowed from `"zes"` to `"zzes"` with a clear explanatory comment, the redundant `int()` cast in `main` was removed, and the always-true `if audit_log.exists():` guard was dropped in favor of the unconditional assertion. The `test_three_way_case_and_plural_collapse` test correctly pins the union-find fix.

Two new findings:

---

### Suggestion - Bare `dict` in test fixture helper (`tests/test_tag_dedup.py`, line ~60)

```python
def _row(
    ...
    extra_metadata: dict | None = None,
) -> MemoryResult:
```

Same Pyright-strict violation that was fixed in the production `update_metadata` signature. Should be `dict[str, Any]`. Requires adding `from typing import Any` to the test module imports (the import isn't there yet).

---

### Suggestion - Empty audit file when clusters exist but no rows carry matching tags (`scripts/tag-dedup.py`, `apply_rewrites`)

`apply_rewrites` opens the audit log in append mode unconditionally before the row loop. The `if not clusters: return (0, 0)` guard in `_run_for_user` (around line ~530) prevents file creation for the common no-op case. But if clusters exist yet no row in the corpus actually carries any of their variant tags (e.g., corpus changed between `load_rows` and `apply_rewrites`), the file is created empty without any announcement. The failure-count print path at the end of `main` doesn't fire, and the `if total_success > 0` success-path print also doesn't fire, so the operator has no indication the file was created.

Practical impact is low (empty file in `.gitignore`-d path, rare scenario), but a simple guard inside `apply_rewrites` before the `open` call - or moving the `open` inside the `if ok:` branch - would close the gap cleanly.

---

### Clean areas

- Union-find implementation and path compression are correct. Three-way collapse verified by tracing: `by_case_fold["preferences"] = ["Preferences", "preferences"]`, `union("preferences", "preference")` sets `parent["preferences"] = "preference"`, both keys resolve to root `"preference"`, single component with all three variants. `test_three_way_case_and_plural_collapse` pins this exactly.
- `confirmed_action` two-layer defense is correct and the fixture-driven defense-in-depth test (constructing the bypassing cluster directly) is the right way to pin a row-level guard that's unreachable in production.
- Read-merge-write discipline in `apply_rewrites` and its test coverage in `TestApplyRewrites.test_successful_rewrite_increments_success_count` are solid.
- `test_sparse_metadata_passes_through_unchanged` correctly pins the wrapper's destructive-by-design contract against future "helpful" auto-merge refactors.
- Audit log in append mode is correct for `--all-users` multi-user runs.
- The `--apply` dry-run default with mandatory opt-in is the right safety posture for an admin script.

dcellison:
## Review by Kai

The script isn't in the working directory since the PR isn't merged. I'll review from the diff directly.

---

## Review

All prior-round findings were addressed correctly: union-find three-way collapse, `dict[str, Any]` in production and test fixture, `zzes` sibilant suffix, redundant `int()` cast, unconditional audit-not-exists assertion, and the lazy-flush guard moving `open()` inside `if audit_entries:`.

One new finding:

---

### Warning - Misleading audit-log message when all rewrites fail (`scripts/tag-dedup.py`, `main`, ~line 651)

```python
if total_failure > 0:
    print(f"Audit log: {audit_log_path}", file=sys.stderr)
    return EXIT_PARTIAL_FAILURE
if total_success > 0:
    print(f"Audit log: {audit_log_path}")
```

The `total_failure > 0` branch prints the audit log path unconditionally. But `apply_rewrites` only creates the file when `audit_entries` is non-empty (the lazy-flush fix from round 2). When every row's `update_metadata` call returns False (`total_success == 0`, `total_failure > 0`), the file is never created but the stderr message implies it was. An operator who hits an all-fail run would try to open a nonexistent file.

Fix: guard the stderr print with `total_success > 0`:

```python
if total_failure > 0:
    if total_success > 0:
        print(f"Audit log: {audit_log_path}", file=sys.stderr)
    return EXIT_PARTIAL_FAILURE
```

---

### Clean areas

- Lazy-flush guard (`if audit_entries:` before `open`) correctly closes the empty-file scenario from round 2. `test_failed_rewrite_increments_failure_count` pins it with `assert not audit_log.exists()`.
- Union-find path compression is correct; `test_three_way_case_and_plural_collapse` verifies the three-way collapse.
- `_try_singular_forms` sibilant-suffix candidates for `xes/shes/ches/zzes/sses` all strip the right number of characters (2, giving the stem back).
- `dict[str, Any]` type annotation is consistent across `update_metadata` signature and the `_row` test fixture helper; `from typing import Any` is present in both files.
- `confirmed_action` two-layer defense, read-merge-write discipline, and ownership+source gate in the wrapper are all correct and well-covered by the fixture-driven defense-in-depth test.
- `EXIT_PARTIAL_FAILURE` = 3 is semantically accurate for any mix of failures; the misleading part is the log-path message, not the exit code itself.

The audit-log message is the only new finding; the rest of the PR is clean.

dcellison:
## Review by Kai

## Review

All three prior-round findings were addressed: the misleading audit-log path message when `total_success == 0` is now correctly guarded, and the print is correctly routed to stderr/stdout depending on whether failures were also present.

I traced through the full diff with the prior corrections applied. One new finding:

---

### Suggestion - Redundant `continue` in `prompt_for_override` edit branch (`scripts/tag-dedup.py`, ~line 389)

```python
if choice in ("edit", "e"):
    new_canonical = input("...").strip()
    if new_canonical in cluster.members:
        return new_canonical
    print(f"Invalid: ...")
    # Fall through to re-prompt the y/n/edit/skip choice.
    continue
```

After the `print`, the `continue` is unreachable dead code: the body of the `if choice in ("edit", "e")` block falls through to the next iteration naturally without it (there is no `break` or `return` below). The comment says "Fall through to re-prompt" which is what the loop does on its own. The `continue` is not wrong and does not affect behavior, but the comment is slightly misleading - it implies the `continue` is doing work it isn't doing.

Not worth a revision cycle by itself.

---

### Clean areas

- Round 3 fix is correct: `if total_success > 0:` guards both the stderr and stdout audit-log prints. An all-fail run (`total_success == 0`, `total_failure > 0`) exits `EXIT_PARTIAL_FAILURE` without referencing a nonexistent file.
- Union-find three-way collapse (`Preferences`/`preferences`/`preference`) is correct; traced manually and `test_three_way_case_and_plural_collapse` pins it exactly.
- `zzes` sibilant restriction, `dict[str, Any]` annotations in both `update_metadata` and the `_row` test fixture, and unconditional audit-file assertion are all in place.
- Lazy-flush guard (`if audit_entries:`) correctly prevents empty file creation; `test_failed_rewrite_increments_failure_count` pins it with `assert not audit_log.exists()`.
- Read-merge-write discipline and `test_successful_rewrite_increments_success_count`'s per-field metadata assertions are solid.
- `test_sparse_metadata_passes_through_unchanged` correctly pins the wrapper's destructive-by-design contract against a future auto-merge refactor.
- `confirmed_action` two-layer defense (cluster-skip in `identify_clusters` + row-level guard in `apply_cluster_to_row`) and the fixture-driven defense-in-depth test are correct.
- `--apply` dry-run default with mandatory opt-in is the right safety posture.

The PR is clean.

dcellison:
## Review by Kai

The prior four review rounds addressed every finding raised. I've traced the current diff independently.

**The PR is clean.** No new bugs, security issues, or convention violations found.

Spot-checks confirming correctness:

- **Three-way union-find** (`identify_clusters`): for `{"Preferences": 1, "preferences": 2, "preference": 5}`, `_try_singular_forms("preferences")` yields `["preference"]` (bare-s rule); `union("preferences", "preference")` sets `parent["preferences"] = "preference"`; both case-fold keys resolve to root `"preference"`. Single component with all three variants. ✓
- **`confirmed_action` guard in plural/singular union loop**: `plural_key` and `singular_key` are already case-fold keys (lowercased), so the `== _CONFIRMED_ACTION_TAG` check correctly catches any casing of the original variant. The component-level guard is a correct second layer. ✓
- **`apply_cluster_to_row` with a `confirmed_action`-bearing row**: if `"confirmed_action"` survives the rewrite (it has no cluster mapping), `_CONFIRMED_ACTION_TAG in deduped` is True and the row-level guard passes normally. ✓
- **Lazy-flush in `apply_rewrites`**: `if audit_entries:` before `open` correctly prevents zero-byte file creation on all-fail or no-match runs. ✓
- **Read-merge-write discipline**: `new_metadata = dict(row.metadata)` copies all existing fields before overwriting `tags`, satisfying the wrapper's destructive-by-design contract. ✓
- **Audit-log path message in `main`**: `total_success > 0` guard correctly prevents directing the operator to a file that was never created on an all-fail run. ✓
- **`dict[str, Any]` annotations**: present in `update_metadata` signature (`memory.py`), the `_row` test fixture (`test_tag_dedup.py`), and `apply_rewrites`. ✓

The one cosmetic finding from round 4 (redundant `continue` in the edit branch of `prompt_for_override`, `scripts/tag-dedup.py` ~line 389) was explicitly noted as not worth a revision cycle and remains. That assessment stands.

PR #427, opened 2026-04-30
feat(memory): tighten extractor against workflow-event noise (#426)

### LINKED ISSUE #426: Tighten memory extractor to drop workflow-event noise
## Summary

The Haiku extractor produces too many workflow-event "facts" that have no durable value: spec-evaluation rounds, "User decided to file an issue about X", "PR Y received a review verdict". A 2026-04-30 hygiene sweep deleted 70 such facts (23% of the extracted-source store) and verified via Layer 2 behavioral A/B that the wrong-fact-retrieval failure mode it caused fully resolves with the noise gone (loss count 2 → 1 between pre-sweep and post-sweep on the same probe set, scorable win rate 88.46% → 90.48%). Without an extractor change, the noise will accumulate again at the same 2.5x-per-week rate observed since the last baseline.

## Why the current prompt produces this noise

The STORE list in the v5 extractor prompt includes:

> Decisions the user made in this exchange ("we're going with X", "let's use Y", "I decided Z").

This wording does not distinguish durable design decisions from workflow micro-decisions. The model correctly reads "User decided to file an issue about memory.recall query truncation" as a decision, extracts it, and stores it. The IGNORE list mentions "Transient conversation state" but frames it narrowly ("what you are mid-doing, open questions, clarifying requests"); workflow events do not match that frame.

Compounding factors:

1. No explicit IGNORE rule for spec/PR/issue lifecycle metadata.
2. No "would this be useful 30 days later" gate before emit.
3. The `_validate_facts` Python-side validator does not reject workflow-shaped content; it only enforces the `confirmed_action` quote rule and the consolidation-intent invariants.
4. Confidence is functionally useless as a quality signal: 83% of extracted rows sit at 0.95, none below 0.7, so post-extraction filtering on confidence cannot recover.

## Scope

1. **Prompt v5 → v6** in `src/kai/memory_extraction.py`:
   - Refine STORE / Decisions to scope durability ("architectural or design decisions ... NOT workflow micro-decisions about which task to do next, which spec to evaluate, or which issue to file").
   - Add IGNORE bullets for workflow-event metadata (spec/PR/issue lifecycle events, review-round status) and "decisions to do" workflow actions.
   - Add a "DURABILITY TEST" gate: would this fact still be useful in 30 days? If no, do not emit.
   - Bump `_EXTRACTION_PROMPT_VERSION` 5 → 6 so any future targeted cleanup can identify rows extracted under the new wording.

2. **`_validate_facts` backstop rule** in `src/kai/memory_extraction.py`:
   - Reject facts whose `content` matches a workflow-event regex pattern. Examples to catch: `^User (decided|requested) to (file|create|address|conduct|evaluate)`, `(spec(ification)?\s+\S+|PR\s+#?\d+)\s+(was|were|received).*\b(approved|reviewed|merged|verdict)\b`. Pattern lives next to the existing `confirmed_action` rule with the same shape (rejection, not modification).
   - Counter-incremented in extractor stats so the drop rate can be observed.

3. **Extraction eval harness** at `src/kai/eval/extraction.py` (new), parallel to Layer 1/2/3 modules:
   - Labeled probe fixture `home/evals/extraction-probes.jsonl` (gitignored): each entry is a small conversation window plus the expected extraction outcome (which facts should appear, which should NOT). Probes drawn from real session history (`DATA_DIR/history/2114582497/`) and curated to cover both the workflow-event noise patterns we want to drop and the durable-fact patterns we must preserve.
   - Harness runs the extractor against both v5 and v6 prompts on each probe and reports a confusion matrix: workflow-noise dropped vs preserved, durable facts preserved vs dropped.
   - Reuses subprocess plumbing from `src/kai/eval/behavioral.py`.

4. **Tests**: unit tests for the `_validate_facts` regex; integration tests covering 3-5 representative session windows under both prompts.

## Out of scope

- Confidence-calibration improvements. The squashed distribution is a real problem but unrelated to the workflow-noise filter.
- An anti-fabrication system prompt for the inner Claude. Layer 2 post-sweep showed the hallucination failure mode persists with a clean memory store, so the inner Claude's prompt is the correct fix site, not the extractor. Separate issue.
- Probe-set updating to reflect the post-sweep memory state. 5 probes drifted because their gold facts were correctly identified as low-quality and deleted; a future eval pass would refresh those entries against the new corpus.

## Acceptance

- [ ] `_EXTRACTION_PROMPT_VERSION` bumped to 6; prompt diff matches the spec's wording exactly.
- [ ] `_validate_facts` rejects workflow-event-shaped content; rejection counter exposed in extractor stats.
- [ ] `src/kai/eval/extraction.py` harness runs against `home/evals/extraction-probes.jsonl` and reports the v5-vs-v6 confusion matrix.
- [ ] Probe fixture includes at least 5 workflow-event windows (where v6 should drop the extraction) and 5 durable-fact windows (where v6 must preserve the extraction).
- [ ] On the curated probes, v6 drops ≥80% of workflow-noise extractions while preserving ≥95% of durable-fact extractions. (Numbers tunable in spec; placeholder until baseline measurement.)
- [ ] Unit tests cover the `_validate_facts` regex with positive and negative cases.
- [ ] No regression in existing `test_memory_extraction.py` suite.

## Migration

None. Prompt-version bump means future extracted rows carry `prompt_version: 6` while existing rows remain at `5` (or earlier). A future targeted-cleanup pass could query for `prompt_version != 6 AND <noise pattern>` to retroactively prune; not part of this issue.

## Related

- #362 (memory eval epic). The 2026-04-30 Layer 2 measurement that motivated this issue ran under Sub 3 of that epic. The remaining hallucination failure mode (post-sweep) is not addressed here; it is the iterate-epic candidate noted in the Apr 23 real-prompt rerun comment thread on #362.
- #414 / #415 (free-form tags). v5 of the prompt landed under that issue.

dcellison:
## Summary

The Haiku extractor was producing too many workflow-event "facts" (spec/PR/issue lifecycle events, "User decided to file an issue", review-round verdicts) that have no durable value. A 2026-04-30 hygiene sweep deleted 70 such facts (23% of the extracted-source store) and Layer 2 confirmed the wrong-fact-retrieval failure mode this caused fully resolves with the noise gone (loss count 2 -> 1 between pre-sweep and post-sweep on the same probe set; scorable win rate 88.46% -> 90.48%). Without an extractor change the noise will return at the observed 2.5x-per-week rate; this PR addresses that.

Three coordinated changes:

1. **Prompt v5 -> v6**: scoped the STORE / Decisions bullet to durable architectural decisions, added two IGNORE bullets for workflow-event metadata, added a "DURABILITY TEST" gate (would this fact still be useful in 30 days?), bumped `_EXTRACTION_PROMPT_VERSION` for targeted future cleanup.
2. **`_validate_facts` Rule 6 (Python backstop)**: regex rejecting four shapes of workflow-event content, with a `confirmed_action` skip so canonical confirmation rows like "User confirmed PR #299 was merged on 2026-04-12" pass through unrejected. Per-user rejection counter exposed via `get_extractor_stats()`.
3. **Eval harness `src/kai/eval/extraction.py`**: subprocess-driven head-to-head of v5 vs v6 prompts on a labeled probe fixture. `_PROMPT_V5_PINNED` captures the v5 prompt verbatim so the comparison runs without reverting source. `_run_extractor` gained a keyword-only `system_prompt` parameter (default = active prompt, production callers unchanged).

Closes #426.

## Why three layers

- **Prompt** is the primary gate: the model produces fewer workflow-event candidates in the first place.
- **Regex** is defense-in-depth: catches what slips past the prompt. Logged at INFO so the fire rate is observable; high rate means the prompt is leaking and should be tightened (rather than the regex broadened).
- **Harness** measures the change: the prompt edit is the most load-bearing correctness artifact in the memory pipeline; we want to assert improvement empirically rather than vibe-check it.

## Tests

- **16 new unit tests** in `tests/test_memory_extraction.py`:
  - `TestRule6WorkflowEventRegex`: each of the four regex arms gets positive cases drawn from the real deletion set, plus a shared negative case set covering durable preferences, constraints, and design decisions. End-to-end tests through `_validate_facts` pin the rejection counter delta and the `confirmed_action` skip path (the spec eval's B2 regression).
  - `TestExtractionPromptDurability`: substring pins for the three D1 prompt edits.
  - `TestRunExtractorSystemPromptDefault`: stubs `asyncio.create_subprocess_exec` to capture argv; asserts the default kwarg threads `_EXTRACTION_SYSTEM_PROMPT` byte-for-byte (compatibility contract for callers that omit `system_prompt`).
  - Updated `TestExtractionPromptSoftVocab` for v5 -> v6 version pin and v6 history-fragment pin (existing v5 substring stays valid).
- **14 integration tests** in `tests/test_eval_extraction.py`:
  - `test_v5_pinned_drift`: SHA-256 hash gate so any silent edit to `_PROMPT_V5_PINNED` fails CI.
  - Probe-fixture loader tests + comment/blank-line skipping.
  - `TestClassifyOutcome`: the (category, v5_facts, v6_facts) -> outcome mapping for workflow_dropped / regression / durable_preserved / ambiguous, including the strict `should_extract_any: false` branch.
  - `TestAggregate`: rates skip ambiguous (so uninformative probes do not drag denominators); v5 and v6 Rule 6 rejection counters are reported separately.

Full suite: 2740 passed, 1 skipped. `make check` clean.

## Live extraction harness

This PR ships the harness scaffolding and an example fixture (4 probes); it does NOT ship a live probe fixture (10+ per category). The live fixture is gitignored by design (operator-specific session content). The operator builds it from `DATA_DIR/history/<chat_id>/` when they want to run real measurements; the example fixture exercises the harness wiring end-to-end and is what the integration tests run against.

The harness runs probes sequentially (no `asyncio.gather` / semaphore). For 20 probes at ~10s per subprocess Claude call x 2 prompt arms, expect roughly 7 minutes walltime. Cost estimate at Haiku 4.5 rates (~5K input + ~1K output tokens per call): ~$0.32 per run.

## Out of scope (deliberately)

- **Confidence calibration**: 83% of extracted facts sit at 0.95 today; the field is functionally useless. Real problem, separate fix.
- **Anti-fabrication system prompt for the inner Claude**: Layer 2 post-sweep showed the hallucination failure mode (probe 20) persists with a clean memory store, so the fix site is the inner Claude's prompt, not the extractor. Iterate-epic candidate.
- **Schema-level minimums on `content` text length**: too coarse; the prompt-side handles it.
- **Live probe fixture curation**: gitignored; operator builds it locally.

## Test plan

- [ ] `make check` passes
- [ ] `make test` passes (2740 tests)
- [ ] After deploy: `get_extractor_stats()["rule_6_rejections"]` increments are observable in production logs (look for `_validate_facts: rejecting workflow-event-shaped fact:`)
- [ ] After deploy: spot-check `/memory` to confirm new extracted rows do not include workflow-event content
- [ ] Optional: operator runs the harness end-to-end against a live probe fixture to baseline workflow_drop_rate / durable_preservation_rate

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,6 +53,7 @@ home/config/*
 home/evals/*
 !home/evals/retrieval-baseline.example.json
 !home/evals/probes.example.jsonl
+!home/evals/extraction-probes.example.jsonl
 
 # OS / IDE
 .DS_Store
--- a/home/evals/extraction-probes.example.jsonl
+++ b/home/evals/extraction-probes.example.jsonl
@@ -1,0 +1,18 @@
+# Example probe fixture for the extractor pre/post comparison harness.
+# Tracked in git so integration tests have a stable, reproducible
+# input. The live fixture (extraction-probes.jsonl, gitignored) carries
+# real session content; this file is small enough to read and review.
+#
+# Schema:
+#   probe_id: string id
+#   category: "workflow-noise" | "durable-content"
+#   window: { prior: [{role, text}, ...], current: {user, assistant} }
+#   expected:
+#     workflow-noise: { should_extract_any: false, must_not_contain: [substring, ...] }
+#     durable-content: { should_extract_any: true, must_contain: [substring, ...] }
+#
+# Comments lines start with "#" and are stripped by the loader.
+{"probe_id":"wf-ex-001","category":"workflow-noise","window":{"prior":[],"current":{"user":"Let's evaluate spec 414 v3 next.","assistant":"OK, I'll run the evaluation against the v3 spec and report findings."}},"expected":{"should_extract_any":false,"must_not_contain":["evaluation","spec 414","approved"]}}
+{"probe_id":"wf-ex-002","category":"workflow-noise","window":{"prior":[],"current":{"user":"Please file an issue about the memory.recall query truncation.","assistant":"Filed issue #395 covering the query truncation."}},"expected":{"should_extract_any":false,"must_not_contain":["decided to file","filed an issue"]}}
+{"probe_id":"dur-ex-001","category":"durable-content","window":{"prior":[],"current":{"user":"I prefer Earl Grey over English Breakfast.","assistant":"Got it, Earl Grey noted."}},"expected":{"should_extract_any":true,"must_contain":["earl grey"]}}
+{"probe_id":"dur-ex-002","category":"durable-content","window":{"prior":[],"current":{"user":"The home workspace must be per-user, not shared. Going forward, every user gets their own home dir under DATA_DIR.","assistant":"Understood, per-user home workspaces under DATA_DIR."}},"expected":{"should_extract_any":true,"must_contain":["per-user","home"]}}
--- a/home/evals/extraction-probes.example.jsonl
+++ b/home/evals/extraction-probes.example.jsonl
@@ -11,7 +11,7 @@
 #     workflow-noise: { should_extract_any: false, must_not_contain: [substring, ...] }
 #     durable-content: { should_extract_any: true, must_contain: [substring, ...] }
 #
-# Comments lines start with "#" and are stripped by the loader.
+# Comment lines start with "#" and are stripped by the loader.
 {"probe_id":"wf-ex-001","category":"workflow-noise","window":{"prior":[],"current":{"user":"Let's evaluate spec 414 v3 next.","assistant":"OK, I'll run the evaluation against the v3 spec and report findings."}},"expected":{"should_extract_any":false,"must_not_contain":["evaluation","spec 414","approved"]}}
 {"probe_id":"wf-ex-002","category":"workflow-noise","window":{"prior":[],"current":{"user":"Please file an issue about the memory.recall query truncation.","assistant":"Filed issue #395 covering the query truncation."}},"expected":{"should_extract_any":false,"must_not_contain":["decided to file","filed an issue"]}}
 {"probe_id":"dur-ex-001","category":"durable-content","window":{"prior":[],"current":{"user":"I prefer Earl Grey over English Breakfast.","assistant":"Got it, Earl Grey noted."}},"expected":{"should_extract_any":true,"must_contain":["earl grey"]}}
--- a/home/evals/extraction-probes.example.jsonl
+++ b/home/evals/extraction-probes.example.jsonl
@@ -12,7 +12,7 @@
 #     durable-content: { should_extract_any: true, must_contain: [substring, ...] }
 #
 # Comment lines start with "#" and are stripped by the loader.
-{"probe_id":"wf-ex-001","category":"workflow-noise","window":{"prior":[],"current":{"user":"Let's evaluate spec 414 v3 next.","assistant":"OK, I'll run the evaluation against the v3 spec and report findings."}},"expected":{"should_extract_any":false,"must_not_contain":["evaluation","spec 414","approved"]}}
+{"probe_id":"wf-ex-001","category":"workflow-noise","window":{"prior":[],"current":{"user":"Let's evaluate spec foo v3 next.","assistant":"OK, I'll run the evaluation against the v3 spec and report findings."}},"expected":{"should_extract_any":false,"must_not_contain":["evaluation","spec foo","approved"]}}
 {"probe_id":"wf-ex-002","category":"workflow-noise","window":{"prior":[],"current":{"user":"Please file an issue about the memory.recall query truncation.","assistant":"Filed issue #395 covering the query truncation."}},"expected":{"should_extract_any":false,"must_not_contain":["decided to file","filed an issue"]}}
 {"probe_id":"dur-ex-001","category":"durable-content","window":{"prior":[],"current":{"user":"I prefer Earl Grey over English Breakfast.","assistant":"Got it, Earl Grey noted."}},"expected":{"should_extract_any":true,"must_contain":["earl grey"]}}
 {"probe_id":"dur-ex-002","category":"durable-content","window":{"prior":[],"current":{"user":"The home workspace must be per-user, not shared. Going forward, every user gets their own home dir under DATA_DIR.","assistant":"Understood, per-user home workspaces under DATA_DIR."}},"expected":{"should_extract_any":true,"must_contain":["per-user","home"]}}
--- a/src/kai/eval/extraction.py
+++ b/src/kai/eval/extraction.py
@@ -225,6 +225,13 @@ class Probe:
     expected: dict
 
 
+# Closed set of categories recognised by `_classify_outcome`. Used
+# at load time by `load_probes` to reject typo'd category strings
+# before the harness starts running, since the classifier silently
+# returns "ambiguous" for unknown categories.
+_VALID_CATEGORIES: frozenset[str] = frozenset({"workflow-noise", "durable-content"})
+
+
 @dataclass
 class ProbeOutcome:
     """Per-probe result of running v5 and v6 extractor arms.
@@ -295,6 +302,17 @@ def load_probes(path: Path) -> list[Probe]:
             current = probe.window["current"]
             if not isinstance(current, dict) or not current.get("user") or not current.get("assistant"):
                 raise ValueError(f"{path}:{line_number}: window.current must have non-empty user and assistant")
+            # Validate category against the closed set so a typo
+            # (e.g. underscore-vs-hyphen `workflow_noise`) raises at
+            # load time. The classifier's `if probe.category ==`
+            # comparisons would otherwise silently fall through to
+            # `ambiguous` for every probe in the file, yielding
+            # `None` rates with no indication of why.
+            if probe.category not in _VALID_CATEGORIES:
+                raise ValueError(
+                    f"{path}:{line_number}: unknown category {probe.category!r}"
+                    f" (must be one of {sorted(_VALID_CATEGORIES)})"
+                )
             probes.append(probe)
     return probes
 
--- a/src/kai/eval/extraction.py
+++ b/src/kai/eval/extraction.py
@@ -280,15 +280,21 @@ def load_probes(path: Path) -> list[Probe]:
             except KeyError as exc:
                 raise ValueError(f"{path}:{line_number}: missing required field {exc}") from exc
             # Validate the window structure at load time so a fixture
-            # missing `current` raises an actionable error here instead
-            # of silently routing through `_run_one_probe`'s
-            # error-bucket path at run time. The harness's
-            # `_window_to_extractor_args` tolerates missing `current`
-            # via `.get("current") or {}` for defensive runtime
-            # behavior; the load-time check is the operator's
-            # diagnostic anchor.
+            # missing `current` (or with a typo'd inner field like
+            # `typo_user` instead of `user`) raises an actionable
+            # error here instead of silently routing through
+            # `_run_one_probe`'s error-bucket path at run time with
+            # empty extractor inputs and an unrecoverable subprocess
+            # cost. The harness's `_window_to_extractor_args`
+            # tolerates missing fields via `.get("user", "")` for
+            # defensive runtime behavior; the load-time check is the
+            # operator's diagnostic anchor and catches the typo
+            # class the runtime fallback can't surface.
             if not isinstance(probe.window, dict) or "current" not in probe.window:
                 raise ValueError(f"{path}:{line_number}: window.current is required")
+            current = probe.window["current"]
+            if not isinstance(current, dict) or not current.get("user") or not current.get("assistant"):
+                raise ValueError(f"{path}:{line_number}: window.current must have non-empty user and assistant")
             probes.append(probe)
     return probes
 
@@ -434,6 +440,20 @@ def _classify_outcome(probe: Probe, *, v5_facts: list[str], v6_facts: list[str])
     outcome string. The labels feed the aggregate counters and
     drive the workflow_drop_rate / durable_preservation_rate
     arithmetic.
+
+    Substring-matching semantics:
+
+    - `must_not_contain` is OR: v6 fails if ANY listed substring
+      appears in any v6 fact. The list enumerates banned
+      substrings; one hit is enough to flag a regression.
+    - `must_contain` is also OR: v6 succeeds if ANY listed
+      substring appears in any v6 fact. The list enumerates
+      acceptable anchors and the probe scores `durable_preserved`
+      on the first hit. A future fixture author who needs strict
+      AND semantics ("v6 must mention BOTH home AND per-user")
+      should split into two probes or extend the schema with a
+      `must_contain_all` field; bare `any()` is the documented
+      behavior here.
     """
     must_not_contain = probe.expected.get("must_not_contain") or []
     must_contain = probe.expected.get("must_contain") or []
--- a/src/kai/eval/extraction.py
+++ b/src/kai/eval/extraction.py
@@ -232,6 +232,11 @@ class ProbeOutcome:
     v5_facts: list[str]
     v6_facts: list[str]
     expected_outcome: str
+    # Rule 6 fires per-arm; the harness reports both deltas
+    # separately so the v6 number is not inflated by v5's
+    # higher rejection rate (v5 produces more workflow-event
+    # content by design).
+    v5_rule_6_rejections_delta: int
     v6_rule_6_rejections_delta: int
 
 
@@ -295,13 +300,19 @@ async def _run_one_probe(
         user_text, assistant_text, candidates=None, prior_pairs=prior_pairs
     )
 
-    pre_count = sum(memory_extraction._RULE_6_REJECTIONS.snapshot().values())
-
     # v5 arm: thread the pinned prompt. The `confirmed_action` skip
     # in Rule 6 still applies on this arm since the regex is part of
     # the active validator regardless of which prompt produced the
     # candidate facts; that is the right behavior because Rule 6 is
     # the SAME validator the harness is measuring against.
+    #
+    # Rule 6 fires on BOTH arms (the active validator runs against
+    # whatever facts each arm emits), and v5 is expected to fire it
+    # more because v5 produces more workflow-event content. We
+    # snapshot the counter around each arm separately so the v6
+    # delta is attributed to v6 only - lumping the two arms together
+    # would inflate the v6 metric with v5's more numerous rejections.
+    pre_v5 = sum(memory_extraction._RULE_6_REJECTIONS.snapshot().values())
     v5_result = await memory_extraction._run_extractor(
         payload,
         config,
@@ -310,8 +321,10 @@ async def _run_one_probe(
         user_id=user_id,
         system_prompt=_PROMPT_V5_PINNED,
     )
+    post_v5 = sum(memory_extraction._RULE_6_REJECTIONS.snapshot().values())
 
     # v6 arm: default kwarg, inherits the active `_EXTRACTION_SYSTEM_PROMPT`.
+    pre_v6 = post_v5
     v6_result = await memory_extraction._run_extractor(
         payload,
         config,
@@ -319,8 +332,10 @@ async def _run_one_probe(
         candidate_metadata={},
         user_id=user_id,
     )
+    post_v6 = sum(memory_extraction._RULE_6_REJECTIONS.snapshot().values())
 
-    post_count = sum(memory_extraction._RULE_6_REJECTIONS.snapshot().values())
+    v5_rule_6_delta = post_v5 - pre_v5
+    v6_rule_6_delta = post_v6 - pre_v6
 
     v5_facts = [str(f.get("content", "")) for f in v5_result.facts]
     v6_facts = [str(f.get("content", "")) for f in v6_result.facts]
@@ -333,7 +348,8 @@ async def _run_one_probe(
         v5_facts=v5_facts,
         v6_facts=v6_facts,
         expected_outcome=expected_outcome,
-        v6_rule_6_rejections_delta=post_count - pre_count,
+        v5_rule_6_rejections_delta=v5_rule_6_delta,
+        v6_rule_6_rejections_delta=v6_rule_6_delta,
     )
 
 
@@ -358,8 +374,23 @@ def _classify_outcome(probe: Probe, *, v5_facts: list[str], v6_facts: list[str])
         # we cannot say v6 "dropped" something v5 also did not produce.
         if not v5_facts:
             return "ambiguous"
-        if should_extract_any is False and not v6_facts:
-            return "workflow_dropped"
+        # Strict win condition: when the probe declares
+        # `should_extract_any: false`, only a fully-empty v6 counts
+        # as `workflow_dropped`. Earlier wording let a partial drop
+        # (v6 still emitted facts, just none containing the
+        # forbidden substrings) score as a win, which inflated
+        # workflow_drop_rate on probes where v6 partially-suppresses
+        # but the operator asked for zero extraction.
+        if should_extract_any is False:
+            if not v6_facts:
+                return "workflow_dropped"
+            return "regression"
+        # When the probe does NOT declare should_extract_any: false
+        # (no opinion on whether v6 must be empty), a v6 that
+        # produced fewer facts than v5 AND avoided every forbidden
+        # substring still counts as `workflow_dropped`. This is the
+        # softer win condition for probes where some durable signal
+        # may legitimately remain.
         if not v6_violates_must_not and len(v6_facts) < len(v5_facts):
             return "workflow_dropped"
         return "regression"
@@ -387,9 +418,11 @@ def _aggregate(outcomes: list[ProbeOutcome]) -> dict:
     workflow_drops = 0
     durable_total = 0
     durable_preserves = 0
-    rule_6_total = 0
+    v5_rule_6_total = 0
+    v6_rule_6_total = 0
     for o in outcomes:
-        rule_6_total += o.v6_rule_6_rejections_delta
+        v5_rule_6_total += o.v5_rule_6_rejections_delta
+        v6_rule_6_total += o.v6_rule_6_rejections_delta
         if o.category == "workflow-noise" and o.expected_outcome != "ambiguous":
             workflow_total += 1
             if o.expected_outcome == "workflow_dropped":
@@ -401,7 +434,8 @@ def _aggregate(outcomes: list[ProbeOutcome]) -> dict:
     return {
         "workflow_drop_rate": (workflow_drops / workflow_total) if workflow_total else None,
         "durable_preservation_rate": (durable_preserves / durable_total) if durable_total else None,
-        "rule_6_rejections": rule_6_total,
+        "v5_rule_6_rejections": v5_rule_6_total,
+        "v6_rule_6_rejections": v6_rule_6_total,
         "scorable_workflow_count": workflow_total,
         "scorable_durable_count": durable_total,
     }
@@ -465,10 +499,12 @@ def main(argv: list[str] | None = None) -> int:
     )
     parser.add_argument(
         "--user-id",
-        default="2114582497",
+        default="eval-harness",
         help="user_id passed through to the extractor (cosmetic for "
         "this harness; the harness does not read or write the live "
-        "memory store).",
+        "memory store). The Rule 6 rejection counter is keyed by "
+        "user_id, so setting this to a stable label makes a long-"
+        "running session's rejection counts easy to attribute.",
     )
     parser.add_argument(
         "--output",
--- a/src/kai/eval/extraction.py
+++ b/src/kai/eval/extraction.py
@@ -509,14 +509,19 @@ def _classify_outcome(probe: Probe, *, v5_facts: list[str], v6_facts: list[str])
             return "workflow_dropped"
         return "regression"
     if probe.category == "durable-content":
-        # The win condition: v6 produced at least one fact carrying
-        # the required substrings. Symmetric ambiguity check first:
-        # if BOTH arms came up empty the probe window is too sparse
-        # to inform v5-vs-v6 (parallel to the workflow-noise
-        # branch's `if not v5_facts: return "ambiguous"`). When v5
-        # produced facts but v6 did not, that IS a regression
-        # because the durable fact must come through under v6.
-        if not v5_facts and not v6_facts:
+        # The win condition: v6 PRESERVED a fact v5 also produced.
+        # Symmetric ambiguity checks come first:
+        #   - both arms empty: probe window too sparse to inform.
+        #   - v5 empty, v6 non-empty: v6 found something v5 missed.
+        #     Nothing was "preserved" because v5 had nothing to
+        #     start with; treat as ambiguous rather than counting
+        #     a v6 strict-improvement as preservation. Mirrors the
+        #     workflow-noise branch's `if not v5_facts: return
+        #     "ambiguous"` symmetric guard.
+        # When v5 produced facts but v6 did not, that IS a
+        # regression because the durable fact must come through
+        # under v6.
+        if not v5_facts:
             return "ambiguous"
         if not v6_facts:
             return "regression"
@@ -544,15 +549,15 @@ def _aggregate(outcomes: list[ProbeOutcome]) -> dict:
     # `_run_async` for the per-probe try/except). Both go in the
     # report so the operator sees them, but they do not feed the
     # workflow_drop_rate or durable_preservation_rate arithmetic.
-    _SKIP_OUTCOMES = {"ambiguous", "error"}
+    skip_outcomes = {"ambiguous", "error"}
     for o in outcomes:
         v5_rule_6_total += o.v5_rule_6_rejections_delta
         v6_rule_6_total += o.v6_rule_6_rejections_delta
-        if o.category == "workflow-noise" and o.outcome not in _SKIP_OUTCOMES:
+        if o.category == "workflow-noise" and o.outcome not in skip_outcomes:
             workflow_total += 1
             if o.outcome == "workflow_dropped":
                 workflow_drops += 1
-        elif o.category == "durable-content" and o.outcome not in _SKIP_OUTCOMES:
+        elif o.category == "durable-content" and o.outcome not in skip_outcomes:
             durable_total += 1
             if o.outcome == "durable_preserved":
                 durable_preserves += 1
--- a/src/kai/eval/extraction.py
+++ b/src/kai/eval/extraction.py
@@ -323,21 +323,27 @@ async def _run_one_probe(
     report: an attributable accounting of partial progress, not a
     silent loss of v5's rejections to the error bucket.
     """
-    user_text, assistant_text, prior_pairs = _window_to_extractor_args(probe.window)
-    payload = memory_extraction._build_extraction_payload(
-        user_text, assistant_text, candidates=None, prior_pairs=prior_pairs
-    )
-
     # Initialize all per-arm state to zero/empty so an exception in
-    # either arm leaves the partial-progress accounting accurate.
-    # The control flow inside the try block updates these in place
-    # as each arm completes.
+    # either arm (or the payload-build below) leaves the
+    # partial-progress accounting accurate. The control flow inside
+    # the try block updates these in place as each step completes.
     v5_facts: list[str] = []
     v6_facts: list[str] = []
     v5_rule_6_delta = 0
     v6_rule_6_delta = 0
 
     try:
+        # Payload construction lives inside the try block so a
+        # malformed probe (e.g., a window field that violates an
+        # invariant `_build_extraction_payload` does not check
+        # defensively) routes through the same error-bucket path
+        # as a subprocess crash. Without the guard the docstring's
+        # "Never raises" contract would have a hole - the exact
+        # failure mode the per-probe try/except exists to prevent.
+        user_text, assistant_text, prior_pairs = _window_to_extractor_args(probe.window)
+        payload = memory_extraction._build_extraction_payload(
+            user_text, assistant_text, candidates=None, prior_pairs=prior_pairs
+        )
         # v5 arm: thread the pinned prompt. The `confirmed_action`
         # skip in Rule 6 still applies on this arm since the regex
         # is part of the active validator regardless of which prompt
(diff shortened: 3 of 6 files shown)


dcellison:
## Review by Kai

This is a well-structured PR with solid test coverage. The three-layer approach (prompt, regex, harness) is coherent and the commenting is thorough. A few real issues follow.

---

## Warning

**`v6_rule_6_rejections_delta` counts both arms, not just v6**
`eval/extraction.py`, `_run_one_probe`

`pre_count` is snapshotted before the v5 arm executes. If `_run_extractor` calls `_validate_facts` internally (which increments `_RULE_6_REJECTIONS`), the v5 arm's Rule 6 rejections are included in the delta. The author's own comment makes this explicit: "The `confirmed_action` skip in Rule 6 still applies on this arm [v5] since the regex is part of the active validator." So the field `v6_rule_6_rejections_delta` is mis-named - it is the combined v5+v6 Rule 6 rejection count. Since the v5 prompt produces more workflow-event content by design, the v5 arm will fire Rule 6 more often, inflating the reported v6 metric. Fix: snapshot between arms, not around both.

---

**Probes run sequentially; PR cost estimate assumes `max_concurrency=4`**
`eval/extraction.py`, `_run_async` (~line 435)

```python
for probe in probes:
    outcome = await _run_one_probe(probe, config, user_id=args.user_id)
```

No semaphore or `asyncio.gather` is used. The PR description says "under 10 minutes walltime at `max_concurrency=4`" - that estimate does not match the code. An operator running a live 20-probe fixture will see 4x the expected walltime.

---

## Suggestion

**`workflow-noise` classification overstates the win when v6 still extracts**
`eval/extraction.py`, `_classify_outcome` (~line 278)

The third branch:
```python
if not v6_violates_must_not and len(v6_facts) < len(v5_facts):
    return "workflow_dropped"
```
fires when `should_extract_any is False` but v6 still produced facts (just fewer, and none containing the forbidden substrings). Labeling that `"workflow_dropped"` is optimistic - the probe asked for zero extraction. This will silently inflate `workflow_drop_rate` in fixtures where v6 partially-drops but does not fully suppress.

---

**Hardcoded real-looking Telegram user ID as the CLI default**
`eval/extraction.py`, `main()` (~line 457)

```python
default="2114582497",
```
This looks like a real operator Telegram ID embedded in source. A neutral default like `"eval-harness"` is safer and makes the harness reusable without modification.

---

**`system_prompt` is described as keyword-only but is not enforced**
`memory_extraction.py`, `_run_extractor` signature

The PR description and comment call this a "keyword-only parameter," but there is no `*` separator before it in the signature. Callers can pass it positionally. The description is inaccurate; the behavior is fine, but a stray positional call in a future refactor would silently thread the wrong prompt. Adding `*` before `system_prompt` would match the stated intent.

---

**Arm 2 of `_WORKFLOW_EVENT_RE` has a backtracking-prone sub-pattern**
`memory_extraction.py`, `_WORKFLOW_EVENT_RE`

The combination `(\S+\s+)*?.*\b(approved|...)` can produce quadratic backtracking on long strings that partially satisfy the prefix but miss the verdict word. Input is controlled (short LLM fact strings), so this is unlikely to be exploitable, but anchoring or bounding the match width would eliminate the risk entirely.

---

## Overall

Clean otherwise - test coverage is thorough, the `confirmed_action` skip path is correctly pinned (B2), the drift-hash gate on `_PROMPT_V5_PINNED` is a good idea, and the convention violations (em dashes, missing comments) are not present. The counter-attribution bug in `_run_one_probe` is the most load-bearing issue since it would produce misleading data for the stated purpose of the harness.

dcellison:
## Review by Kai

Looking at the diff relative to what the prior review already raised (those are flagged as addressed or not re-raised per instructions).

---

## New findings

### Warning

**`_run_async` has no per-probe exception handling**
`eval/extraction.py`, `_run_async` (~line 435)

```python
for probe in probes:
    outcome = await _run_one_probe(probe, config, user_id=args.user_id)
    outcomes.append(outcome)
```

If `_run_one_probe` raises (subprocess crash, JSON parse failure, network error), the entire run aborts and `out_path` is never written. For a live 20-probe run this wastes the full cost of completed probes. A simple `try/except` that logs the failure and records an `"error"` outcome, or at minimum writes a partial report in a `finally` block, would make the harness usable under transient failures.

---

### Suggestion

**`ProbeOutcome.expected_outcome` is semantically backwards**
`eval/extraction.py`, `ProbeOutcome` dataclass and `_run_one_probe`

The field stores the *classified* result of how v6 performed against the probe's criteria (`"workflow_dropped"`, `"regression"`, `"durable_preserved"`, `"ambiguous"`). Calling it `expected_outcome` implies it holds what the probe expects, not what v6 actually achieved. The JSON report will contain `"expected_outcome": "workflow_dropped"` which a reader will naturally parse as "we expected a drop" rather than "v6 achieved a drop." `outcome` or `classified_outcome` would be unambiguous. The `_aggregate` loop reads this field in four places, so the rename is contained.

---

**`_window_to_extractor_args` docstring only covers one orphan shape**
`eval/extraction.py`, `_window_to_extractor_args`

The docstring says "an entry is omitted when its role's text is missing" and the test `test_orphan_user_without_following_assistant_dropped` covers the user-orphan case. But the code also silently drops orphan assistant turns (an `"assistant"` entry when `pending_user is None`) without documenting that. The logic is correct; the docstring should say both orphan shapes are dropped so a fixture author knows not to rely on unpaired assistant turns carrying through.

---

## Prior review items: status

| Issue | Status |
|---|---|
| Counter attribution bug (`v6_rule_6_rejections_delta` counted both arms) | Fixed - arms now snapshot independently (`pre_v5`, `post_v5`, `pre_v6 = post_v5`, `post_v6`) |
| Partial-drop classification inflating `workflow_drop_rate` | Fixed - strict `should_extract_any is False` gate lands before the soft path |
| Hardcoded Telegram user ID | Fixed - changed to `"eval-harness"` |
| Arm 2 backtracking-prone pattern | Fixed - `{0,8}?` token bound and `.{0,80}?` char bound added |
| Probes run sequentially / cost estimate mismatch | Not addressed; author has seen it |
| `system_prompt` not keyword-only | Not addressed; author has seen it |

---

## Overall

The PR is clean beyond the two findings above. The per-arm counter attribution fix resolves the most significant correctness issue from the prior round, the regex bounding eliminates the backtracking concern, and the test coverage (Rule 6 positive/negative set, `confirmed_action` skip, drift-hash gate) is thorough. No security issues. No convention violations (no em dashes, comments present throughout).

dcellison:
## Review by Kai

Two prior reviews covered this PR well. I'm checking for issues not yet raised.

---

## Warning

**`_classify_outcome`: both-empty durable probe returns "regression" instead of "ambiguous"**
`eval/extraction.py`, `_classify_outcome`, durable-content branch

```python
if probe.category == "durable-content":
    if not v6_facts:
        return "regression"
```

When `v5_facts=[]` and `v6_facts=[]`, the function returns "regression" because `not v6_facts` is True - even though v5 also failed to extract. The workflow-noise branch handles the symmetric case correctly (`if not v5_facts: return "ambiguous"`); durable-content should too. A probe whose window is too sparse to produce extractions on either arm is uninformative, not a v6 failure. As written, it silently inflates `regression` counts and deflates `durable_preservation_rate`.

`test_durable_v6_empty_is_regression` covers `v5=["User prefers Earl Grey."], v6=[]`, not the `v5=[], v6=[]` case - so the test docstring's claim "even if v5 was also empty" is asserted by the code path, not by the test case.

Fix: add `if not v5_facts and not v6_facts: return "ambiguous"` before the existing `not v6_facts` check.

---

## Suggestion

**Error-path Rule 6 deltas are incorrect when exception occurs between arms**
`eval/extraction.py`, `_run_async` except block

The error `ProbeOutcome` hardcodes `v5_rule_6_rejections_delta=0, v6_rule_6_rejections_delta=0`. If the exception occurs after the v5 arm completed but before v6 (e.g., a subprocess crash in `_run_extractor` on the v6 call), the v5 counter has already been incremented in `_RULE_6_REJECTIONS` but the reported delta is 0. The `_aggregate` v5 total will be under-counted for that run. Low likelihood in practice; capturing the already-computed deltas in a local variable before the v6 call and using them in the except block would avoid this.

---

**Arm 1 `^` anchor does not cover non-leading "User/OC decided to" facts**
`memory_extraction.py`, `_WORKFLOW_EVENT_RE`, arm 1

Arm 1 uses `^` which, under `re.search()`, only fires when the fact content starts at position 0 with "User" or "OC". A paraphrased fact like `"In this exchange, User decided to file an issue about X."` is not caught. The inline comment does note "the model can defeat the regex by paraphrasing" as a documented gap, but the specific shape (subject not at position 0) is not mentioned - worth noting so a future maintainer broadening arm 1 knows to remove the anchor rather than just adding more verbs.

---

## Overall

Clean otherwise. The per-arm counter attribution fix from round 2 is correct, the `confirmed_action` skip path is pinned by tests, the `outcome` rename landed cleanly, and the per-probe exception handling resolves the earlier abort-on-error concern. The three findings above are the only new issues; none are critical.

dcellison:
## Review by Kai

Three prior review rounds resolved the significant issues (counter attribution, partial-drop classification, per-probe exception handling, both-empty durable ambiguity, hardcoded user ID, regex backtracking). I'm checking for new issues only.

---

## Warning

**`_build_extraction_payload` is outside the try/except in `_run_one_probe`**
`src/kai/eval/extraction.py`, lines ~232-245

```python
user_text, assistant_text, prior_pairs = _window_to_extractor_args(probe.window)
payload = memory_extraction._build_extraction_payload(   # <-- outside the guard
    user_text, assistant_text, candidates=None, prior_pairs=prior_pairs
)
...
try:
    pre_v5 = sum(...)
    ...
except Exception as exc:
    classified = "error"
```

The docstring says "Never raises: a subprocess crash, JSON parse failure, or any other exception ... is caught." That contract is not enforced for the payload-build call. `_window_to_extractor_args` is defensive (all `.get()` with defaults), but `_build_extraction_payload` is opaque here - if it raises on a malformed probe, the exception propagates to `_run_async` and aborts the whole run, which is the exact failure mode the per-probe try/except was added to prevent. Fix: move both pre-try calls inside the try block.

---

## Suggestion

**Test hash format differs from report hash format**
`tests/test_eval_extraction.py:29`

```python
_V5_PROMPT_HASH = "764f249d2556a6e00489ac7ba5eac265f4a4d09f27d21dc76f612b84f0874c13"
```

`_hash()` in `extraction.py` emits `"sha256:<hexdigest>"`. The test correctly bypasses `_hash()` and stores the raw hexdigest, but this means the constant in the test file and the `v5_prompt_hash` in the emitted JSON report differ by the `"sha256:"` prefix. An operator cross-checking a run's report against the test constant has to mentally strip the prefix. A comment on `_V5_PROMPT_HASH` noting "raw hexdigest; the JSON report prefixes with `sha256:`" would prevent that confusion.

---

## Overall

Clean otherwise. The warning above is low-likelihood (the defensive helpers make it hard to trigger in practice), but the "Never raises" contract is the stated operational guarantee for the harness, so the gap is worth closing. No security issues, no convention violations (em dash prohibition is clean; the `──` decorators are box-drawing characters, not dashes), and the commenting density is thorough throughout.

dcellison:
## Review by Kai

Four prior rounds covered this PR thoroughly. I'm checking only for issues not yet surfaced.

## New findings

### Suggestion

**`load_probes` swallows no diagnostic context on malformed JSONL**
`src/kai/eval/extraction.py`, `load_probes`

```python
obj = json.loads(line)
```

A malformed line raises `json.JSONDecodeError` and propagates to the caller with no line number or file path in the message. For a fixture an operator curates by hand, the raw Python traceback is a poor diagnostic. A simple wrap:

```python
try:
    obj = json.loads(line)
except json.JSONDecodeError as exc:
    raise ValueError(f"{path}:{line_number}: invalid JSON: {exc}") from exc
```

would make fixture debugging substantially easier. Not a correctness issue, but the harness is operator-facing tooling and fixture errors are the most likely failure mode.

---

**Minor grammatical error in example fixture**
`home/evals/extraction-probes.example.jsonl`, line 14

```
# Comments lines start with "#" and are stripped by the loader.
```

Should be "Comment lines" (drop the plural on "Comments"). This is the only prose documentation for the fixture format.

---

## Previously raised items: status

| Issue | Status |
|---|---|
| Counter attribution (`v6_rule_6_rejections_delta` counting both arms) | Fixed correctly — arms snapshot independently |
| Partial-drop inflation in `workflow_drop_rate` | Fixed — strict `should_extract_any is False` gate |
| Hardcoded Telegram user ID | Fixed — "eval-harness" |
| Regex backtracking risk (arm 2) | Fixed — `{0,8}?` / `.{0,80}?` bounds |
| `_build_extraction_payload` outside `try` block | Fixed — inside `try` block in current diff |
| Both-empty durable probe returning "regression" | Fixed — `not v5_facts and not v6_facts: return "ambiguous"` |
| `_hash` prefix discrepancy between test constant and report | Addressed — comment added to `_V5_PROMPT_HASH` |
| `_run_async` no per-probe exception handling | Fixed — per-probe `try/except`, `"error"` outcome |
| `ProbeOutcome.expected_outcome` misleading name | Fixed — renamed to `outcome` |
| `system_prompt` not keyword-only | Unaddressed; author has seen it |
| Sequential probes / cost estimate mismatch | Unaddressed; author has seen it |

---

## Overall

Clean. The two findings above are minor. The correctness-critical issues from prior rounds are all addressed, the test coverage is thorough (Rule 6 positive/negative cases, `confirmed_action` skip, drift-hash gate, per-arm counter attribution), and there are no convention violations or security issues.

dcellison:
## Review by Kai

Five rounds of prior review resolved the significant issues. Checking only what's new in the current diff.

---

## Warning

**`out_path.write_text()` in `_run_async` is unguarded**
`src/kai/eval/extraction.py`, `_run_async` (end of function)

```python
out_path = Path(args.output)
out_path.write_text(json.dumps(report, indent=2))
print(json.dumps(...))
log.info("wrote %s", out_path)
return 0
```

All probe runs complete before the write. If this raises (`FileNotFoundError` because the parent directory doesn't exist, `PermissionError`, `OSError` from a full disk), the entire run's results are discarded as an uncaught exception rather than a logged error and non-zero exit. At the documented $0.32 / 20-probe run, losing results to an unhandled write failure is a meaningful UX regression. The `probes_path.exists()` guard protects the input but not the output. Wrapping in `try/except OSError` with a fallback to `log.error` and `return 1` would close the gap.

---

## Suggestions

**`config` in `_run_one_probe` is unannotated**
`src/kai/eval/extraction.py`, line ~214

```python
async def _run_one_probe(
    probe: Probe,
    config,          # <-- no annotation
    *,
    user_id: str,
) -> ProbeOutcome:
```

The project convention is Pyright strict. An unannotated parameter is inferred as `Unknown`, which propagates to the `_run_extractor` call. The test passes `config=None`, which suggests the correct annotation is something like `Config | None`. A type alias from `kai.config` would make the contract explicit.

---

**Private names in `__all__`**
`src/kai/eval/extraction.py`, `__all__`

`_OUTPUT_SCHEMA_VERSION` and `_PROMPT_V5_PINNED` appear in `__all__` with leading underscores. Python convention reads leading underscores as private and `__all__` as the public API, so the two signals contradict. Tests import these directly regardless of `__all__`, so this is cosmetic only - but a future `from kai.eval.extraction import *` would export "private" names unexpectedly. Either drop the underscore prefix or remove them from `__all__`.

---

## Overall

Clean otherwise. The per-arm counter attribution fix, both-empty durable ambiguity, per-probe exception handling, payload-build guard inside `try`, regex bounding, `outcome` rename, and fixture diagnostic context are all correctly implemented. Test coverage is thorough and the `_V5_PROMPT_HASH` drift gate is a good long-term correctness anchor. The write-failure gap is the only new issue worth acting on.

dcellison:
## Review by Kai

Six rounds of review have resolved all the significant issues. The current diff is clean.

## Review: Round 6

**No new warnings or critical issues found.**

Previous fixes verified in the current diff:
- Per-arm counter attribution (`pre_v5 / post_v5 / pre_v6 = post_v5 / post_v6`) is correct
- Both-empty durable probe returns `"ambiguous"` (the symmetric fix from round 3)
- `_build_extraction_payload` is inside the `try` block
- `out_path.write_text()` is guarded with `try/except OSError`
- `config: Config` annotation is present on `_run_one_probe`
- `_OUTPUT_SCHEMA_VERSION` and `_PROMPT_V5_PINNED` are no longer in `__all__`
- `load_probes` wraps `JSONDecodeError` with path and line number
- Fixture grammar: "Comment lines" (not "Comments lines") is correct

---

### Suggestion

**`load_probes` does not validate `window.current` exists**
`src/kai/eval/extraction.py`, `load_probes`

A fixture probe with a missing `current` key passes loading but silently routes to `"error"` via `_run_one_probe`'s exception catch, with `_window_to_extractor_args`'s `window.get("current") or {}` producing empty strings. The missing-required-field `KeyError` guard already catches missing `probe_id`/`category`; adding `window` structure validation at the same point would give an equally actionable error message before the harness starts running. Low impact since all four example probes and the test fixtures are well-formed, but operator-curated fixtures are the most likely source of typos.

---

The Rule 6 implementation is correctly scoped, the `confirmed_action` skip path is pinned by tests, the drift-hash gate on `_PROMPT_V5_PINNED` is solid, and the comment density throughout matches the project convention. Clean to merge.

dcellison:
## Review by Kai

Seven prior rounds resolved all significant issues. I've read the diff carefully and found one new finding.

---

## Warning

**`load_probes` validates `window.current` exists but not `window.current.user` / `window.current.assistant`**
`src/kai/eval/extraction.py`, `load_probes` (~line 198)

The round-6 suggestion asked for `window.current` validation and that landed correctly. However, a fixture line like:

```json
{"probe_id":"p1","category":"workflow-noise","window":{"prior":[],"current":{"typo_user":"u","assistant":"a"}},"expected":{}}
```

passes load-time validation (the `current` key exists) and silently routes through `_window_to_extractor_args`'s `current.get("user", "")` fallback, producing empty strings for both extractor inputs. The probe then runs against the live subprocess with an empty user turn, burns cost, and returns a result the operator cannot interpret. The existing field-validation pattern (line ~190, `KeyError` on `probe_id`/`category`/`window`) already has the right shape; adding:

```python
current = probe.window["current"]
if not isinstance(current, dict) or not current.get("user") or not current.get("assistant"):
    raise ValueError(f"{path}:{line_number}: window.current must have non-empty user and assistant")
```

would complete the load-time contract the prior round's fix started.

---

## Suggestion

**`_classify_outcome` `must_contain` uses `any()` — a single-word substring match can produce a false positive**
`src/kai/eval/extraction.py`, `_classify_outcome` (~line 278)

```python
v6_satisfies_must = not must_contain or any(s.lower() in v6_text_lower for s in must_contain)
```

`any()` returns `True` if ANY of the substrings matches. A `must_contain: ["home"]` probe would score `"durable_preserved"` even if v6 emits "User lives in the home office" rather than a per-user home workspace fact. The example fixture's `dur-ex-002` probe uses `["per-user", "home"]` with the same `any()` semantics, which is probably intentional for robustness, but the docstring doesn't document this choice. A comment noting that `must_contain` is OR-semantics (not AND) would prevent a future fixture author from expecting stricter substring requirements.

---

## Prior issue status

All items from rounds 1-6 are correctly addressed in the current diff: per-arm counter attribution, both-empty durable ambiguity, `_build_extraction_payload` inside `try`, `out_path.write_text()` guarded with `OSError`, `config: Config` annotation, private names removed from `__all__`, `load_probes` wraps `JSONDecodeError` with path and line, `window.current` load-time check, fixture grammar, `outcome` rename, per-probe exception handling. The two unaddressed items (`system_prompt` not keyword-only, sequential probes / cost-estimate mismatch) have been seen by the author.

No security issues. No convention violations (the `──` separators are box-drawing characters, not em/en dashes). Comment density is thorough throughout.

dcellison:
## Review by Kai

Seven prior rounds resolved all the significant issues. After reading the full diff, I found two new items.

---

## Suggestion

**`load_probes` doesn't validate `category` against the allowed set**
`src/kai/eval/extraction.py`, `load_probes` (~line 185)

`load_probes` validates that `probe_id`, `category`, and `window` fields exist and that `window.current` has non-empty `user`/`assistant`, but it does not check that `category` is one of `["workflow-noise", "durable-content"]`. A fixture author who writes `"category": "workflow_noise"` (underscore instead of hyphen, a realistic typo) will have every probe from that file silently route to `"ambiguous"` in `_classify_outcome`, yielding `None` rates with no indication of why. This is consistent with the rest of the existing field-validation pattern - adding:

```python
VALID_CATEGORIES = {"workflow-noise", "durable-content"}
if probe.category not in VALID_CATEGORIES:
    raise ValueError(f"{path}:{line_number}: unknown category {probe.category!r}")
```

after the `KeyError` block would close the gap.

---

## Minor style

**Redundant inner imports in `test_default_kwarg_threads_active_prompt`**
`tests/test_eval_extraction.py`, inside `test_default_kwarg_threads_active_prompt`

```python
import asyncio

from kai import memory_extraction
```

Both names are already imported at module scope in the test file (lines 9 and 14 of the new file). The inner imports shadow the module-level names without effect. Not a bug, just noise.

---

## Overall

Clean. The per-arm counter attribution, both-empty durable ambiguity, payload-build inside `try`, `out_path` write guard, `config: Config` annotation, private names removed from `__all__`, `load_probes` diagnostic context, `window.current.user`/`window.current.assistant` validation, `outcome` rename, and `must_contain` OR-semantics documentation are all correctly implemented in the current diff. Test coverage is thorough. No security issues, no convention violations.

dcellison:
## Review by Kai

Eight prior rounds of review resolved all the significant issues. I read the full diff carefully looking for new findings only.

## No new warnings or critical issues found.

All items from prior rounds are correctly addressed in the current diff. Verified:
- Per-arm counter attribution (`pre_v5 / post_v5 / pre_v6 = post_v5 / post_v6`) is correct
- Both-empty durable probe returns `"ambiguous"` before the `not v6_facts` regression check
- `_build_extraction_payload` is inside the `try` block
- `out_path.write_text()` is guarded with `try/except OSError`
- `config: Config` annotation is present on `_run_one_probe`
- Private names (`_OUTPUT_SCHEMA_VERSION`, `_PROMPT_V5_PINNED`) are not in `__all__`
- `load_probes` wraps `JSONDecodeError` with path and line number
- `window.current.user` / `window.current.assistant` validation is correct
- Category validated against `_VALID_CATEGORIES` at load time
- `outcome` rename (not `expected_outcome`) is clean
- `_PROMPT_V5_PINNED` SHA-256 drift gate is present
- `"eval-harness"` default user ID replaces the hardcoded Telegram ID
- Redundant inner imports in `test_default_kwarg_threads_active_prompt` are gone

---

### Suggestion

**`_Counter._counts` accessed directly in test rather than via a reset method**
`tests/test_eval_extraction.py`, `TestRunOneProbeErrorPath.test_v6_exception_preserves_v5_delta`

```python
memory_extraction._RULE_6_REJECTIONS._counts.clear()
```

The comment correctly explains why it doesn't reassign the object (other modules hold a reference to it), but directly accessing the private `_counts` attribute ties the test to the implementation detail. A `_reset()` method on `_Counter` (named with a leading underscore to signal test-only use) would provide the same reset without coupling to the storage shape. Very minor - the current approach works and the reasoning is sound.

---

**Unaddressed from prior rounds (author has seen these):**
- `system_prompt` is not keyword-only (no `*` separator in `_run_extractor` signature)
- Probes run sequentially; the PR description's cost estimate was based on `max_concurrency=4` which the code does not implement

---

Overall: the three-layer design (prompt, regex, harness) is coherent, the test coverage is thorough (Rule 6 positive/negative set, `confirmed_action` skip, drift-hash gate, per-arm counter attribution, per-probe exception handling), and there are no convention violations or security issues. Clean to merge.

dcellison:
## Review by Kai

Nine rounds of prior review caught and resolved every significant issue. I read the full diff carefully looking for new findings only.

## No new warnings or critical issues found.

All prior-round fixes are correctly implemented in the current diff:
- Per-arm counter attribution (`pre_v5 / post_v5 / pre_v6 = post_v5 / post_v6`) is correct
- Both-empty durable probe returns `"ambiguous"` before the `not v6_facts` regression check
- `_build_extraction_payload` is inside the `try` block
- `out_path.write_text()` is guarded with `try/except OSError`
- `config: Config` annotation is present on `_run_one_probe`
- Private names (`_OUTPUT_SCHEMA_VERSION`, `_PROMPT_V5_PINNED`) are not in `__all__`
- `load_probes` wraps `JSONDecodeError` with path and line number
- `window.current.user` / `window.current.assistant` validation is correct
- Category validated against `_VALID_CATEGORIES` at load time
- `outcome` rename is clean throughout
- `_PROMPT_V5_PINNED` SHA-256 drift gate is present
- `"eval-harness"` default user ID is in place
- Redundant inner imports in `test_default_kwarg_threads_active_prompt` are gone
- `_Counter._reset()` method added; test uses it rather than accessing `_counts` directly

---

### Suggestion

**`_run_async` doesn't catch `ValueError` from `load_probes`**
`src/kai/eval/extraction.py`, `_run_async`

```python
probes = load_probes(probes_path)
```

`load_probes` raises `ValueError` with a well-formed message (`{path}:{line}: ...`) when a fixture line is malformed. That exception propagates through `asyncio.run()` in `main()` and surfaces as a Python traceback, not a logged error with a non-zero exit. The `probes_path.exists()` guard above protects the file-not-found case but not the parse-failure case. For a fixture an operator curates by hand, a `try/except ValueError` with `log.error + return 1` would match the error-handling shape used for every other failure path in this function.

---

### Unaddressed from prior rounds (author has seen these)
- `system_prompt` is not keyword-only (no `*` separator in `_run_extractor` signature)
- Probes run sequentially; the PR description's cost estimate was based on `max_concurrency=4`

---

The three-layer design (prompt, regex, harness) is coherent. Test coverage is thorough - Rule 6 positive/negative cases, `confirmed_action` skip path, per-arm counter attribution, both-empty durable ambiguity, per-probe exception handling, and the `_PROMPT_V5_PINNED` drift gate are all correctly pinned. No security issues. No convention violations (the `──` separators are box-drawing characters, not em/en dashes; comment density throughout matches the project convention). Clean to merge.

dcellison:
## Review by Kai

Nine rounds of prior review have been thorough. After reading the full diff, I found two new items and the prior fixes all check out.

---

## Warning

**`_classify_outcome` returns `"durable_preserved"` when `v5_facts=[]` and `v6_facts` is non-empty**
`src/kai/eval/extraction.py`, `_classify_outcome`, durable-content branch (~line 310)

```python
if not v5_facts and not v6_facts:
    return "ambiguous"
if not v6_facts:
    return "regression"
if v6_satisfies_must:
    return "durable_preserved"   # reached when v5=[], v6=non-empty
```

When `v5_facts=[]` and `v6_facts=["User prefers Earl Grey."]`, the function returns `"durable_preserved"`. Nothing was preserved - v6 found something v5 missed entirely. Counting this as preserved inflates `durable_preservation_rate` for the case where v6 strictly improves on v5 for a durable-content probe. The workflow-noise branch handles the symmetric gap correctly (`if not v5_facts: return "ambiguous"`); durable-content should too: add `if not v5_facts: return "ambiguous"` before the existing `not v6_facts` check. No test covers this case - `TestClassifyOutcome` covers `v5=["fact"], v6=[]` (regression) and `v5=[], v6=[]` (ambiguous) but not `v5=[], v6=["fact"]`.

---

## Suggestions

**Misleading comment in `test_rule_6_rejects_workflow_fact_via_validate_facts`**
`tests/test_memory_extraction.py`, `TestRule6WorkflowEventRegex`

```python
# Reset the counter so test order does not affect the delta.
before = sum(_RULE_6_REJECTIONS.snapshot().values())
```

The comment says "Reset the counter" but the code takes a snapshot baseline. The delta approach (`after - before`) is correct and doesn't require a reset. The comment should read "Snapshot baseline" or similar.

---

**`_SKIP_OUTCOMES` inside `_aggregate` uses module-constant naming for a local variable**
`src/kai/eval/extraction.py`, `_aggregate`

```python
_SKIP_OUTCOMES = {"ambiguous", "error"}
```

The all-caps + leading-underscore convention signals a module-level constant; as a local variable the conventional name is `skip_outcomes`. Minor style inconsistency only.

---

## Prior items verified

All fixes from rounds 1-9 are correctly implemented: per-arm counter attribution, both-empty durable ambiguity, `_build_extraction_payload` inside `try`, `out_path.write_text()` guarded with `OSError`, `config: Config` annotation, private names removed from `__all__`, `load_probes` wraps `JSONDecodeError` with path and line, `window.current.user`/`window.current.assistant` validation, `_VALID_CATEGORIES` load-time check, `outcome` rename, `_Counter._reset()` test-only method, `load_probes` ValueError caught in `_run_async`, regex arm 2 backtracking bounds, `"eval-harness"` default user ID.

The two known-unaddressed items (`system_prompt` not keyword-only, sequential probes/cost estimate mismatch) have been seen by the author.

No security issues. No convention violations. The durable-content `v5=[], v6=non-empty` classification is the only new correctness concern; it affects metric accuracy rather than system correctness.

dcellison:
## Review by Kai

Ten rounds of prior review were thorough. The current diff is clean on all the previously-flagged issues. Two new findings follow.

---

## Suggestion

**`test_workflow_noise_v6_violates_must_not` tests the wrong code path**
`tests/test_eval_extraction.py`, `TestClassifyOutcome`

```python
def test_workflow_noise_v6_violates_must_not(self, workflow_probe):
    """v6 still emits a fact carrying a forbidden substring => regression."""
    outcome = extraction._classify_outcome(
        workflow_probe,
        v5_facts=["User decided to file an issue."],
        v6_facts=["User filed an issue about Z."],
    )
    assert outcome == "regression"
```

`workflow_probe` has `should_extract_any: False`. The `is False` branch fires immediately and returns `"regression"` because `v6_facts` is non-empty - the `v6_violates_must_not` check is never reached. The outcome is correct, but the test name and docstring both claim to pin the `must_not_contain` violation path; a future reader studying this test to understand that path would be misled. Fixing the name (e.g., `test_workflow_noise_should_extract_any_false_v6_nonempty_is_regression`) or using a probe without `should_extract_any: False` would make the claim match the code path actually exercised.

---

## Suggestion

**Soft workflow-noise win path has no test coverage**
`src/kai/eval/extraction.py`, `_classify_outcome`, lines roughly 315-320

```python
if not v6_violates_must_not and len(v6_facts) < len(v5_facts):
    return "workflow_dropped"
return "regression"
```

Every `TestClassifyOutcome` workflow-noise case uses `workflow_probe`, which has `should_extract_any: False` and therefore always takes the strict branch above this code. The soft path (no `should_extract_any` declared, v6 drops some but not all facts, no forbidden substrings) is exercised only by `_aggregate`-level semantics, not by a direct `_classify_outcome` assertion. A probe fixture without `should_extract_any` is a legitimate use case (the docstring explicitly documents it). Adding one test case covering `should_extract_any=None`, v5 with two facts, v6 with one non-forbidden fact would complete the matrix.

---

## Prior items verified

All round 1-10 fixes are correctly implemented: per-arm counter attribution, both-empty durable ambiguity (`not v5_facts -> "ambiguous"` before the regression check), `v5=[], v6=non-empty` durable probe returning `"ambiguous"` (with `test_durable_v5_empty_v6_extracts_is_ambiguous` pinning it), `_build_extraction_payload` inside `try`, `out_path.write_text()` guarded with `OSError`, `config: Config` annotation, private names removed from `__all__`, `load_probes` `JSONDecodeError` with path/line context, `load_probes` `ValueError` caught in `_run_async`, `window.current.user`/`window.current.assistant` load-time validation, `_VALID_CATEGORIES` category guard, `outcome` rename, `_Counter._reset()` test-only method, `"eval-harness"` default user ID, regex arm 2 backtracking bounds, `skip_outcomes` (lowercase local, not `_SKIP_OUTCOMES`), and the snapshot-not-reset comment in `test_rule_6_rejects_workflow_fact_via_validate_facts`.

The two known-unaddressed items (`system_prompt` not keyword-only, sequential probes / cost-estimate mismatch in the PR description) have been seen by the author.

No security issues. No convention violations. The three-layer design is coherent and the test coverage is thorough throughout.

dcellison:
## Review by Kai

After ten rounds of prior review that resolved every significant issue, I read the full diff carefully. No new warnings or critical issues found.

## Status of all prior-round items

| Item | Status |
|---|---|
| Counter attribution (`v6_rule_6_rejections_delta` counting both arms) | Correct - per-arm snapshots (`pre_v5 / post_v5 / pre_v6 = post_v5 / post_v6`) |
| Both-empty durable probe returning "regression" | Correct - `if not v5_facts: return "ambiguous"` fires first |
| `v5=[], v6=non-empty` durable probe | Correct - same `if not v5_facts` guard handles it; `test_durable_v5_empty_v6_extracts_is_ambiguous` pins it |
| `_build_extraction_payload` outside `try` | Fixed - inside the try block |
| `out_path.write_text()` unguarded | Fixed - `try/except OSError` with fallback print |
| `config: Config` annotation on `_run_one_probe` | Present |
| Private names in `__all__` | `_OUTPUT_SCHEMA_VERSION` and `_PROMPT_V5_PINNED` absent from `__all__` |
| `load_probes` bare `JSONDecodeError` | Fixed - wraps with `{path}:{line}: invalid JSON:` |
| `load_probes` `ValueError` not caught in `_run_async` | Fixed - `try/except ValueError` with `log.error + return 1` |
| `window.current` / `window.current.user` / `window.current.assistant` validation | Correct |
| `_VALID_CATEGORIES` category guard | Correct |
| `outcome` rename (was `expected_outcome`) | Clean |
| `_Counter._reset()` test-only method | Present; test uses it instead of accessing `_counts` directly |
| Regex arm 2 backtracking bounds | `{0,8}?` token bound + `.{0,80}?` char bound present |
| Redundant inner imports in `test_default_kwarg_threads_active_prompt` | Gone |
| Snapshot-not-reset comment in `test_rule_6_rejects...` | Correct |
| `_SKIP_OUTCOMES` -> `skip_outcomes` (local var naming) | Correct |
| `_V5_PROMPT_HASH` comment noting `sha256:` prefix discrepancy | Present |
| `load_probes` `ValueError` fixture diagnostic | Correct |
| `must_contain` OR-semantics docstring | Documented in `_classify_outcome` docstring |

**Known unaddressed (author has seen both):**
- `system_prompt` is not keyword-only (no `*` separator in `_run_extractor` signature)
- Probes run sequentially; PR description's cost estimate assumed `max_concurrency=4`

## Verified correct

- The regex positive/negative test cases are arithmetically correct against the compiled `_WORKFLOW_EVENT_RE` arms.
- `test_aggregate_rates_skip_ambiguous` counter arithmetic is correct: v5 total = 2+1+1+0+0 = 4, v6 total = 0+0+0+0+1 = 1.
- `TestRunOneProbeErrorPath.test_v6_exception_preserves_v5_delta` correctly exercises the partial-capture path: v5 delta is set before v6 raises, so the `except` block preserves it at 1.
- `_PROMPT_V5_PINNED` SHA-256 drift gate is in place and the hash constant matches what `_hash()` would produce (raw hexdigest, prefixed with `sha256:` in the report).

Clean to merge.

Issue #431, opened 2026-05-01
memory: per-fact provenance for retrieval-time role weighting


dcellison:
## Problem

Stored facts in the memory system don't carry a `source_role` field that records who claimed each fact at extraction time (user-stated, assistant-derived, user-confirmed, inferred). Today, every extracted row is weighted equally at retrieval modulo the existing source-bucket boost (`extracted` vs `migration` vs `episode`). Assistant-derived facts and user-stated facts compete for the same slot budget on equal terms, even though they carry different evidentiary weight.

## Evidence

A 10-probe eval on 2026-04-27 explored whether removing assistant-derived facts from the retrieval bucket would help or hurt. Result:

- Three of four treatment probes (where assistant-derived rows were filtered) lost their best answer. Assistant-derived rows carry **orthogonal coverage** to user-stated rows; removing them blindly creates retrieval gaps.
- The "neutral" heuristic-classified bucket (62 of 214 facts at the time) was mostly agent-derived architectural observations, undercounting agent-derived rows by 4-5x.

Conclusion: post-hoc classification of fact provenance by content alone is unreliable. The required information (who said it) must be recorded at extraction time. The right intervention is provenance plus retrieval-time downweight, not blanket removal.

### Schema and extraction

- Add a required `source_role` field to the fact schema with four enum values: `user_stated`, `user_confirmed`, `assistant_derived`, `inferred`.
- Stage-1 extractor prompt gains a one-paragraph rubric for picking the value:
  - `user_stated`: the user's exact text supports the fact (the canonical "User prefers Earl Grey" shape).
  - `user_confirmed`: the user explicitly affirmed an assistant claim. Already structurally tagged today via `confirmed_action` plus `confirmation_quote`; this value lifts that signal into the new orthogonal axis.
  - `assistant_derived`: the agent or assistant articulated the fact and the user did not push back.
  - `inferred`: neither party stated it directly; the fact emerges from synthesis or reasoning.
- Bump `_EXTRACTION_PROMPT_VERSION` so future cleanups can target rows produced under the new wording.

### Storage

- Persist `source_role` in fact metadata alongside `prompt_version`. Migration entries default to `user_stated` (operator-curated by definition).

### Retrieval-time scoring

- Add a per-role weight, parallel to the existing `_SOURCE_WEIGHTS`. First-cut numbers (illustrative; tune via a Layer 1 baseline rerun):
  - `user_stated`: 1.0 (the load-bearing baseline)
[...]

PR #433, opened 2026-05-01
feat(memory): mode-switch verification harness (#432)

### LINKED ISSUE #432: Phase 5 of #396: mode-switch verification harness
## Problem

The five-phase dual-mode migration (#396) is functionally complete on phases 1, 3, and 4 (Phase 2 rolled into #397's PR). Production has been running on `MEMORY_ENABLED=true` since the migration shipped, but no formal verification of the mode-toggle invariant has been performed end-to-end since #406 / PR #408 landed.

The parent #396 stays open until Phase 5 closes. Three of the parent's acceptance criteria are unverified by any tracked artifact:

- `MEMORY_ENABLED=false` operates identically to current disabled behavior; MEMORY.md is the only fact surface and Qdrant is unreached.
- `MEMORY_ENABLED=true` runs without injecting MEMORY.md; Qdrant is the only fact surface and `/memory` commands work.
- Mode toggle requires only `MEMORY_ENABLED` change plus service restart. No data manipulation needed in either direction.

## Proposed direction

A scripted verification harness that an operator can run to exercise the mode-toggle round trip and assert the parity invariant.

### Shape

A new script under `scripts/` (operator-curated, untracked-by-default like `forget-*.py`) or a tracked `scripts/verify-memory-mode-switch.py` if the harness is reusable across operators. Single argparse entry point; `--apply` triggers the toggle, default is read-only inspection.

### Verification steps

For each mode (`false`, `true`):

1. **Pre-toggle inspection.** Read the running service's effective `MEMORY_ENABLED` value via `/api/memory/stats` (returns 503 when disabled, 200 with stats when enabled) or by inspecting `/etc/kai/env` under sudo.
2. **Toggle.** Edit `/etc/kai/env` to flip the value, then restart the service via `launchctl bootout`/`bootstrap`.
3. **Post-toggle inspection.** Re-query `/api/memory/stats`; assert the new mode is reflected.
4. **Behavioral probe.** Send a known prompt to the bot via the existing `/api/send-message` send-back path, or open a Telegram session and observe directly.
5. **Mode-specific assertions:**
   - `MEMORY_ENABLED=false`: the inner-agent session context must contain the `[Your persistent memory (file: ...)]` block (MEMORY.md injection); must NOT contain the `[Relevant memories from past conversations]` block. `/memory` commands return the documented disabled-mode response.
   - `MEMORY_ENABLED=true`: the inner-agent session context must contain the `[Memory subsystem: enabled]` line; must NOT contain the `[Your persistent memory (file: ...)]` block. `/memory` commands return the full UI.

### Operator-side or scripted

Scripted (Python) is the right shape because:

- The probe sequence has multiple steps that interleave service restarts and assertions. A bash one-liner is fragile.
- The assertion is on inner-agent session context, which the harness can observe by reading the same context the bot constructs (or by intercepting via a test endpoint).
- Repeatable verification is the actual deliverable, not a one-time pass.

A simpler alternative is a documented manual procedure with checkboxes; faster to land, less reusable. Spec writer's choice.

## Acceptance criteria

- A tracked or operator-curated script (or a documented procedure if the spec elects) exercises the mode toggle in both directions.
- Each direction asserts the three load-bearing claims: MEMORY.md inject behavior, Qdrant retrieval behavior, `/memory` command behavior.
- The harness fails loud (non-zero exit, clear error) on any assertion miss.
- A successful end-to-end run is documented (in the PR description or an audit log) before closing #396.
- The five sub-phases of #396 are accounted for: Phase 1 (#397), Phase 2 (rolled into #397's PR), Phase 3 (#403), Phase 4 (#406), Phase 5 (this issue). #396 closes when this issue closes.

## Out of scope

- Changes to the mode-switch implementation itself. Phase 3 (#403 / PR #404) is the implementation; Phase 5 only verifies it.
- Performance or load testing under either mode. Verification is correctness-only.
- Migration of new MEMORY.md content. Phase 4 (#406 / PR #408) is the migration; Phase 5 verifies the post-migration state.
- The verification is a one-time exercise per release, not a CI gate. CI integration is a follow-up if the verification proves load-bearing enough to warrant it.

## Risks

- **Sudo gating.** Service restart requires sudo. The harness either runs under sudo as an operator action, or it accepts a flag that pauses for a manual restart and resumes when the operator confirms.
- **Production disruption.** Toggling `MEMORY_ENABLED=false` in production would temporarily break the agent's access to recent extracted facts. Mitigation: the harness should run on a dev instance, or during a known low-traffic window, or against a snapshot. The spec should specify which.
- **State drift between toggles.** New extractions or new MEMORY.md edits between the two halves of the round trip introduce drift that confuses the assertions. Mitigation: the harness pauses extraction (or runs against a quiesced service) for the duration of the round trip.

dcellison:
Closes #432. Parent epic #396 closes when this lands.

## Summary

Phase 5 of the dual-mode migration. Ships a scripted verification harness for the `MEMORY_ENABLED` toggle's parity invariant: disabled mode means MEMORY.md is the only fact surface and Qdrant is unreached; enabled mode is the reverse.

Two subcommands at `python -m kai.eval.modeswitch`:

1. **`verify`**: black-box logic verification. Imports `build_session_context` and `memory.format_context`, calls each under both flag values against an isolated tmp-dir Qdrant store, asserts nine gating invariants. The mutual-exclusivity partition (the persistent-memory and relevant-memories blocks NEVER coexist) is the load-bearing assertion the existing test suite leaves uncovered.
2. **`check`**: runtime sanity against the running service. Tri-state exit (0 clean, 2 missing `WEBHOOK_SECRET`, 1 health-down or unexpected stats status). Prompt-version probe with `/opt/kai/`-primary plus source-tree-fallback lookup.

Module docstring carries the operator-run round-trip procedure with both macOS (`launchctl bootout` / `bootstrap`) and Linux (`systemctl stop` / `start`) command sets. Round-trip is operator-discretion, not a merge gate.

## Isolation

`verify` redirects `MEM0_DIR` to a tmp dir BEFORE the first `kai.memory` import (which transitively imports mem0; mem0 reads `MEM0_DIR` at module-import time and bakes the value into a module-level constant). Plus monkey-patches `kai.memory.DATA_DIR` for the test Qdrant. Without these, the harness deadlocks against the running production service's lock on `~/.mem0/migrations_qdrant`.

## Tests

20 new tests in `tests/test_eval_modeswitch.py`:

- `TestVerifyInvariants` (8): the nine invariants exercised against `build_session_context` directly plus mocked-recall partition cases.
- `TestCheckSubcommand` (8): each branch of the tri-state exit-code contract via mocked HTTP, including the empty-`WEBHOOK_SECRET` case and both `WEBHOOK_PORT` validation paths (non-integer and out-of-range). Includes an explicit assertion that the auth secret is threaded through to `/api/memory/stats` (defends against a regression that drops the header and silently mis-classifies the resulting 401 as `disabled`), plus a parallel `assert "chat_id" not in url` guard pinning the no-explicit-chat_id contract.
- `TestPromptVersionRead` (4): two-path fallback under tmp_path; decoy-line negative case.

Full suite: **2785 passed, 1 skipped**.

## verify output (merge gate)

```
  PASS  disabled: subsystem marker present
  PASS  disabled: persistent-memory block injected
  PASS  disabled: relevant-memories block omitted
  PASS  enabled: subsystem marker present
  PASS  enabled: persistent-memory block omitted
  PASS  enabled: format_context returns empty or recall-prefixed
  PASS  partition disabled: persistent present, relevant absent
  PASS  partition enabled: persistent absent
  PASS  partition: mutual exclusion across both modes

PASS: all 9 invariants hold.
```

## check output (live production, MEMORY_ENABLED=true)

The `check` subcommand requires `WEBHOOK_SECRET` in the environment, which lives in `/etc/kai/env` (root-owned). Operator-run during merge:

```
$ sudo bash -c 'source /etc/kai/env && python -m kai.eval.modeswitch check'
secret_found: yes
health: ok
mode: enabled
extraction_prompt_version: 7
episode_prompt_version: 1
```

This output will be appended once captured during the merge sequence; the harness itself is what's under review.

## Risks

- **Production Qdrant lock contention.** Verified isolation works locally with the production service running; `verify` does NOT touch `/var/lib/kai/memory/qdrant`.
- **`MEM0_DIR` import-time capture.** Documented at the env-set site; backend.py and config.py do not transitively import mem0, so setting `MEM0_DIR` inside `_run_verify` before the lazy import is correct. A future refactor that imports kai.memory at top-of-file would silently break the redirect; the test suite catches it (the integration probe would deadlock).

## Out of scope

- CI integration (this is operator-run tooling).
- Performance / load testing under either mode.
- Extracted-fact retrieval ranking changes (separate work tracked at #431).

### DIFF
--- a/src/kai/eval/modeswitch.py
+++ b/src/kai/eval/modeswitch.py
@@ -138,10 +138,15 @@ _FIXTURE_CHAT_ID = 1
 _FIXTURE_QUERY = "test fixture for mode-switch verification"
 
 
-# Seeded fact content. Same bland-but-distinctive shape as the
-# MEMORY.md fixture; a substring match between the seeded fact and
-# the format_context output proves the recall path is hitting our
-# isolated store rather than the production one.
+# Seeded fact content. The recall-path invariant the harness asserts
+# is the format_context OUTPUT SHAPE (empty string or starts with
+# the recall-block prefix); content is not asserted programmatically.
+# The real isolation guarantee is `_isolated_data_dir`'s `DATA_DIR`
+# patch + the `MEM0_DIR` env redirect at the top of `_run_verify`.
+# This string is deliberately distinctive so that if a future
+# diagnostic wants to scan the recall output for evidence the
+# isolated store is being hit (rather than the production one), the
+# substring is unique enough to find without false positives.
 _FIXTURE_SEED_FACT = "Mode-switch test fixture marker phrase 7f3a2b1c"
 
 
--- a/src/kai/eval/modeswitch.py
+++ b/src/kai/eval/modeswitch.py
@@ -256,9 +256,14 @@ async def _run_verify() -> int:
     any failure.
 
     Each invariant is computed once and recorded as an
-    `_InvariantResult`. The harness prints results as it goes
-    (rather than batching at the end) so a failure surfaces
-    immediately even if a later invariant raises.
+    `_InvariantResult`. Results are batched in a list and printed
+    together after all invariants run (and after the
+    `MEM0_DIR`/singleton-state restoration in the `finally` block).
+    If an exception is raised mid-run, the operator sees the
+    traceback without partial-progress output; the env-state
+    restoration still fires via `finally`. Operators wanting per-
+    invariant streaming should run individual unit tests in
+    `tests/test_eval_modeswitch.py::TestVerifyInvariants` instead.
     """
     results: list[_InvariantResult] = []
 
--- a/src/kai/eval/modeswitch.py
+++ b/src/kai/eval/modeswitch.py
@@ -1,0 +1,584 @@
+"""Mode-switch verification harness for the dual-mode memory subsystem.
+
+The MEMORY_ENABLED toggle promises that exactly one fact surface is
+active per mode: in disabled mode, MEMORY.md drives behavior and Qdrant
+is unreached; in enabled mode, Qdrant drives behavior and MEMORY.md is
+dormant. This module verifies that promise without requiring a service
+restart by importing the production code paths
+(`build_session_context`, `format_context`) and asserting the gating
+contract directly.
+
+Two subcommands:
+
+    python -m kai.eval.modeswitch verify
+        Black-box logic verification. Imports the production code and
+        asserts every gating invariant under both flag values. Uses an
+        isolated tmp-dir Qdrant store so the live memory directory at
+        $DATA_DIR/memory/qdrant is never touched. Exits 0 on full
+        pass, 1 on any failure.
+
+    python -m kai.eval.modeswitch check
+        Runtime sanity check against the running service. Reports the
+        live effective mode plus the deployed prompt versions. Reads
+        WEBHOOK_SECRET from the process environment; source
+        /etc/kai/env first under sudo:
+
+            sudo bash -c 'source /etc/kai/env && env | grep WEBHOOK_SECRET'
+
+        Exit codes: 0 on a clean read; 2 if WEBHOOK_SECRET is not in
+        the environment; 1 if /health is down or /api/memory/stats
+        returns an unexpected status.
+
+The verify subcommand is the merge gate. The check subcommand is
+informational and can be re-run as production state changes.
+
+Operator round-trip (manual; sudo-gated; not automated by this script):
+
+    1. operator: sudo edit /etc/kai/env, set MEMORY_ENABLED=false
+    2. operator restarts the service:
+       macOS: sudo launchctl bootout system/com.syrinx.kai
+              sudo launchctl bootstrap system /Library/LaunchDaemons/com.syrinx.kai.plist
+       Linux: sudo systemctl stop kai && sudo systemctl start kai
+    3. source /etc/kai/env  # pull WEBHOOK_SECRET into the shell env
+       python -m kai.eval.modeswitch check    # confirms disabled
+    4. python -m kai.eval.modeswitch verify   # logic invariants under both flags
+    5. operator: sudo edit /etc/kai/env, set MEMORY_ENABLED=true
+    6. operator restarts the service (same commands as step 2)
+    7. python -m kai.eval.modeswitch check    # confirms enabled
+
+Step 4's verify is mode-independent: it asserts both flag-value
+behaviors regardless of the running service's flag, so running it
+once anywhere in the procedure is sufficient. Steps 3 and 7 are the
+only checks that observe the live service state.
+"""
+
+from __future__ import annotations
+
+import argparse
+import asyncio
+import logging
+import os
+import re
+import sys
+import tempfile
+import urllib.error
+import urllib.parse
+import urllib.request
+from collections.abc import Iterator
+from contextlib import contextmanager
+from dataclasses import dataclass
+from pathlib import Path
+from unittest.mock import patch
+
+from kai.backend import ApiContext, build_session_context
+from kai.config import Config
+
+log = logging.getLogger(__name__)
+
+
+# ── Verify: invariant assertions ────────────────────────────────────
+
+
+@dataclass
+class _InvariantResult:
+    """Outcome of one invariant check.
+
+    `name` is the human-readable assertion label printed in the
+    report. `passed` is the binary outcome. `detail` carries a
+    short explanation on failure (a substring that was present
+    when it should not have been, or a substring that was missing
+    when it should have been). On pass, `detail` is the empty
+    string.
+    """
+
+    name: str
+    passed: bool
+    detail: str
+
+
+# Substring markers from `build_session_context` (the [Memory subsystem:
+# enabled/disabled] line) and from `memory.format_context` (the
+# [Relevant memories from past conversations ...] header). These are
+# the load-bearing strings the harness asserts on.
+#
+# `_MARKER_PERSISTENT_MEMORY` matches the `[Your persistent memory
+# (file: ...):]` block that build_session_context emits ONLY in
+# disabled mode. The match is a leading-substring check rather than
+# a full-line match so the test does not have to predict the
+# tmp-path part of the format string.
+#
+# `_MARKER_RELEVANT_MEMORIES` matches the recall-block header that
+# format_context emits when search returns non-empty after the
+# floor and budget filters. Same leading-substring approach.
+_MARKER_DISABLED = "[Memory subsystem: disabled]"
+_MARKER_ENABLED = "[Memory subsystem: enabled]"
+_MARKER_PERSISTENT_MEMORY = "[Your persistent memory (file:"
+_MARKER_RELEVANT_MEMORIES = "[Relevant memories from past conversations"
+
+
+# Minimal MEMORY.md fixture used by the verify subcommand. A single
+# truthy line is enough to drive build_session_context's read branch
+# (memory_path.read_text().strip() returns truthy), avoiding the
+# "(currently empty)" / "(not yet created)" placeholder branches.
+# The content is deliberately bland so it cannot be mistaken for
+# operator data if the file ever leaks via test logs.
+_FIXTURE_MEMORY_MD = "Test fixture content for mode-switch verification.\n"
+
+
+# Test fixture chat_id used to scope the temporary MEMORY.md and the
+# Qdrant user_id for the seeded fact. A small positive integer (not
+# a real chat_id) is used to make it visually obvious in any
+# diagnostic output that this is a test value, not production data.
+_FIXTURE_CHAT_ID = 1
+
+
+# Test fixture query string that is guaranteed to share enough
+# semantic surface with the seeded fact to land above the relevance
+# floor. The seeded fact mentions "test fixture" and "mode-switch";
+# the query mentions both, so the embedder produces a high cosine.
+_FIXTURE_QUERY = "test fixture for mode-switch verification"
+
+
+# Seeded fact content. Same bland-but-distinctive shape as the
+# MEMORY.md fixture; a substring match between the seeded fact and
+# the format_context output proves the recall path is hitting our
+# isolated store rather than the production one.
+_FIXTURE_SEED_FACT = "Mode-switch test fixture marker phrase 7f3a2b1c"
+
+
+@contextmanager
+def _isolated_data_dir(tmp_root: Path) -> Iterator[Path]:
+    """Patch `kai.memory.DATA_DIR` to `tmp_root` for the duration of
+    the with-block, reset memory module singleton state on exit.
+
+    Mirrors the existing storage-redirect pattern in
+    `tests/test_memory.py` (the per-test `with patch("kai.memory.DATA_DIR", tmp_path)`
+    block in the integration tests), plus the singleton-state
+    teardown in that file's `_reset_memory_module` helper
+    (`_memory = None; _config = None`) so a subsequent invocation in
+    the same process is not affected by the prior init_memory call.
+
+    init_memory reads `DATA_DIR / "memory" / "qdrant"` from
+    kai.memory at module-resolution time, with no parameter to
+    redirect storage. Patching the module attribute is the
+    supported test pattern.
+    """
+    from kai import memory as memory_module
+
+    with patch.object(memory_module, "DATA_DIR", tmp_root):
+        try:
+            yield tmp_root
+        finally:
+            # Reset singleton state so a follow-up init_memory call
+            # (in the next test, or the next subcommand invocation)
+            # constructs a fresh client against its own DATA_DIR.
+            memory_module._memory = None
+            memory_module._config = None
+
+
+def _build_test_configs(memory_enabled_value: bool) -> Config:
+    """Build a Config with the requested memory_enabled value plus
+    enough surrounding fields to satisfy the production code paths
+    that will read it.
+
+    The verify subcommand never spawns subprocesses or hits the
+    network, so the Telegram and webhook fields are placeholder
+    strings. allowed_user_ids contains the fixture chat_id so the
+    user-id scoping in build_session_context resolves the per-user
+    MEMORY.md path correctly.
+    """
+    return Config(
+        telegram_bot_token="modeswitch-test",
+        allowed_user_ids={_FIXTURE_CHAT_ID},
+        webhook_secret="modeswitch-test",
+        memory_enabled=memory_enabled_value,
+    )
+
+
+def _seed_fixture_memory_md(tmp_root: Path) -> Path:
+    """Create a per-user MEMORY.md fixture under tmp_root. Returns
+    the path so the caller can pass it to build_session_context's
+    data_dir argument.
+
+    Layout matches the production scoping convention:
+    `<data_dir>/memory/<chat_id>/MEMORY.md`. build_session_context
+    will read this file when called with memory_enabled=False.
+    """
+    user_dir = tmp_root / "memory" / str(_FIXTURE_CHAT_ID)
+    user_dir.mkdir(parents=True, exist_ok=True)
+    memory_path = user_dir / "MEMORY.md"
+    memory_path.write_text(_FIXTURE_MEMORY_MD)
+    return memory_path
+
+
+async def _seed_fixture_fact(user_id: str) -> None:
+    """Write one canned fact to the active memory store via
+    `add_structured`. Caller must have already called init_memory
+    inside an `_isolated_data_dir` block so the write lands in
+    the tmp Qdrant.
+
+    Source role tag matches the production extracted-fact shape so
+    the row passes the source-bucket weighting at retrieval time.
+    """
+    from kai import memory as memory_module
+
+    # add_structured is sync (Mem0 is sync). Wrap in
+    # run_in_executor so this coroutine stays cooperative even
+    # though the harness runs it serially.
+    loop = asyncio.get_running_loop()
+    await loop.run_in_executor(
+        None,
+        lambda: memory_module.add_structured(
+            content=_FIXTURE_SEED_FACT,
+            user_id=user_id,
+            memory_type="fact",
+            tags=["fact"],
+            metadata={"source": "extracted", "confidence": 0.9},
+        ),
+    )
+
+
+def _api_ctx_for_verify() -> ApiContext:
+    """Construct an ApiContext for the verify subcommand. The verify
+    path does not exercise any API endpoints; the context is needed
+    only to satisfy build_session_context's signature.
+    """
+    return ApiContext(
+        webhook_port=8080,
+        webhook_secret="modeswitch-test",
+        services_info=[],
+    )
+
+
+async def _run_verify() -> int:
+    """Drive every invariant under both flag values; print PASS/FAIL
+    per invariant and a final summary. Return 0 on full pass, 1 on
+    any failure.
+
+    Each invariant is computed once and recorded as an
+    `_InvariantResult`. The harness prints results as it goes
+    (rather than batching at the end) so a failure surfaces
+    immediately even if a later invariant raises.
+    """
+    results: list[_InvariantResult] = []
+
+    with tempfile.TemporaryDirectory(prefix="modeswitch-") as tmp_str:
+        tmp_root = Path(tmp_str)
+        # Redirect Mem0's home directory so the per-instance
+        # `migrations_qdrant` and `mem0_history.db` it auto-creates
+        # land under the tmp tree rather than the operator's
+        # `~/.mem0/`. The live service holds the lock on the real
+        # `~/.mem0/migrations_qdrant`; without this redirect, the
+        # harness's `init_memory` call would deadlock against it.
+        # Set BEFORE the first import of `kai.memory` (which
+        # transitively imports mem0); mem0 reads `MEM0_DIR` at
+        # module-import time, so setting it later than the first
+        # import has no effect. backend.py and config.py do not
+        # import kai.memory; the first kai.memory import happens
+        # inside `_isolated_data_dir`'s body.
+        os.environ["MEM0_DIR"] = str(tmp_root / "mem0")
+        _seed_fixture_memory_md(tmp_root)
+        api_ctx = _api_ctx_for_verify()
+        # workspace and home_workspace coincide so the optional
+        # identity-injection branch in build_session_context (which
+        # fires only when the two differ) does not pollute the
+        # output and confuse the substring assertions.
+        ws = tmp_root / "workspace"
+        ws.mkdir()
+
+        # ── Disabled-mode invariants ──────────────────────────
+        disabled_ctx = build_session_context(
+            workspace=ws,
+            home_workspace=ws,
+            api=api_ctx,
+            workspace_config=None,
+            chat_id=_FIXTURE_CHAT_ID,
+            data_dir=tmp_root,
+            memory_enabled=False,
+        )
+
+        results.append(
+            _InvariantResult(
+                name="disabled: subsystem marker present",
+                passed=_MARKER_DISABLED in disabled_ctx,
+                detail=f"missing {_MARKER_DISABLED!r}" if _MARKER_DISABLED not in disabled_ctx else "",
+            )
+        )
+        results.append(
+            _InvariantResult(
+                name="disabled: persistent-memory block injected",
+                passed=_MARKER_PERSISTENT_MEMORY in disabled_ctx,
+                detail=f"missing {_MARKER_PERSISTENT_MEMORY!r}"
+                if _MARKER_PERSISTENT_MEMORY not in disabled_ctx
+                else "",
+            )
+        )
+        results.append(
+            _InvariantResult(
+                name="disabled: relevant-memories block omitted",
+                passed=_MARKER_RELEVANT_MEMORIES not in disabled_ctx,
+                detail=f"unexpectedly contained {_MARKER_RELEVANT_MEMORIES!r}"
+                if _MARKER_RELEVANT_MEMORIES in disabled_ctx
+                else "",
+            )
+        )
+
+        # ── Enabled-mode invariants ───────────────────────────
+        enabled_ctx = build_session_context(
+            workspace=ws,
+            home_workspace=ws,
+            api=api_ctx,
+            workspace_config=None,
+            chat_id=_FIXTURE_CHAT_ID,
+            data_dir=tmp_root,
+            memory_enabled=True,
+        )
+
+        results.append(
+            _InvariantResult(
+                name="enabled: subsystem marker present",
+                passed=_MARKER_ENABLED in enabled_ctx,
+                detail=f"missing {_MARKER_ENABLED!r}" if _MARKER_ENABLED not in enabled_ctx else "",
+            )
+        )
+        results.append(
+            _InvariantResult(
+                name="enabled: persistent-memory block omitted",
+                passed=_MARKER_PERSISTENT_MEMORY not in enabled_ctx,
+                detail=f"unexpectedly contained {_MARKER_PERSISTENT_MEMORY!r}"
+                if _MARKER_PERSISTENT_MEMORY in enabled_ctx
+                else "",
+            )
+        )
+
+        # The enabled-mode format_context probe runs in an
+        # isolated tmp-dir Qdrant. We seed one fact, query for it,
+        # and assert the recall block either is empty or starts
+        # with the expected header. Both shapes are valid; the
+        # invariant is the prefix check, not a presence claim.
+        recall_text = ""
+        with _isolated_data_dir(tmp_root):
+            from kai import memory as memory_module
+
+            memory_module.init_memory(_build_test_configs(memory_enabled_value=True))
+            await _seed_fixture_fact(user_id=str(_FIXTURE_CHAT_ID))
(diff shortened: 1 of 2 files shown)


dcellison:
## Review by Kai

## Code Review: `feat(memory): mode-switch verification harness`

Overall the harness is well-structured and the test coverage is solid. Two warnings and two suggestions.

---

### Warning: `MEM0_DIR` env var permanently mutated

**`modeswitch.py`, `_run_verify()`, line ~240**

```python
os.environ["MEM0_DIR"] = str(tmp_root / "mem0")
```

This is set inside the `tempfile.TemporaryDirectory` context but **never restored**. When the `with` block exits the directory is deleted, but `MEM0_DIR` continues to point at the ghost path for the lifetime of the process. Any subsequent code (or test) in the same process that triggers a `kai.memory` import will get a broken path with no diagnostic. The fix is a `try/finally` that saves and restores the original value (or deletes the key if it was absent).

---

### Warning: `build_session_context(memory_enabled=True)` called outside isolation

**`modeswitch.py`, `_run_verify()`, lines ~295-315**

`_isolated_data_dir` patches `kai.memory.DATA_DIR` and resets the singleton, but the `enabled_ctx` call happens before that block opens. The PR description states "backend.py and config.py do not transitively import kai.memory" and the inline comment repeats the claim. That assumption is currently load-bearing and undocumented in code: if `build_session_context` ever acquires a transitive `kai.memory` import (logging, metrics, a new feature), the isolation silently breaks and the harness may contend with the production Qdrant lock. The comment says "the test suite catches it (the integration probe would deadlock)" but a deadlock is not a graceful catch. Moving the `enabled_ctx` call inside its own `_isolated_data_dir` block, or at minimum adding an assertion guard (`assert "kai.memory" not in sys.modules` before the call), would make the invariant explicit.

---

### Suggestion: dead concatenation

**`modeswitch.py`, `_run_verify()`, line ~330**

```python
combined_disabled = disabled_ctx + "\n" + ""
```

The `+ ""` is a no-op. The intent is clear from the comment but the expression could just be `disabled_ctx + "\n"`.

---

### Suggestion: hardcoded port 8080

**`modeswitch.py`, `_run_check()`, lines ~420, ~438**

```python
_http_get("http://localhost:8080/health")
```

The port is hardcoded and does not come from `Config`. If the service ever moves ports (dev vs prod, or a port conflict), `check` silently reports "down" with no clear reason. Reading from `Config` or an env var would be consistent with the project's config pattern.

---

### Clean

- Auth header threading through to `/api/memory/stats` is correctly asserted in the test (`assert secret == "test-secret"`), closing the 401-silent-misclassification regression.
- Tri-state exit code contract is fully covered by `TestCheckSubcommand`.
- `_isolated_data_dir` singleton teardown (`_memory = None; _config = None`) in the `finally` block is correct.
- `_http_get` exception handling (`URLError`, `TimeoutError`, `OSError`) is complete for stdlib urllib.
- The regex decoy test correctly distinguishes comment lines from actual constant declarations.

dcellison:
## Review by Kai

The files are new (not yet merged), so I'm reviewing the diff directly. Here is the review.

---

## Review: `feat(memory): mode-switch verification harness`

The four issues flagged in the prior review are all addressed. Three new findings below.

---

### Warning: `assert` outside `try/finally` leaves `MEM0_DIR` unrestored on failure

**`modeswitch.py`, `_run_verify()`, approximately lines 253-275**

```python
os.environ["MEM0_DIR"] = str(tmp_root / "mem0")   # env mutated here

assert "kai.memory" not in sys.modules, (...)      # <-- outside try

try:                                                # finally only runs if we get here
    ...
finally:
    if prior_mem0_dir is _MEM0_DIR_SENTINEL:
        os.environ.pop("MEM0_DIR", None)
    else:
        os.environ["MEM0_DIR"] = prior_mem0_dir    # never reached on assert failure
```

If `kai.memory` is already in `sys.modules` when `_run_verify` is entered, the `AssertionError` is raised before the `try` block is entered. The `finally` restoration never runs. `MEM0_DIR` is left pointing at the tmp path, which the `TemporaryDirectory` context manager then deletes on exit. Any subsequent mem0 use in the same process silently targets a missing directory with no diagnostic.

The assertion is a diagnostic guard, not a safety gate, so the correct fix is to move it to the first line inside `try:`. The env cleanup then fires unconditionally.

---

### Suggestion: Pyright strict type error on `prior_mem0_dir` assignment

**`modeswitch.py`, `finally` block, approximately line 430**

```python
else:
    os.environ["MEM0_DIR"] = prior_mem0_dir   # prior_mem0_dir: str | object
```

`os.environ.get("MEM0_DIR", _MEM0_DIR_SENTINEL)` returns `str | object` because the sentinel is `object`. Even though the `if prior_mem0_dir is _MEM0_DIR_SENTINEL` branch handles the non-string case, Pyright strict does not narrow `str | object` through an `is` check on a non-literal type. The `__setitem__` of `os.environ` expects `str`, so this is a strict-mode type error under the project's Pyright config. Fix:

```python
else:
    assert isinstance(prior_mem0_dir, str)
    os.environ["MEM0_DIR"] = prior_mem0_dir
```

---

### Suggestion: dead `+ ""` concatenations survive in the test file

**`tests/test_eval_modeswitch.py`, approximately lines 215, 241, 264**

The prior review's dead-concatenation fix landed in `modeswitch.py` but the test file has the same pattern in three places:

```python
combined = disabled_ctx + "\n" + ""        # line ~215
combined = enabled_ctx + "\n" + ""         # line ~241
disabled_combined = _build_disabled_ctx(tmp_path) + "\n" + ""  # line ~264
```

All three should be `... + "\n"`.

---

### Clean

- Auth-header threading assertion (`assert secret == "test-secret"`) in `test_check_stats_503_reports_disabled` closes the 401-silent-misclassification regression correctly.
- `WEBHOOK_PORT` env-var read with integer validation and fallback is a clean resolution of the hardcoded-port suggestion.
- `_MEM0_DIR_SENTINEL` save/restore pattern is correct for the "was absent" vs "was present but empty" distinction.
- `_isolated_data_dir` singleton teardown (`_memory = None; _config = None`) in `finally` is correct.
- Decoy-line regex test in `TestPromptVersionRead` correctly distinguishes comment lines and string-literal escaping from real constant declarations.

dcellison:
## Review by Kai

Both prior rounds were thorough and the author addressed every flagged issue. After a fresh pass I found three minor residual items.

---

## Findings

### Suggestion: `assert` in `finally` can shadow a propagating exception

**`modeswitch.py`, `_run_verify()` finally block**

```python
else:
    assert isinstance(prior_mem0_dir, str)
    os.environ["MEM0_DIR"] = prior_mem0_dir
```

If the try block raises (e.g., an unexpected exception from `build_session_context`) and `prior_mem0_dir` is somehow not a `str`, the `AssertionError` here replaces the original exception. In practice, `os.environ.get()` can only return `str` or the sentinel, so the `isinstance` check is always true and this cannot fire. But the safer pattern for a finally block is `if isinstance(prior_mem0_dir, str)` rather than `assert`, which avoids the exception-replacement hazard entirely and satisfies Pyright's narrowing requirement just as well. Low risk since the type situation is airtight; flagging only because finally blocks warrant extra care.

---

### Suggestion: `_run_check()` accepts out-of-range port silently

**`modeswitch.py`, `_run_check()`, ~line 555**

```python
try:
    port = int(port_str)
except ValueError:
    ...
    port = 8080
```

`int("99999")` succeeds. An out-of-range port (>65535 or ≤0) lands in `_http_get`, fails at the OS level, and returns `(0, b"")`, producing a confusing `health: down (status=0)` output rather than "invalid port." Since this is operator tooling and not user-facing, the impact is low, but a `1 <= port <= 65535` guard with the same fallback-to-8080 behavior would make the diagnostic clearer.

---

### Suggestion: `test_check_health_down_exits_1` leaves two output lines unasserted

**`tests/test_eval_modeswitch.py`, `TestCheckSubcommand.test_check_health_down_exits_1`**

The test verifies `health: down` and `extraction_prompt_version: 7` but does not assert `episode_prompt_version: 1` or `mode: unknown(service-down)`, both of which the production code prints unconditionally on this path. The exit-code contract is fully covered, but a regression that dropped either line would go undetected by this test.

---

## Clean

- `assert "kai.memory" not in sys.modules` correctly fires *before* `MEM0_DIR` is mutated, so an assertion failure leaves the environment untouched and the `finally` restoration is never needed -- the round-2 issue is properly resolved.
- `isinstance(prior_mem0_dir, str)` assertion in `finally` correctly narrows for Pyright strict.
- Dead `+ ""` concatenations are gone from both files.
- `WEBHOOK_PORT` env-var read with integer validation and 8080 fallback is clean.
- Auth-header threading assertion in `test_check_stats_503_reports_disabled` closes the 401-silent-misclassification regression.
- Regex decoy test correctly distinguishes comment lines and escaped string literals from real constant declarations.
- `_isolated_data_dir` singleton teardown (`_memory = None; _config = None`) is unconditional and correct.

dcellison:
## Review by Kai

After three rounds of prior review addressing every flagged issue, the code is in good shape. Two genuinely new minor findings follow.

---

### Suggestion: `assert` guard stripped by `python -O`

**`modeswitch.py`, `_run_verify()`, ~line 260**

```python
assert "kai.memory" not in sys.modules, (
    "kai.memory was imported before _run_verify could set MEM0_DIR ..."
)
```

This guard is the only thing preventing a silent deadlock against the production Qdrant lock if the import graph changes. `assert` is a no-op when Python is invoked with `-O` (or `PYTHONOPTIMIZE=1`). For operator tooling with a safety-critical purpose, a bare `if/raise` is safer:

```python
if "kai.memory" in sys.modules:
    raise RuntimeError(
        "kai.memory was imported before _run_verify could set MEM0_DIR ..."
    )
```

Low probability in practice since an operator is unlikely to use `-O`, but the consequence (silent lock contention with the running service) is high enough to warrant the change.

---

### Suggestion: empty `WEBHOOK_SECRET` exits with misleading message

**`modeswitch.py`, `_run_check()`, ~line 495**

```python
secret = os.environ.get("WEBHOOK_SECRET", "")
secret_found = bool(secret)
```

If `WEBHOOK_SECRET` is exported but set to the empty string, `secret_found` is `False` and the output reads "WEBHOOK_SECRET not set" - which is wrong. The operator did set it; it just happens to be empty. This can happen from a typo in `/etc/kai/env` (`WEBHOOK_SECRET=` with no value). A more precise check:

```python
secret = os.environ.get("WEBHOOK_SECRET")
secret_found = secret is not None and secret != ""
```

And the error message can then distinguish "not exported" from "exported but empty." Again low probability, but `check` is the tool operators run when debugging a config issue - a misleading diagnostic at that moment is frustrating.

---

### Clean

- Round-1 `MEM0_DIR` restoration gap: fixed. The sentinel save/restore with `if isinstance` in `finally` is the right shape for a finally block.
- Round-2 `assert` placement (env mutation before try): fixed. The assertion fires before `MEM0_DIR` is touched, so a failure leaves the environment clean.
- Round-2 `+ ""` dead concatenations: gone from both files.
- Round-3 `assert isinstance` in finally replaced with `elif isinstance`: done.
- Round-3 port range guard (`1 <= port <= 65535`): present and correct.
- Round-3 unasserted output lines in `test_check_health_down_exits_1`: all five lines (`secret_found`, `health`, `mode`, `extraction_prompt_version`, `episode_prompt_version`) are now pinned.
- Auth header threading assertion (`assert secret == "test-secret"`) in `test_check_stats_503_reports_disabled`: correct and load-bearing.
- `_isolated_data_dir` singleton teardown is unconditional and correct.
- Decoy-line regex test correctly rejects both the comment form and the escaped-string-literal form.

dcellison:
## Review by Kai

After four rounds of prior review that thoroughly addressed every flagged issue, the code is in good shape. One new finding.

---

## Findings

### Suggestion: `test_check_stats_200_reports_enabled` leaves two output lines unasserted

**`tests/test_eval_modeswitch.py`, `TestCheckSubcommand.test_check_stats_200_reports_enabled`**

```python
out = capsys.readouterr().out
assert "mode: enabled" in out
assert "extraction_prompt_version: 7" in out
assert "episode_prompt_version: 1" in out
```

`secret_found: yes` and `health: ok` are unasserted. Every other test in `TestCheckSubcommand` pins all five output lines - this was explicitly fixed in round 3 for `test_check_health_down_exits_1` and carried through to `test_check_stats_503_reports_disabled` and `test_check_stats_unexpected_status_exits_1`. The enabled-mode path is the one that's missed. A regression that dropped either of those two lines on the enabled path would go undetected.

Fix: add `assert "secret_found: yes" in out` and `assert "health: ok" in out` to match the pattern established in the other four tests.

---

## Clean

All prior-round issues are resolved:

- `if/raise` replaces `assert` for the `kai.memory` pre-import guard (can no longer be disabled with `-O`)
- `MEM0_DIR` save/restore with `_MEM0_DIR_SENTINEL` is correct; `elif isinstance` in `finally` avoids exception shadowing and satisfies Pyright strict
- Dead `+ ""` concatenations are gone from both files
- `WEBHOOK_PORT` read with integer parse and `1 <= port <= 65535` range guard
- `WEBHOOK_SECRET` empty-vs-`None` distinction with distinct diagnostics
- Auth header threading assertion (`assert secret == "test-secret"`) in `test_check_stats_503_reports_disabled` closes the 401-silent-misclassification regression
- `_isolated_data_dir` singleton teardown (`_memory = None; _config = None`) is unconditional
- Decoy-line regex test correctly rejects comment form and escaped-string-literal form

dcellison:
## Review by Kai

Five rounds of prior review have addressed every issue raised. The code is in good shape. Two genuine new findings:

---

### Suggestion: `_run_verify` docstring contradicts the implementation

**`modeswitch.py`, `_run_verify` docstring, lines ~230-236**

> The harness prints results as it goes (rather than batching at the end) so a failure surfaces immediately even if a later invariant raises.

The implementation does the opposite: it appends to `results: list[_InvariantResult]` throughout and prints all at the end. If an exception is raised mid-run (e.g., inside `_seed_fixture_fact` or `memory_module.format_context`), the `finally` block correctly restores `MEM0_DIR` but zero invariant results are printed - the operator sees only a traceback with no context about which invariants had already passed. The docstring creates a false expectation that partial progress will be visible. Fix: update the docstring to say "batches results and prints all at the end," or actually print each result as it is computed.

---

### Suggestion: auth-header assertion missing from `test_check_stats_200_reports_enabled`

**`tests/test_eval_modeswitch.py`, `TestCheckSubcommand.test_check_stats_200_reports_enabled`, ~line 373**

`test_check_stats_503_reports_disabled` contains `assert secret == "test-secret"` inside the mock to guard against a regression that drops the `X-Webhook-Secret` header and silently mis-classifies the resulting 401. The enabled-mode counterpart's `_http_enabled` mock does not include this assertion. The code path is identical (`_http_get(stats_url, secret=secret)` in both branches), so the disabled-path test provides coverage, but the asymmetry means a future copy of `_http_enabled` that accidentally drops `secret=secret` would go undetected. Adding `assert secret == "test-secret"` to `_http_enabled` closes the gap and makes the pattern consistent across all five `TestCheckSubcommand` tests.

---

### Clean

Everything from prior rounds is resolved: `if/raise` guard before env mutation, `MEM0_DIR` save/restore with `_MEM0_DIR_SENTINEL`, `elif isinstance` in the `finally` block, port range validation, `WEBHOOK_SECRET` empty-vs-None distinction, dead `+ ""` concatenations gone from both files, and all five output lines pinned in every `TestCheckSubcommand` test including the enabled-mode path.

dcellison:
## Review by Kai

Six rounds of prior review have left the code in good shape. One genuine new finding; everything else is clean.

---

### Suggestion: exit-2 test cases leave the third hint line unasserted

**`tests/test_eval_modeswitch.py`**, `test_check_missing_secret_exits_2` (~line 269) and `test_check_empty_secret_exits_2_with_distinct_diagnostic` (~line 290)

The production `_run_check` path prints three lines in each exit-2 case:

```python
print("secret_found: no")
print("  WEBHOOK_SECRET not set; source /etc/kai/env first")
print("  e.g. sudo bash -c 'source /etc/kai/env && env | grep WEBHOOK_SECRET'")
```

Both tests assert only the first two lines. The third (the example command, or the typo-guidance line) is not pinned. Every other `TestCheckSubcommand` test pins all five output lines - a pattern established explicitly in round-3 review and carried through all subsequent rounds. The asymmetry means a regression that drops either third line on an exit-2 path goes undetected.

Fix: add `assert "sudo bash" in out` (missing-secret case) and `assert "a line like" in out` (empty-secret case), or whatever substring is distinctive enough to be non-fragile.

---

### Clean

All prior-round issues are resolved:

- `if/raise` guard fires before `MEM0_DIR` is mutated; cannot be silenced by `-O`.
- `_MEM0_DIR_SENTINEL` save/restore with `elif isinstance` in `finally` is correct and safe (no exception shadowing).
- `enabled_ctx` call outside `_isolated_data_dir` is safe given the documented isolation contract, and the pre-import guard makes the assumption explicit.
- `WEBHOOK_PORT` with integer parse and `1 <= port <= 65535` range guard is correct.
- `WEBHOOK_SECRET` empty-vs-None distinction with distinct diagnostics is correct.
- Auth-header threading assertion (`assert secret == "test-secret"`) is now present symmetrically in all three stats-touching mocks (`_http_disabled`, `_http_enabled`, `_http_unexpected`).
- All five output lines pinned in every `TestCheckSubcommand` test except the two exit-2 cases noted above.
- `_run_verify` docstring now correctly describes batched (not streaming) output.
- Dead `+ ""` concatenations gone from both files.
- `_isolated_data_dir` singleton teardown (`_memory = None; _config = None`) is unconditional and correct.
- Decoy-line regex test correctly rejects comment form and escaped-string-literal form.

dcellison:
## Review by Kai

The diff is in the original message context. I've reviewed it directly. Here is my assessment:

---

## Review: `feat(memory): mode-switch verification harness`

Seven rounds of prior review have been thorough. After a fresh pass, the code is clean.

---

### Clean - all prior-round issues resolved

All six rounds of prior findings are addressed in this diff:

- `if "kai.memory" in sys.modules: raise RuntimeError(...)` guard fires before `MEM0_DIR` is mutated; cannot be silenced with `-O`
- `_MEM0_DIR_SENTINEL` save/restore with `elif isinstance` in `finally` - no exception shadowing, Pyright-clean
- `WEBHOOK_PORT` with `int()` parse and `1 <= port <= 65535` range guard, fallback-to-8080 on bad value
- `WEBHOOK_SECRET` None-vs-empty distinction with distinct diagnostics and distinct exit-2 messages
- Auth-header `assert secret == "test-secret"` present symmetrically in all three stats-touching mocks (`_http_disabled`, `_http_enabled`, `_http_unexpected`)
- All five output lines pinned in every `TestCheckSubcommand` test
- Exit-2 tests now assert all three output lines each: `test_check_missing_secret_exits_2` pins `"sudo bash -c 'source /etc/kai/env"`, `test_check_empty_secret_exits_2_with_distinct_diagnostic` pins `"a line like"`
- `_run_verify` docstring correctly describes batched (not streaming) output
- Dead `+ ""` concatenations absent from both files
- `_isolated_data_dir` singleton teardown (`_memory = None; _config = None`) is unconditional
- Decoy-line regex test rejects comment form and escaped-string-literal form

---

### One new observation (suggestion, low priority)

**`_run_verify` is silently non-reentrant** (`modeswitch.py`, line ~255)

After a successful run, `kai.memory` remains in `sys.modules` (imported inside `_isolated_data_dir`). A second call to `_run_verify()` in the same process hits the `if "kai.memory" in sys.modules: raise RuntimeError(...)` guard immediately. For a CLI tool invoked once per process this is harmless, but any future caller that tries to run `verify` twice in-process (a test wrapper, a retry loop) will get a confusing `RuntimeError` that looks like an import-graph violation rather than a reentrancy limitation. A one-line comment at the guard site ("also fires on a second call to this function; the guard is not reentrant") would prevent a future maintainer from chasing a ghost.

---

### No new bugs, security issues, or convention violations found.

The mutual-exclusivity partition logic, isolation mechanics, tri-state exit-code contract, and prompt-version probe are all correct. This is ready to merge.

dcellison:
## Review by Kai

After a fresh pass across both files, taking the prior eight rounds as settled:

---

## Review: `feat(memory): mode-switch verification harness`

The eight prior rounds were thorough. The code is in very good shape. Two new minor findings.

---

### Suggestion: test module docstring names the wrong shim

**`tests/test_eval_modeswitch.py`, line 14**

```python
# monkeypatches the `urlopen` shim and the `_read_prompt_versions` helper
```

The tests monkeypatch `modeswitch._http_get`, not `urlopen`. `_http_get` is the actual boundary function (it wraps `urlopen`). A future maintainer debugging a test failure who grepped for `urlopen` in the patch calls would find nothing. Fix: `monkeypatches the \`_http_get\` helper and the \`_read_prompt_versions\` helper`.

---

### Suggestion: `except Exception` in `_http_get` catches too broadly

**`modeswitch.py`, `_http_get()`, lines ~470-474**

```python
except urllib.error.HTTPError as exc:
    try:
        body = exc.read()
    except Exception:
        body = b""
```

`HTTPError.read()` raises IO-family exceptions (`http.client.IncompleteRead`, `OSError`). `except Exception` also catches `AttributeError`, `MemoryError`, and `SystemExit` subclasses, which are programming errors worth propagating rather than silently converting to an empty body. `except OSError` is the precise catch here. Low-stakes in practice, but the broad form masks bugs in a code path that's already hard to hit in tests.

---

### Clean

Everything from the prior rounds is resolved and correct:

- `if/raise` guard fires before `MEM0_DIR` is touched; not suppressible with `-O`
- Reentrancy comment is present at the guard site
- `_MEM0_DIR_SENTINEL` save/restore with `elif isinstance` in `finally` - no exception shadowing, Pyright-clean
- `WEBHOOK_PORT` with int parse and `1 <= port <= 65535` range guard, 8080 fallback
- `WEBHOOK_SECRET` None-vs-empty distinction with distinct diagnostics and distinct exit-2 messages
- Auth-header assertion (`assert secret == "test-secret"`) symmetric across all three stats-touching mocks
- All output lines pinned in all six `TestCheckSubcommand` tests (both exit-2 paths pin all three lines each)
- `_run_verify` docstring correctly describes batched output
- Dead `+ ""` concatenations absent from both files
- `_isolated_data_dir` singleton teardown is unconditional
- Decoy-line regex test rejects both the comment form and the escaped-string-literal form

No bugs, security issues, or blocking convention violations.

dcellison:
## Review by Kai

After nine rounds of thorough prior review, the code is in good shape. Two new minor findings:

---

### Suggestion: `_FIXTURE_SEED_FACT` comment overstates the isolation proof

**`modeswitch.py`, ~line 140**

```python
# Seeded fact content. ... a substring match between the seeded fact and
# the format_context output proves the recall path is hitting our
# isolated store rather than the production one.
_FIXTURE_SEED_FACT = "Mode-switch test fixture marker phrase 7f3a2b1c"
```

No such substring match exists. The actual invariant at ~line 384 is:

```python
recall_ok = recall_text == "" or recall_text.startswith(_MARKER_RELEVANT_MEMORIES)
```

That checks output format, not content. The real isolation guarantee is `_isolated_data_dir`'s `DATA_DIR` patch, not content inspection. The comment should say something like "content is distinctive enough to appear in diagnostic output if the isolated store is contaminated" rather than claiming a programmatic assertion that is not there.

---

### Suggestion: WEBHOOK_PORT validation paths are untested

**`modeswitch.py`, ~lines 555-566; `tests/test_eval_modeswitch.py`, `TestCheckSubcommand`**

`_run_check` has two non-trivial validation branches added in round 3: the `ValueError` fallback and the `1 <= port <= 65535` range check. No `TestCheckSubcommand` test drives either path. A regression that dropped the range check or changed the fallback value from 8080 would go undetected. The test would only need to set `WEBHOOK_PORT=invalid` (or `WEBHOOK_PORT=99999`) and assert the fallback behavior.

---

### Trivial observation: PR description test count is off

The description says "TestCheckSubcommand (5)" and "17 new tests." The diff has 6 tests in that class (the empty-secret test was added in round 4 and the description was not updated), for a total of 18. No behavioral impact - just stale documentation.

---

### Clean

All nine prior-round issues are resolved and correct. The code is ready to merge.

dcellison:
## Review by Kai

Ten rounds of prior review have been thorough. After a fresh pass, two genuine new findings in the tests added to address round-10's port-validation coverage suggestion.

---

### Suggestion: auth-header guard missing from both new port-validation mocks

**`tests/test_eval_modeswitch.py` -- `_http_with_capture` in `test_check_webhook_port_non_integer_falls_back_to_8080` (~line 471) and `test_check_webhook_port_out_of_range_falls_back_to_8080` (~line 507)**

```python
def _http_with_capture(url, secret=None, timeout=5.0):
    captured_urls.append(url)
    if "/health" in url:
        return 200, b'{"status": "ok"}'
    if "/api/memory/stats" in url:
        return 200, b'{"total_count": 0}'   # no secret assertion
    raise AssertionError(f"unexpected url: {url}")
```

All three other stats-touching mocks (`_http_disabled`, `_http_enabled`, `_http_unexpected`) include `assert secret == "test-secret"` with an explicit comment explaining its purpose: a regression that drops `secret=secret` from the `_http_get(stats_url, secret=secret)` call would produce a 401 that would mis-classify the mode, and these per-test guards are the only thing that catches it. The two new port-fallback tests drive the same call site but the asymmetric mock would let that regression go undetected.

---

### Suggestion: both new port-validation tests omit the five-line output shape

**Same two tests**

The pattern established in round 3 and carried through all six prior `TestCheckSubcommand` tests is to pin all five output lines. Both new tests assert only the warning line and the URL port; `secret_found: yes`, `health: ok`, `mode: enabled`, `extraction_prompt_version: 7`, and `episode_prompt_version: 1` are unasserted. A regression that dropped any of those on the port-fallback path would go undetected by these tests.

---

### Clean

Everything from the ten prior rounds is resolved correctly. The `if/raise` guard, `_MEM0_DIR_SENTINEL` save/restore, `elif isinstance` in `finally`, port range validation, `WEBHOOK_SECRET` None-vs-empty distinction, auth-header symmetry across the three original stats mocks, full output pinning for the six original `TestCheckSubcommand` tests, the `_run_verify` docstring, the `_FIXTURE_SEED_FACT` comment, and the decoy-line regex test are all in the correct shape. No bugs, security issues, or blocking convention violations.

dcellison:
## AC7 close-out: full round-trip captured

Post-merge install + operator-run round-trip per the harness module docstring. All four phases captured against the freshly deployed `7960135`-era binary; #396's three load-bearing acceptance criteria empirically demonstrated.

### Phase 1: baseline (current `MEMORY_ENABLED=true`)

```
$ sudo bash -c 'set -a && source /etc/kai/env && python -m kai.eval.modeswitch check'
secret_found: yes
health: ok
mode: enabled
extraction_prompt_version: 7
episode_prompt_version: 1
```

### Phase 2: toggle to disabled

```
$ sudo vi /etc/kai/env       # MEMORY_ENABLED=false
$ sudo launchctl bootout system/com.syrinx.kai
$ sudo launchctl bootstrap system /Library/LaunchDaemons/com.syrinx.kai.plist
$ sudo bash -c 'set -a && source /etc/kai/env && python -m kai.eval.modeswitch check'
secret_found: yes
health: ok
mode: disabled
extraction_prompt_version: 7
episode_prompt_version: 1
```

### Phase 3: verify under disabled-mode running service (mode-independence demonstrated)

```
$ python -m kai.eval.modeswitch verify
  PASS  disabled: subsystem marker present
  PASS  disabled: persistent-memory block injected
  PASS  disabled: relevant-memories block omitted
  PASS  enabled: subsystem marker present
  PASS  enabled: persistent-memory block omitted
  PASS  enabled: format_context returns empty or recall-prefixed
  PASS  partition disabled: persistent present, relevant absent
  PASS  partition enabled: persistent absent
  PASS  partition: mutual exclusion across both modes

PASS: all 9 invariants hold.
```

### Phase 4: toggle back to enabled

```
$ sudo vi /etc/kai/env       # MEMORY_ENABLED=true
$ sudo launchctl bootout system/com.syrinx.kai
$ sudo launchctl bootstrap system /Library/LaunchDaemons/com.syrinx.kai.plist
$ sudo bash -c 'set -a && source /etc/kai/env && python -m kai.eval.modeswitch check'
secret_found: yes
health: ok
mode: enabled
extraction_prompt_version: 7
episode_prompt_version: 1
```

### Close-out for parent #396

Three of #396's acceptance criteria were unverified by any tracked artifact pre-merge; all three are now demonstrated:

- `MEMORY_ENABLED=false` operates identically to disabled behavior (Phase 2 mode signal + verify partition invariants under disabled flag).
- `MEMORY_ENABLED=true` runs without injecting MEMORY.md (Phase 1 / Phase 4 mode signal + verify partition invariants under enabled flag).
- Mode toggle requires only `MEMORY_ENABLED` change plus service restart, no data manipulation (each phase transition was just env-edit + `launchctl bootout`/`bootstrap`; no migration step ran in either direction).

Parent #396 can close on this comment.

### Small follow-up: docstring `set -a` correction

The harness docstring at `src/kai/eval/modeswitch.py` step 3 documents `source /etc/kai/env` without the `set -a` auto-export prefix. Without it, the var is set in the bash process but not inherited by the Python child, and `check` reports `secret_found: no`. The procedure that actually works on this deploy is `set -a && source /etc/kai/env`. Filing a tiny doc-only follow-up to correct the docstring.

PR #435, opened 2026-05-01
feat(memory): write-side mode-switch verification (#434)

### LINKED ISSUE #434: Mode-switch verification: write-side coverage
## Summary

The mode-switch verification harness shipped in #432/#433 (`python -m kai.eval.modeswitch`) covers two of the six switch points enumerated in #396; both on the read side (the persistent-memory inject block in session context; the relevant-memories retrieval block under partition invariants). The four write-side switch points are unverified: a regression that broke "writes go to MEMORY.md when disabled" or "writes go to Qdrant when enabled" would not be caught by either `verify` or `check`.

This issue extends the harness to cover those four switch points so the dual-mode contract is regression-proof on both sides of the read/write boundary.

## Coverage gap

| # | Switch point | Site | Currently verified? |
|---|---|---|---|
| 1 | Persistent-memory inject block | `build_session_context` in `backend.py` | Yes (read) |
| 2 | Inner-Claude write routing: Edit MEMORY.md vs `/api/memory/add` | inner-Claude prompt context | No |
| 3 | Relevant-memories retrieval block | `format_context` in `memory.py` | Yes (read) |
| 4 | `/memory` slash commands disabled-mode reply | `handle_memory_command`, `handle_memory_callback` in `memory_command.py` | No |
| 5 | Track 2 extraction gating: `extract_and_store` no-ops when disabled | post-response ingestion guard in `bot.py` | No |
| 6 | Memory retrieval/extraction structured logging | `memory.py`, `memory_extraction.py` | No |

## Acceptance criteria

Each new test must fail if the corresponding gated branch is broken.

1. **Switch point 5 (load-bearing)**: a deterministic test that `bot.py`'s post-response ingestion path does not invoke `memory_extraction.extract_and_store` under `MEMORY_ENABLED=false`, and does invoke it under `MEMORY_ENABLED=true`, holding `memory_extraction_enabled=true` and `effective_backend="claude"` constant. The flip-test is: changing the `memory_is_enabled()` guard to `True` makes the disabled-side test fail.

2. **Switch point 4**: `handle_memory_command` and `handle_memory_callback` return the disabled-mode reply under `is_enabled()=False`, and the full UI under `is_enabled()=True`. The two existing `if not memory.is_enabled():` early-returns are the contract.

3. **Switch point 2**: under the disabled flag the inner Claude's session context exposes only the `Edit MEMORY.md` write path; under the enabled flag it exposes only the `/api/memory/add` path. Inverse-constraint shape: assert the disabled context does NOT contain `/api/memory/add` plumbing, and the enabled context does NOT contain the MEMORY.md inject block. Partial overlap with switch point 1 is fine; the new piece is the symmetric write-contract assertion.

4. **Switch point 6**: `memory.recall` and `memory.extraction` log lines emit under enabled, are silent under disabled. A regression that left logging on under disabled would leak per-user data to the log surface.

A regression that flips any gated branch must fail the harness.

## Out of scope

- Read-side coverage (already shipped in #432/#433).
- The "Error: None" budget exhaustion fix (separate concern).
- Anti-fabrication or per-fact provenance (#430, #431).

## Implementation notes

- The natural extension point is the existing harness module (`src/kai/eval/modeswitch.py`). Either extend `verify` with write-side invariants, or add a sibling subcommand so the read/write split is visible to operators. Prefer the former unless walltime becomes a problem.
- Switch point 5 verification needs a test boundary at the call site in `bot.py`, not at `extract_and_store` itself; the gating happens before the function is called. Monkeypatch the import-site reference and assert call counts under each flag value.
- Switch point 4 can mock the Telegram update/context machinery and assert reply shape under each flag value.
- Switch point 6 is straightforward: capture log records via `caplog` and assert presence/absence under each flag value.

dcellison:
Closes #434.

## Summary

Extends the mode-switch verification harness shipped in PR #433 with regression coverage for the three write-side switch points the issue body flagged as unverified. Switch point 2 (inner-Claude write routing) is intentionally left to the existing read-side partition invariants, which already verify the contract delivered to the inner Claude (the `[Memory subsystem: ...]` marker plus the `[Your persistent memory (file: ...):]` block presence/absence).

## Coverage added

| # | Switch point | Site | New tests |
|---|---|---|---|
| 4 (callback) | `handle_memory_callback` disabled-mode early return | `memory_command.py` | 1 in `test_memory_command.py::TestCallbackDispatch` |
| 5 (behavioral) | extraction-gating guard in `_handle_response` | `bot.py` | 2 in `test_bot.py::TestHandleResponse` |
| 5 (structural) | AST property: every `extract_and_store` call inside `if memory_is_enabled()` | `bot.py` | 1 in `test_eval_modeswitch.py::TestExtractionCallSiteGating` |
| 6 (harness) | `memory.recall` log line `reason` field | `_run_verify` | 2 invariants in `modeswitch.py` |
| 6 (pytest) | same contract via `caplog` | `format_context` | 2 in `test_eval_modeswitch.py::TestRecallReasonField` |

`/memory` slash-command disabled-mode reply was already covered by the pre-existing `tests/test_memory_command.py::TestCommandDispatch::test_memory_disabled_short_circuits`; only the callback handler's parallel early return needed a new test.

## Verify subcommand

`python -m kai.eval.modeswitch verify` now runs 11 invariants (up from 9). The two new ones drive `format_context` against an isolated tmp-dir Mem0 under each flag value, capture `memory.recall` log records via a `logging.Handler` attached to the `kai.memory` logger, and assert the `reason` field of the JSON payload (compares against `_RECALL_REASON_DISABLED` for the disabled mode; asserts the negative for enabled mode so the success-path shape, where the `reason` field is omitted entirely, is also accepted).

Local run output:

```
PASS  disabled: subsystem marker present
PASS  disabled: persistent-memory block injected
PASS  disabled: relevant-memories block omitted
PASS  disabled: recall log reason is 'disabled'
PASS  enabled: subsystem marker present
PASS  enabled: persistent-memory block omitted
PASS  enabled: format_context returns empty or recall-prefixed
PASS  enabled: recall log reason is not 'disabled'
PASS  partition disabled: persistent present, relevant absent
PASS  partition enabled: persistent absent
PASS  partition: mutual exclusion across both modes

PASS: all 11 invariants hold.
```

## Test count

Suite size: 2791 (+6 from the prior baseline).

## Test placement deviation

The implementation plan placed the switch-point-5 behavioral tests in `TestHandleMessage`. The `if memory_is_enabled()` guard is actually in `_handle_response`, not `handle_message`; the existing `TestHandleMessage` tests mock `_handle_response` entirely and never reach the guard. The new behavioral tests live in `TestHandleResponse` (which already has 13 driver-pattern tests using `_make_update`/`_make_context`/`_fake_stream`/`_done_event`). The placement is documented in each test's docstring.

## Regression contracts

- A refactor that flips the `if memory_is_enabled() and chat_id is not None:` guard to `if True:` fails `test_extraction_skipped_when_memory_disabled`.
- A refactor that adds a new ungated `extract_and_store` call site to `bot.py` fails `test_every_extract_and_store_call_site_is_inside_is_enabled_guard`.
- A refactor that removes the early-return in `handle_memory_callback` fails `test_callback_disabled_short_circuits`.
- A refactor that changes `_RECALL_REASON_DISABLED` without updating the emission site fails `test_recall_reason_is_disabled_under_disabled_mode`.

## Test plan

- [x] `make check` clean
- [x] `make test` passes (2791 passed)
- [x] `python -m kai.eval.modeswitch verify` exits 0 with 11 invariants
- [ ] Bot-review pass clean

### DIFF
--- a/src/kai/eval/modeswitch.py
+++ b/src/kai/eval/modeswitch.py
@@ -271,7 +271,17 @@ def _parse_recall_reason(records: list[logging.LogRecord]) -> str | None:
             # since Python 3.5, so the bare `ValueError` covers
             # both the well-formed-but-invalid-JSON case and the
             # malformed-input case that JSONDecodeError raises.
-            return None
+            #
+            # `continue` rather than `return None` so a single
+            # malformed trailing record does not shadow earlier
+            # well-formed records. Today's callers pass a list with
+            # exactly one `memory.recall` record per capture block,
+            # but a future multi-record caller would silently
+            # observe `None` for "valid records exist but the most
+            # recent is bad," which is the wrong shape. The final
+            # `return None` outside the loop still handles the
+            # no-records case.
+            continue
         reason = payload.get("reason") if isinstance(payload, dict) else None
         return reason if isinstance(reason, str) else None
     return None
--- a/src/kai/eval/modeswitch.py
+++ b/src/kai/eval/modeswitch.py
@@ -255,7 +255,11 @@ def _parse_recall_reason(records: list[logging.LogRecord]) -> str | None:
             continue
         try:
             payload = json.loads(msg[len(prefix) :])
-        except (json.JSONDecodeError, ValueError):
+        except ValueError:
+            # `json.JSONDecodeError` is a subclass of `ValueError`
+            # since Python 3.5, so the bare `ValueError` covers
+            # both the well-formed-but-invalid-JSON case and the
+            # malformed-input case that JSONDecodeError raises.
             return None
         reason = payload.get("reason") if isinstance(payload, dict) else None
         return reason if isinstance(reason, str) else None
@@ -470,6 +474,16 @@ async def _run_verify() -> int:
             # log-line contract that downstream eval harnesses depend
             # on (a missing or wrongly-tagged record under disabled
             # mode would silently break those harnesses).
+            #
+            # Deferred import for the same reason `_isolated_data_dir`
+            # imports `kai.memory` lazily: a top-level import would
+            # capture `kai.memory` in `sys.modules` BEFORE the
+            # MEM0_DIR redirect at the start of `_run_verify` could
+            # take effect, deadlocking the harness against the live
+            # `~/.mem0/migrations_qdrant` lock the running service
+            # holds. The `if "kai.memory" in sys.modules:` guard at
+            # the top of `_run_verify` enforces the invariant; this
+            # deferred form preserves it.
             from kai.memory import _RECALL_REASON_DISABLED
 
             with _isolated_data_dir(tmp_root):
--- a/src/kai/eval/modeswitch.py
+++ b/src/kai/eval/modeswitch.py
@@ -222,14 +222,25 @@ def _capture_recall_logs() -> Iterator[list[logging.LogRecord]]:
     # CLI's `logging.basicConfig(level=logging.WARNING, ...)` would
     # otherwise filter INFO-level records out before they reach the
     # handler.
+    #
+    # Disable propagation while the handler is attached so the
+    # captured records do not also fan out to the root logger. The
+    # CLI entry point's `basicConfig(level=WARNING)` already filters
+    # them, but during pytest runs the root logger may be configured
+    # at INFO and the records would appear duplicated in console
+    # output. Restored in `finally` so other code paths that rely on
+    # propagation are not affected after the with-block.
     prior_level = logger.level
+    prior_propagate = logger.propagate
     logger.setLevel(logging.INFO)
+    logger.propagate = False
     logger.addHandler(handler)
     try:
         yield records
     finally:
         logger.removeHandler(handler)
         logger.setLevel(prior_level)
+        logger.propagate = prior_propagate
 
 
 def _parse_recall_reason(records: list[logging.LogRecord]) -> str | None:
--- a/src/kai/eval/modeswitch.py
+++ b/src/kai/eval/modeswitch.py
@@ -505,15 +505,17 @@ async def _run_verify() -> int:
             # on (a missing or wrongly-tagged record under disabled
             # mode would silently break those harnesses).
             #
-            # Deferred import for the same reason `_isolated_data_dir`
-            # imports `kai.memory` lazily: a top-level import would
-            # capture `kai.memory` in `sys.modules` BEFORE the
-            # MEM0_DIR redirect at the start of `_run_verify` could
-            # take effect, deadlocking the harness against the live
-            # `~/.mem0/migrations_qdrant` lock the running service
-            # holds. The `if "kai.memory" in sys.modules:` guard at
-            # the top of `_run_verify` enforces the invariant; this
-            # deferred form preserves it.
+            # Deferred to preserve the `if "kai.memory" in sys.modules:`
+            # invariant guarded at the top of `_run_verify`. A
+            # top-level `from kai.memory import _RECALL_REASON_DISABLED`
+            # in this module would put `kai.memory` into `sys.modules`
+            # before `_run_verify` runs, tripping the assertion. The
+            # constant itself is a string literal and does not boot
+            # Mem0 or touch the migrations_qdrant lock; the deadlock
+            # concern that motivates `_isolated_data_dir`'s lazy
+            # import does not apply directly to this constant. The
+            # consistency-with-the-guard rationale is what carries
+            # the deferral.
             from kai.memory import _RECALL_REASON_DISABLED
 
             with _isolated_data_dir(tmp_root):
--- a/src/kai/eval/modeswitch.py
+++ b/src/kai/eval/modeswitch.py
@@ -245,8 +245,9 @@ def _capture_recall_logs() -> Iterator[list[logging.LogRecord]]:
 
 def _parse_recall_reason(records: list[logging.LogRecord]) -> str | None:
     """Pull the `reason` field from the most-recent `memory.recall`
-    log record in `records`. Returns None if no such record exists
-    or if the record's payload is malformed.
+    log record in `records`. Returns None if no such record exists,
+    if the record's payload is malformed, or if the payload's
+    `reason` field is absent.
 
     The payload format is fixed at the producer
     (`memory._emit_recall_log`): a compact JSON object preceded by
@@ -255,9 +256,17 @@ def _parse_recall_reason(records: list[logging.LogRecord]) -> str | None:
     producer's `log.info("memory.recall %s", json.dumps(payload, ...))`.
 
     Returns the value of the `reason` field cast to `str`, or None
-    on any decode failure or shape mismatch. The harness callers
-    compare against `_RECALL_REASON_DISABLED` (or its absence) and
-    treat None as a separate-and-failing outcome.
+    on any decode failure, shape mismatch, or success-path payload
+    where `reason` is omitted entirely (per `_base_recall_payload`'s
+    docstring: `reason` is set only on short-circuit lines).
+
+    Caller-side interpretation differs by mode. The disabled-mode
+    caller compares `parsed == _RECALL_REASON_DISABLED` and treats
+    None as a failing outcome (it neither equals nor implies the
+    expected reason string). The enabled-mode caller compares
+    `parsed != _RECALL_REASON_DISABLED` and treats None as a
+    PASSING outcome, since the success-path payload omits `reason`
+    entirely and that absence is part of the documented contract.
     """
     prefix = "memory.recall "
     for record in reversed(records):
--- a/src/kai/eval/modeswitch.py
+++ b/src/kai/eval/modeswitch.py
@@ -56,6 +56,7 @@ from __future__ import annotations
 
 import argparse
 import asyncio
+import json
 import logging
 import os
 import re
@@ -178,6 +179,87 @@ def _isolated_data_dir(tmp_root: Path) -> Iterator[Path]:
             # constructs a fresh client against its own DATA_DIR.
             memory_module._memory = None
             memory_module._config = None
+
+
+@contextmanager
+def _capture_recall_logs() -> Iterator[list[logging.LogRecord]]:
+    """Capture every LogRecord emitted to the `kai.memory` logger
+    during the with-block and yield the (initially empty) list to
+    the caller. The list is populated as records arrive and is
+    safe to inspect after the with-block exits.
+
+    Why a manual `logging.Handler` rather than pytest's `caplog`
+    fixture: the harness runs from a CLI entry point, NOT under
+    pytest, so the fixture is unavailable. The pytest test pair at
+    `tests/test_eval_modeswitch.py::TestRecallReasonField` uses
+    `caplog` (the idiomatic pytest tool); the two surfaces capture
+    the same records via different mechanisms, both correct for
+    their context.
+
+    Records are appended in the order they arrive. The harness's
+    callers only inspect the most-recent record (a single
+    `format_context` call is made per use of this contextmanager),
+    but the list keeps prior records too in case future callers
+    want to assert on multi-call sequences.
+
+    Each fresh use creates a new `_ListHandler` instance and a new
+    list, so a residual handler from a prior run cannot pollute
+    THIS run's records. Removal in the `finally` block guards
+    against handler leaks across calls.
+    """
+    records: list[logging.LogRecord] = []
+
+    class _ListHandler(logging.Handler):
+        def emit(self, record: logging.LogRecord) -> None:
+            records.append(record)
+
+    handler = _ListHandler()
+    handler.setLevel(logging.INFO)
+    logger = logging.getLogger("kai.memory")
+    # Save and lower the logger level so the `memory.recall` line
+    # (emitted via `log.info(...)`) is forwarded to our handler
+    # regardless of the harness's outer logging configuration. The
+    # CLI's `logging.basicConfig(level=logging.WARNING, ...)` would
+    # otherwise filter INFO-level records out before they reach the
+    # handler.
+    prior_level = logger.level
+    logger.setLevel(logging.INFO)
+    logger.addHandler(handler)
+    try:
+        yield records
+    finally:
+        logger.removeHandler(handler)
+        logger.setLevel(prior_level)
+
+
+def _parse_recall_reason(records: list[logging.LogRecord]) -> str | None:
+    """Pull the `reason` field from the most-recent `memory.recall`
+    log record in `records`. Returns None if no such record exists
+    or if the record's payload is malformed.
+
+    The payload format is fixed at the producer
+    (`memory._emit_recall_log`): a compact JSON object preceded by
+    the literal prefix `"memory.recall "`. Splitting on the first
+    space and `json.loads`-decoding the tail is the inverse of the
+    producer's `log.info("memory.recall %s", json.dumps(payload, ...))`.
+
+    Returns the value of the `reason` field cast to `str`, or None
+    on any decode failure or shape mismatch. The harness callers
+    compare against `_RECALL_REASON_DISABLED` (or its absence) and
+    treat None as a separate-and-failing outcome.
+    """
+    prefix = "memory.recall "
+    for record in reversed(records):
+        msg = record.getMessage()
+        if not msg.startswith(prefix):
+            continue
+        try:
+            payload = json.loads(msg[len(prefix) :])
+        except (json.JSONDecodeError, ValueError):
+            return None
+        reason = payload.get("reason") if isinstance(payload, dict) else None
+        return reason if isinstance(reason, str) else None
+    return None
 
 
 def _build_test_configs(memory_enabled_value: bool) -> Config:
@@ -377,6 +459,40 @@ async def _run_verify() -> int:
                 )
             )
 
+            # Disabled-mode recall-log reason invariant. Drives
+            # `format_context` under a config with memory_enabled=False
+            # so init_memory short-circuits at its
+            # `if not config.memory_enabled: return` guard, leaving
+            # `_memory` and `_config` as None. format_context then
+            # short-circuits via its `if not is_enabled() or _config
+            # is None:` branch and emits a `memory.recall` log line
+            # with `reason: "disabled"`. The invariant verifies the
+            # log-line contract that downstream eval harnesses depend
+            # on (a missing or wrongly-tagged record under disabled
+            # mode would silently break those harnesses).
+            from kai.memory import _RECALL_REASON_DISABLED
+
+            with _isolated_data_dir(tmp_root):
+                from kai import memory as memory_module
+
+                memory_module.init_memory(_build_test_configs(memory_enabled_value=False))
+                with _capture_recall_logs() as disabled_recall_records:
+                    await memory_module.format_context(_FIXTURE_QUERY, user_id=str(_FIXTURE_CHAT_ID))
+
+            disabled_reason = _parse_recall_reason(disabled_recall_records)
+            disabled_reason_ok = disabled_reason == _RECALL_REASON_DISABLED
+            results.append(
+                _InvariantResult(
+                    name="disabled: recall log reason is 'disabled'",
+                    passed=disabled_reason_ok,
+                    detail=(
+                        ""
+                        if disabled_reason_ok
+                        else f"got reason={disabled_reason!r}, expected {_RECALL_REASON_DISABLED!r}"
+                    ),
+                )
+            )
+
             # ── Enabled-mode invariants ───────────────────────────
             enabled_ctx = build_session_context(
                 workspace=ws,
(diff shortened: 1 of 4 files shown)


dcellison:
## Review by Kai

The repo being reviewed (dcellison/kai) is not checked out locally - this is a remote PR review. I have enough from the diff and the bot.py exploration to complete the review. The critical finding from the Explore agent is confirmed: `memory_is_enabled` is imported with `from kai.memory import is_enabled as memory_is_enabled` at line 3638 of bot.py - a bound reference, not a module-qualified call.

---

## PR Review: feat(memory) write-side mode-switch verification (#434)

### Critical

None.

### Warning

**`test_bot.py` - monkeypatch target misses the production call site**

Both new `TestHandleResponse` tests patch:
```python
monkeypatch.setattr("kai.memory.is_enabled", lambda: False)
```

But `bot.py:3638` uses:
```python
from kai.memory import is_enabled as memory_is_enabled
```

Because `memory_is_enabled` is a bound name in `kai.bot`'s namespace (not a module-qualified lookup), patching `kai.memory.is_enabled` replaces the attribute on the `kai.memory` module object but leaves `kai.bot.memory_is_enabled` pointing at the original function. The guard at `bot.py:3640` never sees the mock.

The practical consequence: `test_extraction_skipped_when_memory_disabled` likely passes vacuously - `is_enabled()` still returns `True` in bot.py, but because `_ingest_memory` is a nested closure defined inside the `if memory_is_enabled()` block, the real `is_enabled()` returning `True` means the task IS created; the test then passes only if `extract_and_store` doesn't get called for some other reason (e.g., a missing session_id or a config guard). That's a silent false positive. The regression contract claimed in the docstring ("a refactor that flips the guard fails this test") is not actually pinned.

Fix:
```python
monkeypatch.setattr("kai.bot.memory_is_enabled", lambda: False)
```

The `extract_and_store` patch is correct as-is - `bot.py` uses the module-qualified `memory_extraction.extract_and_store(...)` form, so patching `kai.memory_extraction.extract_and_store` reaches the call site.

---

### Suggestion

**`modeswitch.py:_parse_recall_reason` - redundant exception types**

```python
except (json.JSONDecodeError, ValueError):
```

`json.JSONDecodeError` is a subclass of `ValueError` since Python 3.5. Catching `ValueError` already covers `JSONDecodeError`; listing both is harmless but misleading.

---

**`test_bot.py` - `asyncio.all_tasks()` is fragile**

```python
new_tasks = asyncio.all_tasks() - {asyncio.current_task()}
```

This gathers every task on the event loop, including any that pytest-asyncio or other concurrently-running fixtures may have created. A snapshot diff would be more robust:

```python
tasks_before = asyncio.all_tasks()
await _handle_response(...)
new_tasks = asyncio.all_tasks() - tasks_before - {asyncio.current_task()}
```

This is unlikely to cause false failures in practice because pytest-asyncio typically provides a fresh loop per test, but the pattern is fragile.

---

**`modeswitch.py` - mid-function import placement**

```python
# inside _run_verify():
from kai.memory import _RECALL_REASON_DISABLED
```

This import appears mid-function, interleaved with the results list construction. The convention in the surrounding code is to put module-level imports at the top. If this must be deferred (e.g., to avoid circular imports or to match how the other `from kai import memory as memory_module` deferred imports work), a comment explaining why would help. Currently there's no explanation for why this one import is deferred while others are at module level.

---

### Overall

The structural AST test (`TestExtractionCallSiteGating`) is well-designed and correctly handles the `BoolOp(And, [Call, Compare])` shape by walking `if_node.test` recursively rather than doing a direct comparison. The `_capture_recall_logs` context manager properly saves/restores the logger level and removes the handler in a `finally` block. The `reset_memory_module` fixture teardown is unconditional, which is the right pattern.

The main actionable item is the monkeypatch target in the two `TestHandleResponse` tests - it needs to be `kai.bot.memory_is_enabled` to actually pin the regression it claims to pin.

dcellison:
## Review by Kai

Looking at the diff against the prior review thread. Per instructions, I am not re-raising the monkeypatch target warning (`kai.memory.is_enabled` vs `kai.bot.memory_is_enabled`) since the code is unchanged and the author has seen it.

---

## PR Review: feat(memory) write-side mode-switch verification (#435)

### Prior review follow-ups

All three suggestions were addressed:

- `asyncio.all_tasks()` fragility - fixed with `tasks_before` snapshot in both `test_bot.py` tests.
- Mid-function deferred import comment - added with a clear explanation tying it to the `sys.modules` guard at the top of `_run_verify`.
- `except (json.JSONDecodeError, ValueError)` - collapsed to bare `except ValueError` with an explanatory comment confirming the subclass relationship.

### New findings

**Suggestion - `test_eval_modeswitch.py:306` - assertion ordering exposes implicit floor dependency**

```python
assert result.startswith(_MARKER_RELEVANT_MEMORIES), f"..."
reason = modeswitch._parse_recall_reason(caplog.records)
assert reason != memory_module._RECALL_REASON_DISABLED, ...
```

The first assertion assumes the canned result (`score=0.95`) clears the relevance floor configured by `_build_test_configs`. If that floor is ever tuned above 0.95, `format_context` returns `""` instead of the recall-prefixed string, the first assert fires, and the reason assertion - the actual contract under test - is never reached. The two assertions are independent; swapping their order (check reason first, then output shape) makes the test fail with the more informative message under a floor regression rather than an output-shape failure that looks unrelated to this test's stated purpose.

**Suggestion - `modeswitch.py:_capture_recall_logs` - propagation noise**

The `kai.memory` logger's default `propagate=True` means every captured `INFO` record also flows to the root logger. Under the CLI entry point (where `logging.basicConfig(level=logging.WARNING)` is set) this is filtered, but during `pytest` runs the root logger may be at `INFO`, causing the captured records to appear in console output. Setting `handler.propagate = False` is not possible on handlers, but adding `logger.propagate = False` inside the context and restoring it in `finally` would suppress the double-output. Low impact since it's test/eval code only.

**Suggestion - `test_eval_modeswitch.py:TestExtractionCallSiteGating._guard_contains_memory_is_enabled`**

The predicate matches any `Call` whose `func.id == "memory_is_enabled"` anywhere inside `if_node.test`, including inside a `UnaryOp(Not, ...)`. A guard written as `if not memory_is_enabled():` would satisfy the predicate but has inverted semantics - code inside it runs only when memory is disabled, which is the opposite of the intended gate. The current production shape (`if memory_is_enabled() and ...`) is not at risk, but the docstring could note this limitation so a future reader does not assume the test catches inverted guards.

### Overall

Clean structural additions. The `_capture_recall_logs` context manager's save/restore of `logger.level` is correct. The `_attach_parents` / parent-chain walk in `TestExtractionCallSiteGating` correctly handles the `BoolOp(And, [Call, Compare])` production shape via recursive `ast.walk`. The `test_callback_disabled_short_circuits` monkeypatch target (`memory_command.memory.is_enabled`) is correct for the module-qualified call form used in `memory_command.py`. No critical or warning-level findings beyond what was already raised in the prior review.

dcellison:
## Review by Kai

## PR Review: feat(memory) write-side mode-switch verification (#435)

### Critical

None.

### Warning

None.

### Suggestion

**`modeswitch.py:_parse_recall_reason` - early return on malformed record shadows valid earlier records**

```python
for record in reversed(records):
    msg = record.getMessage()
    if not msg.startswith(prefix):
        continue
    try:
        payload = json.loads(msg[len(prefix):])
    except ValueError:
        return None   # <-- stops here even if an earlier record is valid
```

If the most-recent `memory.recall` record has a malformed JSON payload, the function returns `None` immediately rather than `continue`-ing to earlier records. For the current callers (single `format_context` call per capture block), this is harmless - there is exactly one `memory.recall` record per call. But if a future caller passes a multi-record list with a bad trailing entry, the valid earlier records are silently unreachable. `continue` instead of `return None` on the `except ValueError` branch would be safer and costs nothing:

```python
except ValueError:
    continue
```

The final `return None` outside the loop still handles the no-records case.

---

**`test_eval_modeswitch.py:TestExtractionCallSiteGating._guard_contains_memory_is_enabled` - polarity limitation undocumented at the call site**

The method's docstring correctly notes the polarity limitation (an inverted `not memory_is_enabled()` guard satisfies the predicate). But `test_every_extract_and_store_call_site_is_inside_is_enabled_guard` - which calls this predicate - does not reference the limitation in its own docstring or assertions, so a reader maintaining only the test function would not know the structural check cannot catch inverted guards. A one-line cross-reference in the test docstring pointing to the behavioral counterpart in `TestHandleResponse` would close the gap.

---

### Addressed from prior review

All three prior suggestions are cleanly resolved:
- `asyncio.all_tasks()` fragility fixed with `tasks_before` snapshot in both `test_bot.py` tests.
- `except (json.JSONDecodeError, ValueError)` collapsed to `except ValueError` with explanatory comment.
- Mid-function `from kai.memory import _RECALL_REASON_DISABLED` now carries an explicit comment tying the deferral to the `sys.modules` guard rationale at the top of `_run_verify`.

### Overall

The `_capture_recall_logs` context manager is correctly structured - save/restore of both `logger.level` and `logger.propagate`, `removeHandler` in an unconditional `finally`, fresh handler per invocation. The `TestExtractionCallSiteGating` AST walk is correct for the production `BoolOp(And, [Call, Compare])` shape. The `reset_memory_module` fixture teardown is unconditional. The `test_callback_disabled_short_circuits` monkeypatch target (`memory_command.memory.is_enabled`) is correct for the module-qualified call form in `memory_command.py`. No critical or warning-level findings.

dcellison:
## Review by Kai

The PR branch isn't checked out locally; I have everything I need from the diff and the prior thread. Working from the diff directly.

---

## PR Review: feat(memory) write-side mode-switch verification (#435)

### Critical

None.

### Warning

None.

### New suggestions

**`modeswitch.py:_parse_recall_reason` docstring - None description is caller-dependent**

The closing sentence of the docstring reads:

```
The harness callers compare against `_RECALL_REASON_DISABLED` (or its absence) and
treat None as a separate-and-failing outcome.
```

This is accurate for the disabled-mode caller (`disabled_reason == _RECALL_REASON_DISABLED`), where `None` produces `False`. It is inaccurate for the enabled-mode caller (`enabled_reason != _RECALL_REASON_DISABLED`), where `None` is a passing outcome - it's the documented success-path shape when the `reason` field is omitted from the payload. A future maintainer relying on the docstring alone would expect both callers to reject `None`, which misrepresents the enabled-mode contract. Recommend splitting the description:

```
Disabled-mode callers treat None as failing (it neither equals nor
implies the expected reason string). Enabled-mode callers accept None
as the success-path shape, where the `reason` field is omitted entirely.
```

---

**`test_eval_modeswitch.py:TestExtractionCallSiteGating.test_every_extract_and_store_call_site_is_inside_is_enabled_guard` - call-site count assertion gives no count context**

```python
assert call_sites, "expected at least one `extract_and_store` call site in bot.py"
```

If the assertion fires (someone removed the call site), the message alone doesn't tell a reader how many call sites the test previously found. Adding the current count to the success path - either as a comment or as a complementary `len(call_sites)` in the message - would give future readers a baseline without changing the assertion semantics:

```python
assert call_sites, (
    "expected at least one `extract_and_store` call site in bot.py - "
    "the production path in _handle_response's post-response block was removed"
)
```

Minor; the existing message is serviceable.

---

### Previously raised, not re-raised

The `kai.memory.is_enabled` vs `kai.bot.memory_is_enabled` monkeypatch target warning from the first review pass is not re-raised; the code is unchanged and the author has seen it.

### Overall

The `_capture_recall_logs` context manager is correctly structured - fresh handler and list per invocation, `logger.propagate = False` saved/restored in `finally`, `removeHandler` unconditional. The `_parse_recall_reason` `continue`-on-ValueError fix from the prior review is in place. The deferred `from kai.memory import _RECALL_REASON_DISABLED` is placed after the existing `_isolated_data_dir` blocks in `_run_verify` execution order, so `kai.memory` is already in `sys.modules` when it runs - the placement is safe. The `test_callback_disabled_short_circuits` monkeypatch target (`memory_command.memory.is_enabled`) is correct for the module-qualified call form. No critical or warning-level new findings.

dcellison:
## Review by Kai

## PR Review: feat(memory) write-side mode-switch verification (#435)

After four prior review rounds, the code is in good shape. Two minor new findings below.

### Critical

None.

### Warning

None.

### Suggestion

**`test_bot.py` - `return_exceptions=True` hides task failure in the positive test**

In `test_extraction_invoked_when_memory_enabled`:

```python
if new_tasks:
    await asyncio.gather(*new_tasks, return_exceptions=True)
extract_mock.assert_called_once()
```

If `_ingest_memory` raises before reaching `extract_and_store` (e.g., a config guard or missing attribute raises mid-task), the exception is swallowed by `return_exceptions=True` and the failure surfaces only as `AssertionError: Expected 'mock' to have been called once` with no indication of the underlying cause. In the disabled-mode counterpart, swallowing a task exception is safe because a task appearing at all is already the regression; in the enabled-mode test, a crashing task and a missing task are indistinguishable from the assertion message. Removing `return_exceptions=True` here propagates the task exception directly, making the failure unambiguous. The disabled-mode test is fine as-is.

---

**`test_eval_modeswitch.py` - `_attach_parents` docstring contradicts the type-ignore it accompanies**

```python
child.parent = parent  # type: ignore[attr-defined]
```

The docstring says "pyright tolerates it under the strict-but-permissive attribute model," but the `# type: ignore[attr-defined]` on the same line proves the opposite: pyright flags it as an error and the ignore is silencing that error. The comment is wrong and would mislead a reader deciding whether the ignore is load-bearing. Suggest replacing "pyright tolerates it" with "pyright flags this as an attr-defined error on strict AST nodes, hence the ignore."

---

### Previously raised, not re-raised

The `kai.memory.is_enabled` vs `kai.bot.memory_is_enabled` monkeypatch target mismatch in the two `TestHandleResponse` tests was raised in the first review pass. Code is unchanged; not re-raised.

### Overall

The `_capture_recall_logs` context manager is correctly structured: fresh handler and list per invocation, `prior_propagate` saved and restored unconditionally in `finally`, `removeHandler` in `finally`. The `_parse_recall_reason` `continue`-on-ValueError fix from the prior round is present. The deferred `from kai.memory import _RECALL_REASON_DISABLED` is placed after the prior `_isolated_data_dir` calls in execution order, so `kai.memory` is already in `sys.modules` when it runs. `test_callback_disabled_short_circuits` uses the correct monkeypatch target (`memory_command.memory.is_enabled`) for the module-qualified call form in `memory_command.py`. No critical or warning-level new findings.

dcellison:
## Review by Kai

The `object()` sentinel in the enabled-mode test is safe - confirmed `format_context` routes through the module-level `search()` wrapper and `is_enabled()` only checks `_memory is not None`.

---

## PR Review: feat(memory) write-side mode-switch verification (#435)

### Critical

None.

### Warning

None.

### Suggestion

**`test_eval_modeswitch.py` - `object()` sentinel coupling is undocumented at the setup site**

```python
memory_module._memory = object()
```

This works because `format_context` never calls any attribute on `_memory` directly - it routes everything through the module-level `search()` wrapper, and `is_enabled()` only checks `_memory is not None`. That invariant is load-bearing for the test but is not stated anywhere near the assignment. If a future refactor adds `_memory.client_ready()` or similar to the `format_context` hot path, the test starts raising `AttributeError` instead of failing cleanly, and the failure would be confusing. A one-line comment would close this gap:

```python
# Truthy sentinel only - format_context never calls attrs on _memory directly;
# is_enabled() checks `_memory is not None` and all search dispatches go through
# the module-level search() wrapper, which is mocked below.
memory_module._memory = object()
```

---

### Previously raised, not re-raised

- `kai.memory.is_enabled` vs `kai.bot.memory_is_enabled` monkeypatch target in the two `TestHandleResponse` tests - raised in round 1, code unchanged.

### Overall

Clean. The `_capture_recall_logs` context manager is correctly structured - fresh handler and list per invocation, `prior_propagate` and `prior_level` both restored unconditionally in `finally`, `removeHandler` in `finally`. The `_parse_recall_reason` `continue`-on-`ValueError` fix from round 3 is in place. The deferred `from kai.memory import _RECALL_REASON_DISABLED` is correctly placed after the prior `_isolated_data_dir` calls, so `kai.memory` is already in `sys.modules`. The `_guard_contains_memory_is_enabled` polarity-caveat is now documented in both the predicate docstring and the test docstring. The `return_exceptions=True` asymmetry between the disabled and enabled gather calls is correct and well-explained. No critical or warning-level new findings.

dcellison:
## Review by Kai

## PR Review: feat(memory) write-side mode-switch verification (#435)

### Critical

None.

### Warning

None.

### Suggestions

**`modeswitch.py` - `enabled_recall_records` scoping could surprise future readers**

`enabled_recall_records` is bound by the `as` clause of a `with` statement nested inside another `with` block:

```python
with _isolated_data_dir(tmp_root):
    ...
    with _capture_recall_logs() as enabled_recall_records:
        recall_text = await memory_module.format_context(...)

enabled_reason = _parse_recall_reason(enabled_recall_records)  # used here
```

Python's flat scoping means this works correctly - `enabled_recall_records` is accessible after both context managers exit. But a reader who comes from a language with block scoping may assume the variable is out of scope. A one-line comment above `_parse_recall_reason(enabled_recall_records)` noting that the list is populated during the `with` block and remains valid after exit would pre-empt that confusion. The disabled counterpart already has this pattern and doesn't need a comment because it reads `disabled_recall_records` immediately and visually close to the block; the enabled case has the `recall_ok` result block interleaved, pushing the two uses further apart.

**`test_eval_modeswitch.py:TestRecallReasonField` - `reset_memory_module` fixture comment undersells its scope**

The fixture docstring says it saves/restores `_memory` and `_config`. It also runs `memory_module._memory = None` and `memory_module._config = None` at setup time (before `yield`), not just at teardown. The "Save and restore" framing implies a pure teardown fixture; the explicit setup mutation is load-bearing for both tests (it ensures the disabled-mode test starts with cleared singletons and is the state `test_recall_reason_is_not_disabled_under_enabled_mode` overwrites). Noting the setup reset in the docstring - "resets both singletons to None before yielding, then restores" - would make the fixture contract unambiguous.

### Overall

All prior suggestion-level findings from rounds 1-5 are cleanly resolved. The `_capture_recall_logs` save/restore of `prior_level` and `prior_propagate` is unconditional in `finally`. The `_parse_recall_reason` `continue`-on-`ValueError` fix is in place. The disabled-mode recall invariant correctly uses a fresh `_isolated_data_dir` block rather than leaking state from the earlier disabled-mode blocks. The `test_callback_disabled_short_circuits` monkeypatch target (`memory_command.memory.is_enabled`) is correct for the module-qualified form in `memory_command.py`. No critical or warning-level new findings.

dcellison:
## Review by Kai

Looking at this diff after seven prior review rounds, focusing only on new findings not already raised.

## PR Review: feat(memory) write-side mode-switch verification (#435)

### Critical

None.

### Warning

None.

### Suggestions

**`test_eval_modeswitch.py:TestRecallReasonField` - `reset_memory_module` fixture ordering creates hidden coupling with `monkeypatch`**

Both tests in `TestRecallReasonField` use `monkeypatch` alongside `reset_memory_module`. The fixture's `finally` block restores `_memory` and `_config` unconditionally. But `monkeypatch.setattr(memory_module, "search", ...)` in `test_recall_reason_is_not_disabled_under_enabled_mode` targets the same module object whose `_memory` attribute the fixture also touches. Fixture teardown order in pytest runs fixtures in reverse-declaration order, so `reset_memory_module` (the first fixture) tears down AFTER `monkeypatch` (the second). The ordering is safe here because `reset_memory_module` only restores `_memory`/`_config` while monkeypatch only restores `search`, but the fixture docstring gives no hint that it was designed to coexist with a simultaneously-active monkeypatch on the same module. A one-line note ("safe to combine with monkeypatch patches to other attributes on the same module object - fixture and monkeypatch restore disjoint attributes") would prevent a future test author from accidentally adding a `monkeypatch.setattr(memory_module, "_memory", ...)` alongside the fixture and creating a teardown conflict (monkeypatch would restore `_memory` first, then the fixture would overwrite it with the saved value, leaving the module in unexpected state).

Low-impact; teardown ordering is deterministic here. Documentation gap only.

---

**`modeswitch.py` - disabled-mode recall block rebinds `memory_module` without explicit guard**

The disabled-mode recall block introduces a `from kai import memory as memory_module` inside its `_isolated_data_dir` scope, on the line immediately following the existing disabled-mode `_isolated_data_dir` block that also imports `memory_module`. In `_run_verify`'s overall flow, `memory_module` is a local name reused across multiple blocks. If the disabled-mode recall block's `_isolated_data_dir` exits (resetting `_memory`/`_config` to None) and a future edit splits the `enabled_reason` block into its own `_isolated_data_dir` that again rebinds `memory_module`, the name would point to the module as initialized under the NEW isolated dir. The current code is safe because `enabled_recall_records` is captured inside the `_isolated_data_dir` block and `_parse_recall_reason` consumes the already-captured list. This is a latent fragility if the block structure is refactored. Not actionable now; noting for future maintenance awareness.

---

### Previously raised, not re-raised

The `kai.memory.is_enabled` vs `kai.bot.memory_is_enabled` monkeypatch target mismatch in `test_extraction_skipped_when_memory_disabled` and `test_extraction_invoked_when_memory_enabled` was raised as a Warning in round 1. Code is unchanged; not re-raised.

---

### Overall

Seven rounds of review have produced clean, well-documented additions. The `_capture_recall_logs` context manager correctly saves/restores `prior_level` and `prior_propagate` unconditionally in `finally`, with `removeHandler` also in `finally`. `_parse_recall_reason` has the `continue`-on-`ValueError` fix from round 3. The deferred `from kai.memory import _RECALL_REASON_DISABLED` is correctly placed after prior `_isolated_data_dir` blocks. `TestExtractionCallSiteGating` correctly handles the `BoolOp(And, [Call, Compare])` production shape. `test_callback_disabled_short_circuits` uses the correct monkeypatch target for the module-qualified call form in `memory_command.py`. The `return_exceptions` asymmetry between the disabled and enabled gather calls is correct. No critical or warning-level new findings.

PR #441, opened 2026-05-07
feat(memory): per-fact speaker attribution with retrieval-time downweight

dcellison:
## Summary

- Adds per-fact speaker attribution (user / assistant / episode_summary) and reuses the existing confidence field for retrieval-time and browse-time demotion. Setting speaker weights to 1.0 and treating missing confidence as 1.0 reproduces the prior ranking minus the retired source-weight boosts.
- Replaces the older `_SOURCE_WEIGHTS` / `_source_weight` source-as-quality-proxy with `_SPEAKER_WEIGHTS` / `_speaker_weight`. Demote-only by design (every value in [0.0, 1.0]); raw cosine remains an upper bound on adjusted score, which keeps the floor filter intuitive.
- /memory fact list defaults to weighted-quality sort with a Sort: Quality / Sort: Recent toggle. Fact detail renders Speaker and Confidence; episode detail renders Speaker only.

## Implementation map

- `kai.memory`: four module-scope constants (`_MIGRATION_SPEAKER` / `_MIGRATION_CONFIDENCE` / `_LEGACY_SPEAKER` / `_LEGACY_CONFIDENCE`), `_read_time_speaker` helper, `_SPEAKER_WEIGHTS` / `_speaker_weight`, per-hit recall log payload extended with `speaker` and `confidence`, `build_migration_metadata` helper.
- `kai.memory_extraction`: fact schema bumps to v8 with required `speaker` enum, prompt gains a `speaker` field block with four worked examples, `_validate_facts` defense-in-depth check forces `speaker="assistant"` when a fact's `confirmation_quote` appears verbatim in an ASSISTANT message in the window. `_generate_episode` pins `speaker="episode_summary"` and `confidence=1.0` at write time.
- `kai.memory_command`: `_send_facts_list` accepts a `sort` kwarg with quality-default re-sort, `_build_facts_list_view` renders the toggle row, `_build_fact_view` renders Speaker / Confidence per source, callback handler parses the new sort segment.
- `kai.eval.retrieval`: `ConfigOverride` carries three speaker-weight axes; `_apply_override` / `_restore_overrides` helpers replace the inline mutation block in `run_sweep`. CLI flags `--user-weight`, `--assistant-weight`, `--episode-summary-weight` replace the retired `--extracted-weight`. Default sweep grid is 24 speaker-weight combinations when floor and overfetch are pinned at production.
- `scripts/migrate-memory-md-to-qdrant.py`: routes through `build_migration_metadata` so the script and the tests share one code path.

## Legacy default rationale

Extracted-legacy rows (`source = "extracted"` with no speaker in metadata) default to `_LEGACY_SPEAKER = "assistant"` and `_LEGACY_CONFIDENCE = 0.5`. The pre-spec extractor produced facts indiscriminately from both speakers AND stripped third-person voice from extracted text, so post-hoc classification of these rows from prose alone is not recoverable. The conservative default biases toward demotion of unclassifiable content; a user-claimed fact mistakenly defaulted to assistant loses 30% retrieval weight, while an assistant-claimed fact mistakenly defaulted to user would inflate to full weight. The first error degrades; the second masks. If production data shows the default is wrong, swap the constants in a same-day PR; the change is purely read-side, no Qdrant rewrite required.

Migration rows are operator-authored MEMORY.md content. `_MIGRATION_SPEAKER = "user"` and `_MIGRATION_CONFIDENCE = 0.9` (the modest discount reflects that imported text is not freshly conversational).

## Calibration plan

Step 11 of the implementation order is the calibration sweep against a production snapshot. The harness changes ship with this PR so the sweep can run without a second deploy. The §5 hard floor compares against the pre-spec weighted baseline (the old `_SOURCE_WEIGHTS` table on the same probes against the same snapshot); a configuration is accepted only when p@1 stays within 0.05 of that baseline AND p@3 OR MRR is at least 0.05 above. The 24-configuration sweep table and five before / after sample retrieval lists land in the closing comment per the issue body's acceptance criterion.

## Test plan

- [x] All 2818 existing tests pass (`make test`).
- [x] `make check` passes (ruff lint + ruff format).
- [x] §6.1 round-trip tests verify speaker and confidence preserve through Mem0 / Qdrant for all three speaker values plus the two distinct legacy-default fallbacks.
- [x] §6.2 weight function tests pin speaker_weight times confidence and the demote-only invariant across every (in-enum speaker, in-range confidence) pair.
- [x] §6.3 format-context tests pin the new weighted-score ordering, the floor-applies-to-raw-cosine property, and the new log payload fields.
- [x] §6.4 extractor speaker tests pin the schema, the conservative default, and the defense-in-depth assistant-quote override.
- [x] §6.5 episode storage test pins speaker / confidence on the metadata bundle.
- [x] §6.6 migration helper test pins the build_migration_metadata dict shape.
- [x] §6.7 /memory rendering tests pin the default sort, toggle button shape, and per-source Speaker / Confidence rendering including the legacy-default fallback path.
- [x] §6.8 eval-harness apply / restore tests pin the round-trip restore property the sweep finally block depends on.
- [ ] Calibration sweep against the production snapshot (operator-driven; this PR ships the harness).
- [ ] End-to-end smoke against the production snapshot (operator-driven).

### DIFF
--- a/scripts/migrate-memory-md-to-qdrant.py
+++ b/scripts/migrate-memory-md-to-qdrant.py
@@ -314,7 +314,7 @@ def _do_forward_migration(args: argparse.Namespace, chunks: list[Chunk]) -> int:
 
     Returns the script exit code.
     """
-    from kai.memory import add_structured, search
+    from kai.memory import add_structured, build_migration_metadata, search
 
     user_id_str = str(args.user_id)
     scores: list[float] = []
@@ -356,11 +356,12 @@ def _do_forward_migration(args: argparse.Namespace, chunks: list[Chunk]) -> int:
                 user_id=user_id_str,
                 memory_type="fact",
                 tags=tags,
-                metadata={
-                    "source": "migration",
-                    "section": section,
-                    "subsection": subsection,
-                },
+                # build_migration_metadata centralizes the migration-
+                # row metadata shape so the writer here and the tests
+                # in tests/test_memory.py drive the same code path.
+                # The helper layers in `speaker` / `confidence` (the
+                # migration constants) on top of section / subsection.
+                metadata=build_migration_metadata(section=section, subsection=subsection),
             )
             if mid is None:
                 # add_structured returns None on store failure or
--- a/src/kai/eval/retrieval.py
+++ b/src/kai/eval/retrieval.py
@@ -32,7 +32,7 @@ Design rationale (the part that is not obvious from the code):
   numerator and denominator.
 
 - The sweep mutates module-level state in `kai.memory`
-  (`_SOURCE_WEIGHTS`, `_SEARCH_OVERFETCH`) and replaces `_config` with
+  (`_SPEAKER_WEIGHTS`, `_SEARCH_OVERFETCH`) and replaces `_config` with
   a `dataclasses.replace(...)` clone for each grid point. Restoration
   in `try/finally` defends against a probe raising mid-sweep. The
   process is short-lived and has no other consumers of the memory
@@ -58,6 +58,8 @@ from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any
 
+from kai.config import Config
+
 log = logging.getLogger(__name__)
 
 
@@ -74,22 +76,28 @@ log = logging.getLogger(__name__)
 _K_VALUES: tuple[int, ...] = (1, 3, 5)
 
 
-# Default sweep grid. Every grid axis multiplied gives 6*4*3 = 72
-# configurations; at typical per-call latency (50-200 ms) the full
-# sweep runs in a few minutes for a probe set of a few dozen probes.
-# Operators can narrow each axis via CLI flags.
+# Default sweep grid. Three speaker-weight axes plus floor and
+# overfetch. The full Cartesian product is 6 * 2 * 4 * 3 * 3 = 432
+# configurations; the calibration plan in spec §5 holds floor and
+# overfetch at production values, leaving 24 configurations for the
+# typical sweep run. Operators can narrow each axis via CLI flags.
 _DEFAULT_FLOOR_GRID: tuple[float, ...] = (0.15, 0.20, 0.25, 0.30, 0.35, 0.40)
-_DEFAULT_EXTRACTED_WEIGHT_GRID: tuple[float, ...] = (1.0, 1.2, 1.5, 2.0)
+_DEFAULT_USER_WEIGHT_GRID: tuple[float, ...] = (0.85, 1.0)
+_DEFAULT_ASSISTANT_WEIGHT_GRID: tuple[float, ...] = (0.5, 0.6, 0.7, 0.8)
+_DEFAULT_EPISODE_SUMMARY_WEIGHT_GRID: tuple[float, ...] = (0.7, 0.85, 1.0)
 _DEFAULT_OVERFETCH_GRID: tuple[int, ...] = (10, 20, 30)
 
 
 # Production defaults that the harness treats as the baseline
 # configuration. Used both when the operator runs a single eval
 # (no --sweep) without overrides and when computing the schema
-# example baseline.
+# example baseline. The three speaker weights map to the entries
+# in `_SPEAKER_WEIGHTS`; the legacy/extracted source-weight pair
+# the older table carried no longer exists.
 _PRODUCTION_FLOOR = 0.30
-_PRODUCTION_EXTRACTED_WEIGHT = 1.2
-_PRODUCTION_LEGACY_WEIGHT = 0.6
+_PRODUCTION_USER_WEIGHT = 1.0
+_PRODUCTION_ASSISTANT_WEIGHT = 0.7
+_PRODUCTION_EPISODE_SUMMARY_WEIGHT = 0.85
 _PRODUCTION_OVERFETCH = 20
 
 
@@ -134,17 +142,37 @@ class ConfigOverride:
 
     Each override flows into the harness's sweep loop and is the
     only mutable state per iteration: floor (replaces
-    `_config.memory_search_floor`), extracted_weight (replaces
-    `_SOURCE_WEIGHTS["extracted"]`), and overfetch (replaces
-    `_SEARCH_OVERFETCH` on the module). The legacy weight (the ""
-    bucket in `_SOURCE_WEIGHTS`, currently 0.6) is intentionally
-    fixed: the eval is for tuning the "extracted" source's relative
-    boost, not for re-tuning the legacy floor.
+    `_config.memory_search_floor`), three speaker weights (replace
+    the matching entries in `_SPEAKER_WEIGHTS`), and overfetch
+    (replaces `_SEARCH_OVERFETCH` on the module). The unknown-
+    speaker fallback weight is intentionally not part of the grid:
+    its job is to keep an unrecognized speaker class rankable, not
+    to be tuned.
     """
 
     floor: float
-    extracted_weight: float
+    user_weight: float
+    assistant_weight: float
+    episode_summary_weight: float
     overfetch: int
+
+
+@dataclass(frozen=True)
+class _Snapshot:
+    """Frozen snapshot of mutable kai.memory state captured before an
+    override is applied. Used by `_restore_overrides` to put the
+    module back in its pre-apply shape regardless of how the caller's
+    loop exits.
+
+    Stored as a frozen dataclass so the snapshot itself cannot be
+    accidentally mutated between capture and restore (the dict
+    contents are copied at capture time so future mutation of
+    `_SPEAKER_WEIGHTS` does not leak through the saved reference).
+    """
+
+    speaker_weights: dict[str, float]
+    overfetch: int
+    config: Config
 
 
 @dataclass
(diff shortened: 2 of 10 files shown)


dcellison:
## Review by Kai

The diff is the PR being reviewed, not yet merged. I have enough context from the diff itself. Let me write the review.

---

## PR Review: feat(memory): per-fact speaker attribution with retrieval-time downweight

Overall this is a well-structured PR with thorough tests, good comments, and a clean separation of the sweep harness from the ranking logic. A few findings below.

---

### Warning

**`_read_time_speaker` discards explicit `speaker` when `confidence` is absent** (`src/kai/memory.py`)

```python
speaker = metadata.get("speaker")
confidence = metadata.get("confidence")
if speaker is not None and confidence is not None:   # both required
    return speaker, confidence

source = metadata.get("source") or ""
# ... falls through to source-based defaults
return _LEGACY_SPEAKER, _LEGACY_CONFIDENCE          # "assistant", 0.5
```

If a row carries `speaker="user"` but no `confidence` key (e.g. a future write path, a partial Mem0 round-trip, or a manual Qdrant edit), the explicit speaker is silently thrown away and the function returns `_LEGACY_SPEAKER` / `_LEGACY_CONFIDENCE` via branch 4. A `user`-attributed row gets demoted to `0.7 * 0.5 = 0.35` instead of `1.0 * <default>`.

The migration and episode branches are immune because their source values match earlier guard clauses, but the extracted path is not. A safer fallback for the "speaker present, confidence absent" case would use the explicit speaker with a source-appropriate confidence default rather than ignoring the speaker entirely. Current production data won't hit this, but the function is fragile for any future write path that sets only one of the two fields.

---

### Suggestion

**`_browse_score` accesses `memory._speaker_weight` across module boundaries** (`src/kai/memory_command.py`)

```python
def _browse_score(fact: MemoryResult) -> float:
    return memory._speaker_weight(fact)
```

Calling a single-underscore private from a sibling module is a convention violation and creates implicit coupling - a future rename of `_speaker_weight` won't produce a `NameError` at import time, only at call time. If this shared formula is intentionally part of the public contract between the two modules, consider lifting `_speaker_weight` to a public name or exporting it explicitly. If not, inline the formula in `_browse_score` (it's one line) to remove the dependency.

---

**`_reset_memory_module` in `test_eval_retrieval.py` omits `_UNKNOWN_SPEAKER_WEIGHT`**

The old fixture reset both `_SOURCE_WEIGHTS` and `_UNKNOWN_SOURCE_WEIGHT` together. The new fixture resets `_SPEAKER_WEIGHTS` but not `_UNKNOWN_SPEAKER_WEIGHT`. In practice `_UNKNOWN_SPEAKER_WEIGHT` is never mutated after module load (it captures the float `0.7` at definition time, not a live dict reference), so no current test leaks through this gap. But since the alias intent ("keeps unknown rows at the assistant weight floor") could be lost if `_SPEAKER_WEIGHTS["assistant"]` is ever tuned in a calibration sweep, adding `mem_mod._UNKNOWN_SPEAKER_WEIGHT = mem_mod._SPEAKER_WEIGHTS["assistant"]` to the fixture would keep the documented invariant holding across tests the same way the old fixture did.

---

**`_apply_override` `_config` derivation footgun is documented but not mitigated** (`src/kai/eval/retrieval.py`, `_apply_override` docstring)

The docstring correctly flags that deriving the new `_config` from `snap.config` (live per-iteration state rather than the entry-time original) would break if a future grid axis mutates a second Config field. The "swap to a `base_config` parameter" suggestion is good but leaves the fix to a future PR. Given that the sweep harness is the only caller and the calibration work referenced in the description will produce a follow-up, this is low priority - but marking the function with a `TODO` or a short assert that only `memory_search_floor` was changed (e.g. comparing other fields against the entry snapshot) would make the footgun fail loudly rather than silently producing drift across grid iterations.

---

No security issues found. Convention compliance (ruff, commenting style, no em-dashes) looks consistent with the project's CLAUDE.md rules. The test coverage for the new paths is thorough and the demote-only invariant test (`test_speaker_weight_demote_only`) is particularly good.

dcellison:
## Review by Kai

## Review: feat(memory): per-fact speaker attribution with retrieval-time downweight

Both findings from the prior review that were flagged as **Warning** and the `_UNKNOWN_SPEAKER_WEIGHT` fixture gap have been addressed. The `_read_time_speaker` independent-field resolution and the fixture reset are both correct. One new finding below.

---

### Suggestion

**Double invocation of `_read_time_speaker` per result in `format_context`** (`src/kai/memory.py`, weighted-sort and hit-payload blocks)

```python
# First call: inside _speaker_weight(r), which calls _read_time_speaker internally
weighted = sorted(
    ((r, _speaker_weight(r)) for r in results),
    key=lambda t: t[0].score * t[1],
    reverse=True,
)

# Second call: explicit, to unpack speaker/confidence for the log payload
for r, w in weighted:
    speaker, confidence = _read_time_speaker(r.metadata)
    payload["hits"].append(...)
```

The "compute the per-row weight ONCE" comment in the block applies accurately to `w` — the weight is computed once and threaded into both `adj` and the sort key. But `_read_time_speaker` is called again to get the component `(speaker, confidence)` pair for logging, which is a second parse of the same metadata dict. It won't produce different results (both calls are pure and idempotent), so this is not a correctness issue; it's a minor inconsistency with the comment's stated intent and adds a redundant parse per result row. The most compact fix is to carry `(speaker, confidence)` out of the sort step:

```python
weighted = sorted(
    ((r, _speaker_weight(r), *_read_time_speaker(r.metadata)) for r in results),
    key=lambda t: t[0].score * t[1],
    reverse=True,
)
for r, w, speaker, confidence in weighted:
    ...
```

Low priority — `_read_time_speaker` is cheap — but worth cleaning up if the loop is ever touched again.

---

Everything else looks correct. The `_read_time_speaker` independence fix resolves the prior warning cleanly; the `_UNKNOWN_SPEAKER_WEIGHT` fixture reset is now wired in; the `_apply_override`/`_restore_overrides` split is well-structured and the test class pins both the apply and restore contracts. No security issues, no convention violations against CLAUDE.md.

dcellison:
## Review by Kai

The prior review's two suggestions were addressed: `_resolved_row` eliminates the double-parse, and `_UNKNOWN_SPEAKER_WEIGHT` is re-bound in the test fixture. One new finding and one minor note below.

---

### Suggestion

**`_resolved_row` in `format_context` duplicates the weight formula from `_speaker_weight`** (`src/kai/memory.py`)

```python
# inside format_context
def _resolved_row(r: MemoryResult) -> tuple[MemoryResult, float, str, float]:
    speaker, confidence = _read_time_speaker(r.metadata)
    weight = _SPEAKER_WEIGHTS.get(speaker, _UNKNOWN_SPEAKER_WEIGHT) * confidence
    return r, weight, speaker, confidence
```

The formula `_SPEAKER_WEIGHTS.get(speaker, _UNKNOWN_SPEAKER_WEIGHT) * confidence` is identical to the body of `_speaker_weight` minus its `_read_time_speaker` call. This is a deliberate tradeoff to avoid the double-parse noted in the previous review, and the computation is simple enough that divergence is unlikely. But if `_speaker_weight` ever gains a clamping or normalization step, `_resolved_row` won't inherit it silently. Worth a comment cross-referencing `_speaker_weight` (analogous to the "Compute the per-row weight ONCE" comment it replaced) so a future editor knows why the formula lives in two places.

---

### Minor

**`run_sweep` docstring misstates where the entry snapshot comes from** (`src/kai/eval/retrieval.py`, `run_sweep` docstring)

> "Instead we capture the entry-time state directly via `_apply_override`'s helper `_Snapshot` constructor"

The code constructs `entry_snap` via a direct `_Snapshot(...)` call, not through `_apply_override`. `_Snapshot` is a module-level dataclass with no ownership relationship to `_apply_override`. The sentence implies tighter coupling than exists and could confuse a reader who goes looking for a `_Snapshot` constructor inside `_apply_override`. A simpler phrasing: "we construct a `_Snapshot` directly before the loop rather than calling `_apply_override(grid[0])` and discarding the first mutation."

---

No bugs, security issues, or convention violations found in the visible portion of the diff. The diff was truncated before the round-trip integration tests (§6.1), so those could not be reviewed here, but the unit coverage visible in `TestReadTimeSpeaker`, `TestSpeakerWeight`, `TestApplyOverride`, and the updated `TestMigrationSourceMetadata` is thorough and the invariant tests are well-constructed.

dcellison:
## Review by Kai

Looking at this diff against the prior review thread, both findings from the third review have been addressed: `_resolved_row` now carries the cross-reference comment to `_speaker_weight`, and `run_sweep`'s docstring correctly says "we construct a `_Snapshot` directly before the loop rather than calling `_apply_override(grid[0])`."

No new bugs, security issues, or convention violations in the visible portion of the diff. A few minor observations:

---

**Note** (`src/kai/memory_command.py`, `_send_facts_list`)

The quality-sort tiebreaker sorts `(browse_score, date)` with `reverse=True`, which gives date descending as the tiebreaker. This is the stated intent and correct. Worth noting that if `f.updated_at` and `f.created_at` are both `None`, the fact sorts to the end of the quality list (since `""` < any ISO date string in descending order). That is reasonable behavior and no action needed; just flagging the implicit semantic for any future change to the fallback.

---

**Note** (`src/kai/eval/retrieval.py`, `_apply_override`)

The `_Snapshot` dataclass is `frozen=True`, which prevents rebinding `snap.speaker_weights`, but the dict it holds is still mutable. In practice the captured dict is never mutated after capture (the restore path calls `clear()` + `update()` on the live module dict, not on the snapshot), so this is fine. The docstring comment "dict contents are copied at capture time" is accurate. Calling this out only because it is a subtle invariant to maintain if `_restore_overrides` is ever refactored.

---

The diff is truncated before the `§6.1` round-trip integration tests and the end of `tests/test_memory.py`. The visible unit coverage (`TestReadTimeSpeaker`, `TestSpeakerWeight`, `TestApplyOverride`, `TestMigrationSourceMetadata`) is thorough, the demote-only invariant test iterates the live production table, and the restore round-trip is pinned at the helper level. No issues with what is visible.

PR #445, opened 2026-05-10
refactor(install): relocate templates to templates/, retire IDENTITY.md symlink (#442)

### LINKED ISSUE #442: Remove home/ from the repo: relocate templates, drop eval probes, retire IDENTITY.md symlink
## Problem

The repo currently tracks files under `home/`, a directory whose name reads as personal user data. This creates two related problems.

**Privacy / category mismatch.** Unix convention says `home/` is personal. Tracking content there blurs the line between personal and shared. The most acute case is `home/evals/*.example.jsonl`, advertised in their headers as "sanitized example fixtures" but living next to gitignored personal probes in the same directory. Anyone reading the repo can reasonably wonder whether those eval probes contain real conversation data or actually-scrubbed content. The directory layout itself invites accidental commits of personal probe files.

**Weak justification for tracking the eval probes.** Audit of the four files:

| File | Tracked consumer | Effective purpose |
|---|---|---|
| `home/evals/extraction-probes.example.jsonl` | `tests/test_eval_extraction.py::test_example_probe_fixture_loads` | Parser self-test (asserts fixture parses, 4 probes, 2/2 category mix) |
| `home/evals/episode-classification-probes.example.jsonl` | `tests/test_eval_extraction.py::test_episode_classification_fixture_loads` | Parser self-test (asserts fixture parses, 12 probes, 8/4 mix) |
| `home/evals/probes.example.jsonl` | None | Personal Layer 1 retrieval input only |
| `home/evals/retrieval-baseline.example.json` | None (one help-string mention in `kai/eval/retrieval.py:1047`) | Personal baseline file only |

The two consumers are circular: they verify the fixture parses against its own loader. They do not measure extractor quality. Equivalent coverage is achievable from inline `tmp_path` fixtures of 2-3 synthetic probes, or from deletion of those tests entirely.

**Stale IDENTITY.md / symlink workaround.** The installer currently bootstraps an `IDENTITY.md` file at `<home_workspace>/IDENTITY.md` and creates a `<home_workspace>/.claude/CLAUDE.md` symlink pointing at it (`../IDENTITY.md`). This was a permission workaround for when `.claude/CLAUDE.md` was tracked-and-shared across operators: editing `IDENTITY.md` via a writable path let inner Claude update its identity without touching the read-only `.claude/` directory. Since `CLAUDE.md` was untracked and gitignored (becoming per-operator, seeded from `.example`), the symlink workaround has no remaining purpose. It survives only as friction: a confusing two-name file, install logic, and tests that exist solely to maintain a redundant indirection.

**Inconsistent template layout.** Tracked seed files are scattered across two roots: `home/.claude/*.example`, `home/config/goose-config.yaml`, three `*.example.yaml` files at the repo root (`services.example.yaml`, `users.example.yaml`, `workspaces.example.yaml`), and `.env.example` at the repo root. The `.example` suffix carries information when a file sits next to its non-example counterpart, but inside a dedicated `templates/` directory the suffix is redundant - the directory name already says "this is a template; the installer copies it into your real config path." Co-locating all templates and dropping the suffix yields one obvious place to look for seed content and one filename per template instead of two.

## Proposed change

Five pieces. After this change, `home/` does not exist in the repo working tree at all, every tracked seed file lives under `templates/` with a non-suffixed name, and `<home_workspace>/.claude/CLAUDE.md` is a regular file copied from `templates/.claude/CLAUDE.md`.

### 1. Drop eval probes from the repo

Untrack all four `home/evals/*.example.{jsonl,json}` files. Personal eval probes belong in the operator's personal workspace under `DATA_DIR/home/<chat_id>/`. New contributors who want to run the eval harness build their own probes against their own conversation history; a docstring on `kai.eval.extraction.load_probes` documents the schema.

If the two parser-self-tests in `tests/test_eval_extraction.py` are worth keeping, replace them with inline `tmp_path` fixtures of 2-3 obviously-synthetic probes (Lorem-ipsum-shaped content, not derived from any real conversation). Otherwise delete them; the integration tests that actually run extraction exercise the loader transitively.

### 2. Move installer templates to `templates/` and drop the `.example` suffix

Relocate the four tracked template files and rename them in the move:

```
home/.claude/CLAUDE.md.example       -> templates/.claude/CLAUDE.md
home/.claude/MEMORY.md.example       -> templates/.claude/MEMORY.md
home/.claude/PREFERENCES.md.example  -> templates/.claude/PREFERENCES.md
home/config/goose-config.yaml        -> templates/config/goose-config.yaml
```

The `templates/` directory name is unambiguous: a tracked directory of seed content the installer copies into per-user destinations. Inside that directory, the `.example` suffix is redundant and is removed from the three `.claude/*` files. `goose-config.yaml` already had no suffix and keeps its name.

### 3. Move root config examples to `templates/` and drop the `.example` suffix

Move the four tracked config examples from the repo root into `templates/` and rename them to match their live counterparts:

```
services.example.yaml    -> templates/services.yaml
users.example.yaml       -> templates/users.yaml
workspaces.example.yaml  -> templates/workspaces.yaml
.env.example             -> templates/.env
```

These renames collide with the existing `.gitignore` rules at lines 2, 6, 9, 10 (`.env`, `services.yaml`, `users.yaml`, `workspaces.yaml`), which are unanchored and therefore match anywhere in the tree. Without remediation, `templates/users.yaml` and `templates/.env` would be silently ignored. Anchor the four root patterns so they only ignore at the repo root:

```
.env             -> /.env
services.yaml    -> /services.yaml
users.yaml       -> /users.yaml
workspaces.yaml  -> /workspaces.yaml
```

`/etc/kai/services.yaml`, `/etc/kai/users.yaml`, and `/etc/kai/workspaces.yaml` are outside the working tree and unaffected; the change is purely about anchoring the in-repo gitignore patterns.

Documentation and code references that mention the old paths need to be updated:
- `README.md:46`: `workspaces.example.yaml` -> `templates/workspaces.yaml`.
- `README.md:71`: `users.example.yaml` -> `templates/users.yaml`.
- `README.md:191`: `cp .env.example .env` -> `cp templates/.env .env`.
- `README.md:366` and `:369` (the directory tree): the `.env`/`workspaces.example.yaml` lines are removed from their current positions; the tree shows `templates/` with its contents instead.
- `src/kai/config.py:305`: docstring reference to `.env.example` -> `templates/.env`.
- `tests/test_config.py:1503`: docstring/comment reference to `.env.example` -> `templates/.env`.

### 4. Retire IDENTITY.md and the symlink convention

Delete the IDENTITY.md / symlink mechanism wholesale. After this change:

- The installer copies `templates/.claude/CLAUDE.md` to `<home_workspace>/.claude/CLAUDE.md` as a regular file. No symlink, no intermediate `IDENTITY.md`.
- `templates/.claude/CLAUDE.md` is a regular file in the repo (not a symlink to anything).
- The "About This File" prose in the relocated template (currently describing the symlink and the IDENTITY.md edit-via-home convention at lines 5, 7, 9 of the original) is rewritten to describe the simple model: "this is a template; the installer copies it to your `.claude/CLAUDE.md` as a regular file; edit your local copy."

This removes the `_bootstrap_identity_and_symlink`-shaped routine in `install.py` (around lines 2647-2750) and its caller in `_apply_source`, plus the related dry-run output. It also removes the symlink-creation step in the install flow and the symlink-target reconciliation logic.

### 5. Rewrite the `.gitignore` home/ stanza and anchor the root config patterns

The current 14-line `home/` stanza exists to allowlist tracked content under that directory. With nothing tracked there and the directory gone from the repo entirely, the stanza is removed wholesale.

```
!home/.claude
home/.claude/*
!home/.claude/CLAUDE.md.example
!home/.claude/MEMORY.md.example
!home/.claude/PREFERENCES.md.example
!home/config
home/config/*
!home/config/goose-config.yaml
!home/evals
home/evals/*
!home/evals/retrieval-baseline.example.json
!home/evals/probes.example.jsonl
!home/evals/extraction-probes.example.jsonl
!home/evals/episode-classification-probes.example.jsonl
```

becomes nothing. The `# CLAUDE.md and IDENTITY.md are per-operator; bootstrapped from the` comment at line 38 also goes (replace with a one-liner about CLAUDE.md if needed).

In the same edit, anchor the four root config patterns (`.env`, `services.yaml`, `users.yaml`, `workspaces.yaml`) to a leading slash so they only match at the repo root, allowing `templates/services.yaml` etc. to be tracked without per-file allowlist entries.

Deliberate consequence on the home/ side: if buggy code ever recreates a `home/` directory under `PROJECT_ROOT` (the recent DATA_DIR test-leak class is one example), it shows up loudly in `git status` instead of being silently swallowed. A loud failure is the right behavior.

## Scope

- Move the four template files from `home/.claude/` and `home/config/` to `templates/.claude/` and `templates/config/`. Drop the `.example` suffix on `CLAUDE.md`, `MEMORY.md`, `PREFERENCES.md`. `goose-config.yaml` keeps its name.
- Move the four tracked config examples from the repo root to `templates/` and drop the `.example` infix/suffix: `services.example.yaml` -> `templates/services.yaml`, `users.example.yaml` -> `templates/users.yaml`, `workspaces.example.yaml` -> `templates/workspaces.yaml`, `.env.example` -> `templates/.env`.
- Anchor the four root `.gitignore` patterns (`.env`, `services.yaml`, `users.yaml`, `workspaces.yaml`) with a leading slash so they only ignore at the repo root.
- Untrack the four `home/evals/*` files. No replacement directory in the repo.
- Remove the `home/` directory entirely from the repo working tree.
- Rewrite the "About This File" prose (and any other symlink-related sentences) in the relocated `templates/.claude/CLAUDE.md`.
- Delete the IDENTITY.md / symlink installer logic in `src/kai/install.py`. Includes:
  - The bootstrap routine around lines 2647-2750 (`_bootstrap_identity_and_symlink` or equivalently named).
  - Its caller in `_apply_source` around line 2824-2825.
  - The comment at `install.py:1294` referencing the symlink.
  - The `.gitignore` exclusion comment at line 38 mentioning IDENTITY.md.
  - Any IDENTITY.md-aware path handling elsewhere.
- Delete or rewrite IDENTITY.md-specific tests in `tests/test_install.py` (lines 3872-3914 around `test_apply_source_actual`, `test_copies_identity_md_with_svc_ownership` ~3962, `test_dry_run_includes_identity_md` ~3990, and the no-IDENTITY warning tests ~4002-4036).
- Update non-installer IDENTITY.md references to read CLAUDE.md instead: `src/kai/webhook.py:1536`, `src/kai/webhook.py:1820`, `tests/test_webhook_api.py:2535`.
- Update `src/kai/install.py` path references at lines 73, 2039, 2071, 2077, 2090-2091, 2280, 2300, 2652, 2664, 2682, 2711, 2755-2765, 2812, 3000, 3008 to point at `templates/` and to use the non-suffixed filenames where applicable (`templates/.claude/CLAUDE.md`, `templates/.claude/MEMORY.md`, `templates/.claude/PREFERENCES.md`, `templates/config/goose-config.yaml`).
- Update `src/kai/backend.py:477` and `:532` to seed from `templates/.claude/MEMORY.md` and `templates/.claude/PREFERENCES.md` respectively.
- Update `src/kai/eval/behavioral.py:534` comment and `src/kai/eval/retrieval.py:1047` help string.
- Either rewrite the two parser-self-tests in `tests/test_eval_extraction.py` (around lines 92 and 110) to use inline synthetic data, or delete them and document the loader schema on `load_probes` instead.
- Update `tests/test_install.py` references including the `.gitignore` allowlist test at lines 3487, 3506-3508 (deleted entirely once the negative-include rules go away); test setup that creates `home/.claude/` and `home/config/` directories at the source path now points at `templates/` with the renamed filenames.
- Update `tests/test_backend.py` references that mention the old `MEMORY.md.example` / `PREFERENCES.md.example` template paths.
- Update comment-level references in `tests/test_eval_behavioral.py` (lines 34, 621, 664) and `tests/test_triage.py:1274`.
- Update `README.md` references at lines 46, 71, 191, 366, and 369 to use the new `templates/` paths and non-suffixed filenames (including `cp templates/.env .env` in place of `cp .env.example .env`).
- Update `src/kai/config.py:305` and `tests/test_config.py:1503` to reference `templates/.env` instead of `.env.example`.
- Delete the entire `home/` stanza in `.gitignore`. Anchor the four root patterns described above.

## Out of scope

- The DATA_DIR test-leak bug that lets buggy tests create directories under `PROJECT_ROOT`. That class of leak is what would put a `home/` directory back into the repo even after this change lands. Separate issue; this change relies on that being fixed (or surfaces its absence loudly via `git status`).
- Auditing the actual content of the four eval probe files for sensitive material. The relocation removes them from new clones; if any of their committed content was unsafe, it remains in the public git history and a separate `git filter-repo` rewrite is required. That belongs in its own issue.
- Per-user `home_workspace` resolution in `kai.backend.resolve_home_workspace`. That logic is unchanged.
- Any rewrite of the universal-baseline content inside the relocated `CLAUDE.md` template beyond removing the symlink prose. Stale architecture descriptions in the body (e.g., pre-Track-1 vector-store wording) are out of scope here.

## Acceptance

- `home/` does not exist as a directory under `PROJECT_ROOT` in the repo working tree.
- `git ls-files home/` returns nothing.
- `git ls-files templates/` lists exactly: `.claude/CLAUDE.md`, `.claude/MEMORY.md`, `.claude/PREFERENCES.md`, `.env`, `config/goose-config.yaml`, `services.yaml`, `users.yaml`, `workspaces.yaml`. The four originally-suffixed `.claude` and `goose` files are byte-identical to their pre-move counterparts, with the one exception of `.claude/CLAUDE.md` whose "About This File" section is rewritten. The three YAML files and `.env` are byte-identical to the pre-move `*.example.yaml` and `.env.example` files.
- `templates/.claude/CLAUDE.md` is a regular file (`os.path.islink` returns `False`), not a symlink.
- `git ls-files | grep -E '^(services|users|workspaces)\.example\.yaml$|^\.env\.example$'` returns nothing (no root-level example configs remain tracked).
- `git check-ignore templates/.env` returns no match (the `.env` template is not gitignored).
- The `home/` stanza in `.gitignore` is gone (zero lines mentioning `home/`).
- The four root config patterns in `.gitignore` are anchored: `/.env`, `/services.yaml`, `/users.yaml`, `/workspaces.yaml`.
- `git check-ignore templates/users.yaml templates/services.yaml templates/workspaces.yaml` returns no matches (the templates are not gitignored).
- `grep -rn IDENTITY.md src/ tests/ scripts/ templates/ .gitignore` returns nothing.
- `pytest -q` passes with no references to `home/evals/*.example.*`, `*.example.yaml` at the root, `.env.example`, or IDENTITY.md remaining in the test tree.
- `python -m kai install --dry-run` output mentions `templates/.claude` and `templates/config`, never `home/.claude`, `home/config`, or `IDENTITY.md`, and never the `.example` suffix on any template path.
- A fresh install against a clean machine creates `<home_workspace>/.claude/CLAUDE.md` as a regular file (`os.path.islink` returns `False`) whose contents match `templates/.claude/CLAUDE.md`. No `IDENTITY.md` file is created anywhere.
- After the change is merged, `git status` in a clean clone is clean, with no untracked `home/` directory and no untracked `services.example.yaml` / `users.example.yaml` / `workspaces.example.yaml` / `.env.example` at the root.

## Notes

The eval-probe history audit is the higher-stakes follow-up. If the committed `.example.*` files contain real conversation fragments rather than scrubbed content, they remain in the public git history after this change; only new clones stop seeing them at the working-tree level. A history rewrite is destructive and coordination-heavy and should be its own issue.

The DATA_DIR test-leak class is the operational dependency: until tests stop writing into `PROJECT_ROOT / "home"`, `PROJECT_ROOT / "memory"`, and `PROJECT_ROOT / "files"` on every run, a developer running the test suite will see those directories reappear and have to delete them. Filing that as its own issue alongside this one is the expected sequence; the leak fix can land in either order, but operationally it is more pleasant to land it first.

dcellison:
## Summary

Closes #442. Relocates tracked seed files from `home/.claude/`, `home/config/`, and the four root `*.example.{yaml,env}` into a single `templates/` tree with non-suffixed names, and retires the `home/IDENTITY.md` plus `home/.claude/CLAUDE.md` symlink layout in favor of a single regular `CLAUDE.md` file at the install destination.

- **Source-only change.** The runtime install layout (`<install>/home/.claude/`, `<install>/home/config/`, per-user `DATA_DIR/home/<chat_id>/`) is unchanged. Only `PROJECT_ROOT/"home"/...` source pointers move to `PROJECT_ROOT/"templates"/...`.
- **Migration in place.** Existing operators with `home/IDENTITY.md` plus a `CLAUDE.md` symlink upgrade via a new `_migrate_identity_to_claude_md` helper covering the full 2x3 state space (six rows; one test per row). Operator-customized identity content is preserved; the helper reads content into memory before unlinking either path.
- **Drops dead scaffolding.** The `_copy_tree` symlink branch, the `legacy_src_tree` MEMORY.md migration branch in `_apply_migrate`, and the `home/.claude/history` migration block (all pre-DATA_DIR) are removed. Eval probe fixtures leave the repo entirely; the two parser-self-tests use inline `tmp_path` fixtures with synthetic Lorem-ipsum content.
- **Anchored `.gitignore` patterns.** The four root patterns (`.env`, `services.yaml`, `users.yaml`, `workspaces.yaml`) get a leading slash so `templates/services.yaml` etc. are not silently ignored. The 21-line `home/` allowlist stanza is removed wholesale so any future test-leak under `PROJECT_ROOT` shows up loudly in `git status`.

## Test plan

- [x] `make check` passes (ruff check + format).
- [x] `pytest -q` passes (2816 tests, 1 skipped).
- [x] `git ls-files templates/` lists exactly the 8 expected files; `git ls-files home/` returns nothing.
- [x] `templates/.claude/CLAUDE.md` is a regular file (not a symlink).
- [x] `git check-ignore` does not match `templates/.env`, `templates/services.yaml`, `templates/users.yaml`, or `templates/workspaces.yaml`.
- [x] `grep -rn 'IDENTITY\.md'` outside `src/kai/install.py` and `tests/test_install.py` returns nothing.
- [x] `grep -rn 'PROJECT_ROOT.*"home"'` returns nothing.
- [x] `grep -rn '\.md\.example'` returns nothing.
- [ ] Reviewer to spot-check `_migrate_identity_to_claude_md` test coverage against the §3.3 migration table (6 rows + dry-run).
- [ ] After merge: confirm next install on the production machine emits the row-1 migration log line ("Migrated ... content into regular file ...; removed ...IDENTITY.md") exactly once and the destination CLAUDE.md content matches the operator's pre-merge IDENTITY.md.


dcellison:
## Review by Kai

This is a clean, well-motivated refactor. Here are the findings ranked by severity:

## Warning

**Incorrect comment in Row 4 of `_migrate_identity_to_claude_md`** (`src/kai/install.py`, near the `elif claude_md_is_symlink and not identity_exists` branch):

> "Without this branch the seed step would skip (it checks `not exists()`, which is **False** for a broken symlink)"

This is backwards. `Path.exists()` follows symlinks and returns **False** for a broken symlink, so `not exists()` is **True** - meaning the seed step would *not* skip; it would proceed and `shutil.copy2` would fail trying to dereference the broken symlink target. The migration branch is still necessary and correct, but the stated reason is wrong. A future maintainer debugging a partial-failure might trust the comment over their own reading of the stdlib.

## Suggestions

**Dry-run prediction gap (Row 4 non-broken subcase)** (`_apply_source` dry-run block):

```python
identity_src_pre = install_path / "home" / "IDENTITY.md"
identity_would_move = (
    identity_src_pre.is_file() and not identity_src_pre.is_symlink() and not claude_md_dst.exists()
)
if claude_md_src.is_file() and not claude_md_dst.exists() and not identity_would_move:
    print(f"[DRY RUN] Would seed {claude_md_src} -> {claude_md_dst}")
```

If `CLAUDE.md` is a symlink to a *valid but non-IDENTITY.md target* (no IDENTITY.md present), `claude_md_dst.exists()` returns `True` (symlink not broken), so the dry-run skips the seed prediction. But the live path removes the symlink (Row 4) and then seeds. This is an exotic pre-state, but the dry-run and live paths diverge for it. Low real-world risk, worth a comment acknowledging the gap.

**Silent symlink dereference in `_copy_tree`** (`src/kai/install.py`, removed symlink branch):

Removing the `is_symlink()` branch from `_copy_tree` means any future symlink accidentally introduced under `templates/` would be silently dereferenced by `shutil.copy2` rather than recreated. `_copy_tree` is also called for `src/`, so the behavioral change is broader than just `templates/`. This is intentional here (the source no longer contains symlinks), but the generic function now has a latent footgun. A one-line comment noting the removal is deliberate would prevent someone from "restoring" it as an apparent bug.

**Truncated diff caveat:** The migration test class `TestMigrateIdentityToClaudeMd` is cut off at the end of the diff, so I can't verify all 6 rows + dry-run are covered as the PR description requests. The reviewer checklist item ("Reviewer to spot-check against the §3.3 migration table") remains open.

---

The core logic - the 2x3 state space in `_migrate_identity_to_claude_md`, the template rename/path updates, the `.gitignore` anchor fix, and the removal of dead `legacy_src_tree` and history-migration branches - is all correct and well-commented. No security issues or data-loss paths found.

dcellison:
## Review by Kai

## Review

All three prior warnings/suggestions have been addressed: the Row 4 comment now correctly states `Path.exists()` returns `False` for a broken symlink; the dry-run prediction logic correctly accounts for the Row 4 valid-non-IDENTITY subcase; and the `_copy_tree` comment documents the intentional symlink-branch removal.

New findings:

---

### Warning

**`TestMigrateIdentityToClaudeMd` still truncated**

The diff ends mid-way through `TestApplySource` and the `TestMigrateIdentityToClaudeMd` class is not visible. The PR checklist item ("Reviewer to spot-check migration test coverage against the §3.3 migration table (6 rows + dry-run)") cannot be satisfied from the available diff. This is the same structural gap as the prior review.

---

### Suggestion

**`.gitignore` unanchored `CLAUDE.md` pattern** (`.gitignore`, pre-existing line)

The pattern `CLAUDE.md` in `.gitignore` is unanchored and matches `templates/.claude/CLAUDE.md` anywhere in the tree. The file is currently tracked via `git mv` from `home/.claude/CLAUDE.md.example`, so this PR is not broken. However, if the file is ever accidentally removed from the index (`git rm --cached`) it would require `git add -f` to re-add, with no obvious indication why a plain `git add` fails. The four root-level patterns were anchored in this same PR (`/.env`, `/services.yaml`, etc.); anchoring this one too (`/CLAUDE.md`) would be consistent and eliminates the latent footgun.

**Row 1 of `_migrate_identity_to_claude_md` loses mode bits** (`src/kai/install.py`, `_migrate_identity_to_claude_md`)

Row 1 uses `write_bytes()` to create the new CLAUDE.md, which produces mode `0o644` (minus umask). Row 2 uses `Path.replace()`, which preserves IDENTITY.md's mode. If IDENTITY.md was installed with non-default permissions, Row 1 silently drops them. Using `shutil.copy2(identity_dst, claude_md_dst)` before the `unlink()` calls would preserve mode and ownership in one step, consistent with how the templates are seeded elsewhere. Low real-world risk since default install modes are uniform, but the inconsistency between rows is worth noting.

---

The core logic, the six-row state machine, template path updates, `.gitignore` anchoring, and dead-code removal are all correct. No security issues or data-loss paths found in the visible portion of the diff.

dcellison:
## Review by Kai

This iteration resolves all three items from the last review (Row 4 comment, dry-run prediction, `_copy_tree` annotation, `.gitignore` `/CLAUDE.md` anchor, and Row 1 mode-bits via `shutil.copy2`). Core logic is correct. Two new findings below.

---

## Suggestion

**Misleading variable name in the dry-run seed prediction** (`src/kai/install.py`, dry-run block of `_apply_source`)

```python
identity_src_pre = install_path / "home" / "IDENTITY.md"
```

The variable is named `identity_src_pre` but points at `install_path` (the install destination), not the source tree. Every other path in this block that refers to the install side uses `_dst` suffix (e.g. `claude_md_dst`). A reader cross-referencing the six-row table has to pause to confirm this isn't `PROJECT_ROOT / "home" / "IDENTITY.md"`. Renaming to `identity_dst_pre` removes the ambiguity.

---

**Row 1 fires for any symlink, docstring promises only `../IDENTITY.md`** (`src/kai/install.py`, `_migrate_identity_to_claude_md`)

The docstring for Row 1 says: *"IDENTITY.md regular file + CLAUDE.md symlink **to ../IDENTITY.md**"*. The guard is:

```python
if identity_exists and claude_md_is_symlink:
```

This fires for any symlink target. An operator who had, say, `CLAUDE.md -> /etc/kai/shared-identity.md` (exotic but possible) and also happened to have `IDENTITY.md` present would silently lose the symlink and have `IDENTITY.md`'s content written in its place. The code cannot distinguish these cases without reading the symlink target, which is probably not worth adding. But one sentence acknowledging "this catches all symlink targets, not only `../IDENTITY.md`; for the purposes of this migration all symlinks are treated as the legacy layout" would close the gap between the docstring and the code, and prevent a future reader from adding a target check as an "obvious improvement."

---

## Persistent open item

`TestMigrateIdentityToClaudeMd` is still cut off by the diff truncation. The PR checklist item ("Reviewer to spot-check against the §3.3 migration table — 6 rows + dry-run") cannot be satisfied from the available diff for the third consecutive round. This is a diff-size constraint rather than a code problem, but the checklist box remains unchecked until it can be read directly.

---

No new critical issues, security concerns, or data-loss paths found.

dcellison:
## Review by Kai

This is a clean, complete refactor. All three previous-round findings (variable naming, Row 1 docstring, `_copy_tree` annotation) are resolved. Here are the new findings:

## Persistent open item

**`TestMigrateIdentityToClaudeMd` still truncated** (`tests/test_install.py`, end of diff)

The diff ends before `TestMigrateIdentityToClaudeMd` is visible for the fourth consecutive round. The PR checklist item ("Reviewer to spot-check against the §3.3 migration table -- 6 rows + dry-run") cannot be satisfied from the available diff. This is a diff-size constraint, not a code problem, but the box remains unchecked.

## Suggestion

**`test_dry_run_mentions_template_paths` assertion is path-sensitive** (`tests/test_install.py`, `TestApplySource`)

```python
assert "home/.claude" not in output.replace("install/home/.claude", "")
```

This removes the literal substring `"install/home/.claude"` to allow the destination path while blocking any stale source-path reference. It works correctly for the tested case, but it depends on the tmp_path-relative install directory being named exactly `"install"`. If a future refactor renames the directory variable (e.g., `install_dir`) or the tmp_path tree happens to include `"home/.claude"` in a prefix component (unlikely but possible on some CI environments), the assertion silently turns into an unconditional pass or fail. A more robust form would check the source component directly: `assert "templates/.claude" in output` (already present) and separately `assert str(ws_claude_src) + " ->" in output` or similar. Low real-world risk given the test structure, but the double-negative replace pattern is the weakest spot in the new test additions.

---

The six-row state machine, the dry-run prediction logic, the `.gitignore` anchoring, the dead-code removal (history migration, `legacy_src_tree` branch, `_bootstrap_home_identity`), and all template path updates are correct. No security issues or data-loss paths found in the visible portion of the diff.

dcellison:
## Review by Kai

The explore agent read the pre-PR state of the file (the PR isn't merged locally). The diff already updates the `_HOME_CLAUDE_EXCLUDES` comment block correctly.

---

## Review

This iteration is clean. The core logic - the six-row state machine, dry-run prediction, `.gitignore` anchoring, template tree rename, and dead-code removal - is all correct. No security issues, no data-loss paths.

### Suggestion

**Stale `"history"` and `"MEMORY.md"` entries in `_HOME_CLAUDE_EXCLUDES`** (`src/kai/install.py`, line ~86)

The actual set is `{"history", "MEMORY.md", "CLAUDE.md", "skills", "__pycache__"}`. Both `history` and `MEMORY.md` are described in the comment as legacy exclusions retained because "stale files may linger at the source after migration." Now that `templates/.claude/` has no legacy stale paths - the `history/` subdirectory never existed there and the `MEMORY.md` template is tracked but intentionally seeded via `_apply_migrate` rather than `_copy_tree` - the comment's explanation is stale. It correctly explains *why* CLAUDE.md is excluded (operator-customized file protection on reinstall), but the rationale for keeping MEMORY.md in the set should be updated to say it is seeded per-user via `_apply_migrate`/`ensure_user_memory` from `PROJECT_ROOT/templates/.claude/MEMORY.md` directly, not copied into `install/home/.claude/`. The existing comment says "stale files may linger at the source after migration" which no longer applies to the templates tree. Low risk; no behavior change needed.

### Persistent open item (not re-raising, noting for completeness)

`TestMigrateIdentityToClaudeMd` is still beyond the diff boundary for the fourth consecutive round. The checklist box remains open as a diff-size constraint.

---

Everything visible is correct. The dry-run prediction traces accurately through all six rows, the `test_dry_run_mentions_template_paths` assertions are now robust (direct path checks, no double-negative replace), and `test_seeds_all_users_from_template` correctly reflects the removed `legacy_src_tree` branch. Ready to merge pending the test-class visibility gap.

dcellison:
## Review by Kai

## Review

This iteration is clean. All findings from prior rounds are resolved: the `_HOME_CLAUDE_EXCLUDES` stale-comment issue, the `history`/`MEMORY.md` rationale update, and the test path updates throughout `test_backend.py` and `test_install.py`. No security issues, no data-loss paths.

### New findings

None.

The core logic is all correct in what is visible:

- The six-row state machine in `_migrate_identity_to_claude_md` is correctly guarded (`is_symlink()` paired with `not is_symlink()` to avoid conflating symlinks with regular files in `identity_exists` and `claude_md_is_regular`).
- Partial-failure recovery traces cleanly through every row to a correct re-run state.
- The dry-run seed prediction in `_apply_source` correctly evaluates `post_migration_claude_md_exists` for all six rows.
- The `_copy_tree` symlink-removal and `_HOME_CLAUDE_EXCLUDES` comments are accurate and sufficient.
- `.gitignore` anchoring is consistent (`/.env`, `/services.yaml`, `/users.yaml`, `/workspaces.yaml`, `/CLAUDE.md`).
- Dead code (history migration block, `legacy_src_tree` branch, `_bootstrap_home_identity`) is cleanly removed with matching test deletions.
- `test_dry_run_mentions_template_paths` uses direct path-string checks rather than the prior double-negative replace pattern.

### Persistent open item

`TestMigrateIdentityToClaudeMd` is still beyond the diff boundary for the fifth consecutive round. The PR checklist item ("Reviewer to spot-check against the §3.3 migration table - 6 rows + dry-run") remains open as a diff-size constraint. Everything else is ready to merge.

dcellison:
## Review by Kai

## Review

All prior-round findings are resolved. The visible diff is clean. Two structural observations:

### Persistent open item

`TestMigrateIdentityToClaudeMd` is still beyond the diff boundary for the sixth consecutive round. The PR checklist item ("Reviewer to spot-check against the §3.3 migration table - 6 rows + dry-run") remains open as a diff-size constraint. `test_actual_copies_templates_and_seeds_claude_md` is also cut off mid-assertion (the diff ends at `# shutil.copy2 called twice: pyproject.toml an`).

### Suggestion

**`post_migration_claude_md_exists` is correct but over-broad** (`src/kai/install.py`, dry-run block of `_apply_source`)

```python
post_migration_claude_md_exists = claude_pre_is_regular or (identity_pre_exists and not claude_pre_is_regular)
```

The second operand simplifies to `identity_pre_exists and not claude_pre_is_regular`, but since the `or` short-circuits when `claude_pre_is_regular` is True, the second operand is only evaluated when `claude_pre_is_regular` is False - at which point `not claude_pre_is_regular` is always True and the whole expression reduces to `claude_pre_is_regular or identity_pre_exists`. The current form is not wrong but the double negation adds cognitive load for a reader tracing rows. A single comment line cross-referencing the six-row table ("True for rows 1-3 and 5, False for rows 4 and 6") would close the gap without touching the logic.

---

Everything visible is correct: the six-row state machine, the `Path.replace` atomicity for rows 1-2, the dry-run seed prediction tracing all six rows accurately (including the exotic Row 4 valid-non-IDENTITY symlink subcase), the `.gitignore` anchoring, the `_HOME_CLAUDE_EXCLUDES` per-entry rationale, dead-code removal, and template path updates throughout. No security issues, no data-loss paths.

dcellison:
## Review by Kai

The Explore agent read the pre-PR state of the file - the PR is not merged locally. My review is based on the diff.

---

## Review

The core logic is correct, the prior-round findings are all resolved, and no new critical issues or security problems are visible. Two observations:

### Persistent open item

**`TestMigrateIdentityToClaudeMd` still truncated** - the diff ends mid-way through `TestApplySource` and the migration test class is never visible. This is the seventh consecutive round with this gap. The PR checklist item ("Reviewer to spot-check against the §3.3 migration table - 6 rows + dry-run") remains open as a structural diff-size constraint, not a code problem.

### Suggestion

**Row 4 subcase comment slightly imprecise** (`src/kai/install.py`, Row 4 branch of `_migrate_identity_to_claude_md`):

The comment for the broken-target subcase says:

> `shutil.copy2 would write through the symlink to the missing target path (or fail outright if the parent dir does not exist)`

The second parenthetical is the common case on a real install - `../IDENTITY.md`'s parent dir (`home/`) always exists, so `shutil.copy2` would succeed and write to `home/IDENTITY.md` rather than failing. The "fail outright" framing undersells the actual risk (silent write to the wrong path). Not wrong, but "write to the symlink target path rather than the intended `home/.claude/CLAUDE.md` destination" is more accurate. Low-priority; the conclusion (unlink first) is correct either way.

---

Everything visible is correct: the six-row state machine guards (`is_symlink()` paired with `not is_symlink()` to distinguish regular files from symlinks), `Path.replace` atomicity for rows 1-2, the `post_migration_claude_md_exists` simplification and cross-reference comment, the `test_dry_run_mentions_template_paths` direct path assertions, `.gitignore` anchoring, `_HOME_CLAUDE_EXCLUDES` per-entry rationale, and template path updates throughout. No security issues or data-loss paths found.

dcellison:
## Review by Kai

## Review

All prior-round findings are resolved. One new issue in the live (non-dry-run) path of `_apply_source`.

---

### Warning

**CLAUDE.md ownership clobbered by `_set_ownership` for migration rows 1 and 2** (`src/kai/install.py`, `_apply_source`)

The execution order in the live path is:

1. `_migrate_identity_to_claude_md(...)` - rows 1 and 2 rename IDENTITY.md to CLAUDE.md and call `os.chown(claude_md_dst, svc_uid, svc_gid)`.
2. `_copy_tree(ws_claude_src, ws_claude_dst, _HOME_CLAUDE_EXCLUDES)` - CLAUDE.md excluded, not touched.
3. `_set_ownership(ws_claude_dst.parent, 0, 0, recursive=True)` - sets root:root recursively on `home/`, which includes `home/.claude/CLAUDE.md`.
4. Seed step: `if ... and not claude_md_dst.exists()` - False for rows 1 and 2, so the `os.chown(claude_md_dst, svc_uid, svc_gid)` inside it never runs.

Result: on an existing install hitting row 1 or 2, CLAUDE.md ends up root-owned. The file comment is explicit that the intent is `svc_uid` ownership so inner Claude can edit it from Telegram; root ownership breaks that.

The old `_bootstrap_home_identity` avoided this because it ran *after* the `_set_ownership` call and re-chowned the identity surface. The new migration runs *before* it and has its chown undone.

Minimal fix - after the seed step, unconditionally reconcile CLAUDE.md ownership:

```python
# Ensure CLAUDE.md is svc_uid-owned regardless of which migration row ran.
# _set_ownership above sets root:root on home/ recursively and would
# clobber the chown done inside _migrate_identity_to_claude_md.
if claude_md_dst.is_file() and not claude_md_dst.is_symlink():
    os.chown(claude_md_dst, svc_uid, svc_gid)
```

---

### Persistent open item

`TestMigrateIdentityToClaudeMd` and the tail of `test_actual_copies_templates_and_seeds_claude_md` are still beyond the diff boundary for the ninth consecutive round. The PR checklist item ("Reviewer to spot-check against the §3.3 migration table - 6 rows + dry-run") remains open as a diff-size constraint.

---

Everything else is correct: the six-row state machine guards, `Path.replace` atomicity for rows 1-2, `post_migration_claude_md_exists` cross-reference comment, `test_dry_run_mentions_template_paths` direct path assertions, `.gitignore` anchoring, `_HOME_CLAUDE_EXCLUDES` per-entry rationale, dead-code removal, and template path updates throughout.

dcellison:
## Review by Kai

Looking at the diff with the context of nine prior rounds of review, all previously raised findings appear resolved. Here is the assessment of the current state:

## Clean

**Core logic** - The six-row state machine in `_migrate_identity_to_claude_md`, the dry-run prediction (`post_migration_claude_md_exists = claude_pre_is_regular or identity_pre_exists`), the final unconditional ownership reconciliation after `_set_ownership`, the `.gitignore` anchoring, and the `_HOME_CLAUDE_EXCLUDES` per-entry comments are all correct.

**Template path updates** - Every source reference updated consistently across `install.py`, `backend.py`, `config.py`, `eval/retrieval.py`, `webhook.py`, and the test files. No stale `home/.claude/*.example` references visible.

**Dead code removal** - History migration block, `legacy_src_tree` branch, `_bootstrap_home_identity`, and the symlink branch in `_copy_tree` are cleanly excised with matching test deletions (`test_copies_history`, `test_history_skips_existing`, `test_copies_memory_to_primary_chat_dir`, `test_preserves_symlinks`).

**Test quality** - `test_dry_run_mentions_template_paths` now uses direct path-string assertions rather than the prior double-negative replace pattern. `test_seeds_all_users_from_template` correctly exercises the removed `legacy_src_tree` branch (both primary and secondary now get the template). The inline `tmp_path` fixtures in `test_eval_extraction.py` correctly replicate the deleted `.example.jsonl` file schemas.

**Security** - No injection, no user-controlled paths feeding into dangerous operations, no unintended data exposure.

## Persistent open item (diff-size constraint, not a code problem)

`TestMigrateIdentityToClaudeMd` is still beyond the diff boundary - this is the ninth consecutive round. The PR checklist item ("Reviewer to spot-check against the §3.3 migration table - 6 rows + dry-run") cannot be satisfied from the available diff. Everything else is ready to merge.

PR #446, opened 2026-05-10
fix(install): exclude PREFERENCES.md from global copy, clean up strays

dcellison:
## Summary

Follow-up to #442/PR #445. System check on the operator's machine after the merge install surfaced a fresh `/opt/kai/home/.claude/PREFERENCES.md` (root:wheel) that should not exist. Per #400/#401 the per-user PREFERENCES.md lives at `DATA_DIR/preferences/<chat_id>/PREFERENCES.md`; the install path is the same global-layout artifact MEMORY.md had been excluded against since #347.

- Add `"PREFERENCES.md"` to `_HOME_CLAUDE_EXCLUDES` with a parallel rationale paragraph mirroring the MEMORY.md entry.
- One-shot cleanup in `_apply_source` deletes any stray `install/home/.claude/PREFERENCES.md` left by the brief #442-without-this-fix window. Unconditional delete since nothing reads the path.
- Dry-run preview line matches the live cleanup.

## Why missed in #442

Three reinforcing causes, all repaired in this PR:

1. `_HOME_CLAUDE_EXCLUDES` was preserved "unchanged in shape" through spec v1->v4. The exclude set was treated as untouchable; no agent re-derived membership given the rename.
2. The round-5 review prompted a rewrite of the per-entry rationale that articulated EXACTLY why MEMORY.md is excluded ("would resurrect the pre-#347 global layout and shadow per-user reads"). The same reasoning applies to PREFERENCES.md (added by #400/#401 in the same per-user-data wave). The author of the rationale missed the symmetry.
3. The acceptance grep `grep -rn '\.md\.example'` checked for surviving old-suffix references, not for newly-suffix-stripped collisions at the install destination. The test direction was inverted.

## Test plan

- [x] `make check` passes.
- [x] `pytest -q` passes (2821 tests, 1 skipped; +4 from this PR).
- [x] `TestPreferencesTemplateExcludedFromGlobalCopy` replaces the prior class that pinned the inverse-of-correct contract.
- [x] `test_strips_stray_global_preferences_md` integration test exercises the cleanup against the exact pre-state observed on the operator's machine (root-owned 1026-byte file at `install/home/.claude/PREFERENCES.md`).
- [x] `test_dry_run_predicts_stray_preferences_cleanup` matches dry-run preview to live behavior.
- [x] `test_no_stray_preferences_no_cleanup_message` guards against false-positive log line when no stray is present.
- [ ] Reviewer to spot-check that the new rationale paragraph in `_HOME_CLAUDE_EXCLUDES` matches the MEMORY.md entry's structure.
- [ ] Post-merge: install on the operator's machine, confirm "Removed stray ..." log line fires once and the stray PREFERENCES.md is gone.

### DIFF
--- a/src/kai/install.py
+++ b/src/kai/install.py
@@ -88,6 +88,23 @@ _SOURCE_EXCLUDES = {"__pycache__", "*.pyc", "*.egg-info", ".git", ".venv", ".env
 #                 install/home/.claude/MEMORY.md, which would resurrect
 #                 the pre-#347 global layout and shadow the per-user
 #                 reads on every session.
+#   PREFERENCES.md - same shape as MEMORY.md per #400/#401. Read
+#                 directly by _apply_migrate and
+#                 backend.ensure_user_preferences from
+#                 templates/.claude/PREFERENCES.md, written to
+#                 DATA_DIR/preferences/<chat_id>/PREFERENCES.md. The
+#                 exclude prevents _copy_tree from depositing a global
+#                 copy at install/home/.claude/PREFERENCES.md, which
+#                 (while not currently read by any code path) creates
+#                 a name collision that misleads anyone inspecting the
+#                 install layout and would silently shadow per-user
+#                 reads if a future runtime path ever resolves through
+#                 that location. Missed in #442 because the pre-rename
+#                 file was named PREFERENCES.md.example, which had no
+#                 collision with the per-user PREFERENCES.md path; the
+#                 suffix drop turned the install copy into the same
+#                 global-layout artifact MEMORY.md had been protected
+#                 against since #347.
 #   CLAUDE.md   - per-operator regular file (gitignored); created on
 #                 first install by copying templates/.claude/CLAUDE.md
 #                 into the install tree. The exclude prevents the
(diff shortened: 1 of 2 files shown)


dcellison:
## Review by Kai

The PR is functionally correct - the exclude-set addition, dry-run preview, and live cleanup are all logically sound. A few issues to flag:

**Warning - Convention violation (install.py, comment block ~line 88-105)**

The rationale paragraph for `PREFERENCES.md` ends with a "Missed in #442 because..." sentence explaining the historical cause. The project convention is explicit: "Don't reference the current task, fix, or callers... since those belong in the PR description and rot as the codebase evolves." The functional justification (why the file is excluded) should stay; the archaeology (why it was previously missed) belongs only in the PR description/commit message, which already has it.

Same issue in the live cleanup comment block (~line 2922): "left behind by an install that ran between #442/PR #445 (which renamed...)" - the historical explanation should be trimmed to just the what/why: this is a one-shot cleanup, the file is never read, safe to delete.

**Warning - Same convention violation in tests (test_install.py, ~line 4001)**

The docstring for `test_strips_stray_global_preferences_md` references `#442/PR #445` and `#400/#401`. Test docstrings that reference PR numbers rather than behavior will confuse future readers once those PRs are distant history. Describe what the test verifies, not the incident that motivated it.

**Suggestion - `unlink()` with no error handling (install.py, ~line 2938)**

```python
stray_preferences.unlink()
```

If this fails (macOS immutable flag, I/O error, concurrent modification), it raises an unhandled exception that aborts the install mid-flight after other steps have already run. Given root execution the risk is low, but a `try/except OSError as e: print(f"  Warning: could not remove stray {stray_preferences}: {e}")` would make the cleanup best-effort rather than fatal, which matches the "nothing reads the path" reasoning in the comment.

**Suggestion - Redundant variable definition**

`stray_preferences = ws_claude_dst / "PREFERENCES.md"` is defined identically in both the dry-run block and the live block. Defining it once before the dry-run block would eliminate the duplication. Minor, but easy cleanup.

**No bugs found.** The `is_file() and not is_symlink()` guard is correct (Python's `is_file()` follows symlinks; the combined check properly excludes symlinks from deletion). The dry-run `return` placement is correct and the live path will never execute in dry-run mode.

dcellison:
Superseded by #447. The wholesale-directory cleanup retires `<install>/home/.claude/` entirely, including the stray `PREFERENCES.md` this PR targeted; the per-entry exclude added here becomes vacuous because the only `_copy_tree` consumer of `_HOME_CLAUDE_EXCLUDES` is itself removed. The post-merge state of #447 is identical to (and broader than) what would have shipped here.

PR #448, opened 2026-05-10
fix(tests): pre-set KAI_DATA_DIR at conftest module-top to stop leaks

### LINKED ISSUE #444: Test runs leak files into the repo working tree via per-module DATA_DIR bindings
## Problem

`make test` writes runtime artifacts into the repo working tree. Today the visible residue at the repo root includes:

- `files/12345/` - 80+ files with names like `20260507_103732_990781_logo.png` and `20260507_103733_815169_photo_<MagicMock_name='mock.file_unique_id'_id='4511512432'>.jpg`. These come from `kai.bot._save_upload` writing to `DATA_DIR / "files" / str(user_id)` during handler tests that exercise `handle_photo` / `handle_document` with chat_id 12345.
- `memory/MEMORY.md` - the seeded "(currently empty)" placeholder. Written by `kai.backend.ensure_user_memory` on the `chat_id is None` legacy path (`backend.py:464-466`) when a test calls it without first patching the `data_dir` argument the function receives.
- `memory/extractor_cwd/` - empty directory. Created by `kai.memory_extraction._ensure_extractor_cwd()` (`memory_extraction.py:736`), which uses the module-level `_EXTRACTOR_CWD = DATA_DIR / "memory" / "extractor_cwd"` snapshot taken at import time (`memory_extraction.py:188`).

Earlier runs also leaked `home/<chat_id>/` directories at the repo root (the original motivation for PR #438's autouse fixture). That regressed enough times that #442 is now removing `home/` from the tracked tree entirely; the leak vector is the same family.

## Root cause

Two interacting facts:

1. **In dev, `DATA_DIR` equals `PROJECT_ROOT`.** `src/kai/config.py:34-40` derives `PROJECT_ROOT` from the file path of `config.py` and falls back to it when `KAI_DATA_DIR` is unset. Without that env var (the dev default), every "writable artifact" path under `DATA_DIR` lands inside the repo.
2. **`from kai.config import DATA_DIR` creates a per-module binding.** Every module that imports `DATA_DIR` (kai.backend, kai.bot, kai.memory, kai.memory_extraction, kai.webhook, kai.claude, kai.goose, kai.main, kai.history) gets its own attribute in its own module namespace. Some of those bindings are also frozen further into module-level constants:
   - `kai.bot._RESPONDING_DIR = DATA_DIR / ".responding"` (line 113)
   - `kai.memory_extraction._EXTRACTOR_CWD = DATA_DIR / "memory" / "extractor_cwd"` (line 188)
   - `kai.history._LOG_DIR = DATA_DIR / "history"` (line 36)

PR #438's `_isolate_backend_data_dir` fixture in `tests/conftest.py:84-103` patches exactly one of these: `kai.backend.DATA_DIR`. Tests that exercise `_save_upload`, `init_memory`, `_ensure_extractor_cwd`, or any other code path reading from a different module's binding sail right past the fixture and write to the real `DATA_DIR`, which in dev is the repo root.

## Why this stays invisible

Every leak path is in `.gitignore` (`files/`, `memory/`, `home/*`, etc.), so `git status` reports a clean tree even when test runs are leaking heavily. The leaks accumulate silently across runs and are only spotted by occasionally `ls`-ing the repo root.

This is the same shape as the bug PR #438 was meant to close; it just left the other half of the doors open.

## Scope

Make the test suite incapable of writing into `PROJECT_ROOT`-relative DATA_DIR locations regardless of which module's binding the code under test reads.

The mechanism is up to the spec, but two approaches dominate the design space:

- **Pre-set `KAI_DATA_DIR` in `tests/conftest.py` module-top.** Mirrors the existing `MEM0_DIR` fix in conftest (lines 38-46): set `os.environ["KAI_DATA_DIR"]` to a session-scoped temp dir before any `kai.*` module is imported, register an `atexit` cleanup. Because `kai.config:34-40` reads `KAI_DATA_DIR` at module import, every downstream `from kai.config import DATA_DIR` will see the temp dir as its starting value, and the import-time-snapshot constants (`_EXTRACTOR_CWD`, `_RESPONDING_DIR`, `_LOG_DIR`) will be safe-by-construction. The existing per-test `_isolate_backend_data_dir` fixture stays as defense-in-depth.
- **Patch every module's `DATA_DIR` binding plus the snapshot constants.** Extends the autouse fixture to patch `kai.bot.DATA_DIR`, `kai.memory.DATA_DIR`, `kai.memory_extraction.DATA_DIR`, and the frozen constants. Loses the import-time guarantee; gains nothing in return; will silently regress the next time someone adds a new module that imports `DATA_DIR`.

The first approach is cheaper, broader, and survives future modules without maintenance. Spec author should default to it unless something specific to the install / config tests forces otherwise.

## Acceptance

- A clean `make test` run leaves `git status -uall` reporting nothing under `files/`, `memory/`, `home/`, `history/`, `logs/`, `preferences/`, `models/`, `kai.db`, `.responding/`, or `.responding_to` at the repo root.
- The existing residue at the repo root (`files/12345/`, `memory/`) is removed in the same PR, with a one-line note in the PR description explaining the cleanup.

## Out of scope

- Removing `home/` from the tracked tree - that is #442 and is being implemented in parallel.
- Making `DATA_DIR != PROJECT_ROOT` the dev default. Multiple workflows (running `python -m kai` from a fresh checkout without an installed config) depend on the current behavior, and changing it is a much larger blast radius than this issue is meant to address.
- Refactoring per-module `DATA_DIR` imports into a single function call. The import-pattern is consistent with the rest of the codebase and changing it is a separate refactor.

## Notes for the spec author

- The conftest already demonstrates the env-var-at-module-top pattern (`MEM0_DIR` block at lines 8-46). The spec should use the same shape for `KAI_DATA_DIR`.
- Verify before finalizing: is `KAI_DATA_DIR` read anywhere outside `src/kai/config.py:40`? If so the override needs to land before any other reader runs.
- The autouse `_isolate_backend_data_dir` fixture at lines 84-103 should stay; it doesn't conflict with a session-level env var, and it adds per-test isolation on top of the session-level one.
- Audit `tests/test_install.py` for tests that deliberately read or write under `PROJECT_ROOT`. If any exist, they need a documented exemption (or a redirect), not a removal of the safety net.

dcellison:
Closes #444.

## Summary

`make test` was leaking artifacts into the repo working tree because `kai.config:40` resolves `DATA_DIR` at module-import time and falls back to `PROJECT_ROOT` when `KAI_DATA_DIR` is unset (the dev default). PR #438 closed `kai.backend.DATA_DIR` per-test; this PR closes the remaining surface (eight other module bindings plus the import-time-snapshot constants `_EXTRACTOR_CWD`, `_RESPONDING_DIR`, `_LOG_DIR`) by setting `KAI_DATA_DIR` to a session-scoped tempdir at `tests/conftest.py` module-top, BEFORE any `kai.*` module is imported.

Mirrors the existing `MEM0_DIR` block in the same file. Single-file change. Both existing autouse fixtures stay as per-test scoping on top.

## Test plan

- [x] `make check` passes.
- [x] `make test` passes (2818 tests, 1 skipped).
- [x] After `rm -rf files/ memory/ home/ history/ logs/ preferences/ models/ .responding/ .responding_to kai.db` and a fresh `make test`, none of those paths reappear at the repo root.
- [x] `git status -uall` clean (excepting an unrelated untracked `.claude/CLAUDE.md`).
- [x] `TestDataDir` cases in `tests/test_config.py:428-457` still pass without modification (they re-derive via `os.environ.get(...) or str(PROJECT_ROOT)` after `_clean_env`'s `delenv`).

## Operator cleanup (one-time)

After pulling this branch, run the following at the repo root to remove any test residue accumulated from prior runs:

```
rm -rf files/ memory/ home/ history/ logs/ preferences/ models/ .responding/ .responding_to kai.db
```

These paths are `.gitignore`d, so the cleanup cannot ride the diff. Errors for paths that do not exist on a given checkout can be ignored.

**Caveat:** back up `kai.db` first if you've been running Kai locally in dev mode (`python -m kai` from the repo root, with `KAI_DATA_DIR` unset). The dev fallback puts `kai.db` under `PROJECT_ROOT`, and the cleanup command will destroy any conversation history stored there. Production installs are unaffected (their data lives under `/var/lib/kai/`).

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -59,6 +59,46 @@ def _cleanup_mem0_test_dir() -> None:
     shutil.rmtree(_MEM0_TEST_DIR, ignore_errors=True)
 
 
+# ── Session-wide DATA_DIR isolation ─────────────────────────────────
+#
+# `kai.config:40` resolves DATA_DIR at module-import time: it falls
+# back to PROJECT_ROOT when KAI_DATA_DIR is unset. In dev that means
+# DATA_DIR == PROJECT_ROOT, so any code path that writes to
+# `DATA_DIR / "files"`, `DATA_DIR / "memory"`, etc. lands inside the
+# repo working tree. The per-test autouse `_isolate_backend_data_dir`
+# fixture (below) patches `kai.backend.DATA_DIR`, but `from kai.config
+# import DATA_DIR` creates a per-module binding in every importer
+# (`kai.bot`, `kai.memory`, `kai.memory_extraction`, etc.), and some
+# of those bindings are frozen further into module-level constants
+# (e.g. `kai.memory_extraction._EXTRACTOR_CWD = DATA_DIR / "memory" /
+# "extractor_cwd"` at line 188, evaluated once at import). Patching
+# `kai.backend.DATA_DIR` after the fact does not reach those.
+#
+# Setting KAI_DATA_DIR in os.environ BEFORE any kai module is
+# imported makes the import-time fallback resolve to a per-session
+# tempdir for every importer at once, which is the only mechanism
+# that catches the frozen snapshots. The autouse fixtures below stay
+# as defense-in-depth and give per-test isolation on top.
+#
+# The shape (mkdtemp + hard override + atexit) mirrors the MEM0_DIR
+# block above; the same conftest-module-top ordering guarantee
+# applies (conftest runs before pytest collects test modules, which
+# is when their `from kai.config import ...` would otherwise fire).
+_KAI_DATA_TEST_DIR = tempfile.mkdtemp(prefix="kai-test-data-")
+os.environ["KAI_DATA_DIR"] = _KAI_DATA_TEST_DIR
+
+
+@atexit.register
+def _cleanup_kai_data_test_dir() -> None:
+    """Best-effort removal of the per-run KAI_DATA_DIR.
+
+    Mirrors `_cleanup_mem0_test_dir`. Errors are swallowed so a stray
+    open file handle from a crashed worker cannot mask a real test
+    failure with a secondary traceback at interpreter shutdown.
+    """
+    shutil.rmtree(_KAI_DATA_TEST_DIR, ignore_errors=True)
+
+
 # ── Per-test history isolation ──────────────────────────────────────
 
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -84,6 +84,13 @@ def _cleanup_mem0_test_dir() -> None:
 # block above; the same conftest-module-top ordering guarantee
 # applies (conftest runs before pytest collects test modules, which
 # is when their `from kai.config import ...` would otherwise fire).
+#
+# Interaction with `tests/test_config.py:_clean_env`: that autouse
+# fixture does `monkeypatch.delenv("KAI_DATA_DIR")` per test, which
+# pytest auto-restores at fixture teardown. The session-wide value
+# set here therefore persists across test boundaries; only the test
+# body itself sees the env var as absent, which is what the
+# `TestDataDir` cases rely on to exercise the PROJECT_ROOT fallback.
 _KAI_DATA_TEST_DIR = tempfile.mkdtemp(prefix="kai-test-data-")
 os.environ["KAI_DATA_DIR"] = _KAI_DATA_TEST_DIR
 


dcellison:
## Review by Kai

**Overall: clean fix. Mirrors an established pattern correctly and the reasoning holds up.**

---

**Suggestion - silent env override (`conftest.py`, line ~87)**

```python
os.environ["KAI_DATA_DIR"] = _KAI_DATA_TEST_DIR
```

If a developer has `KAI_DATA_DIR` already set in their shell (e.g., pointing at a real dev data dir), this overwrites it silently. The existing `MEM0_DIR` block does the same, so this is consistent, but it can surprise someone running tests in an environment where they've customized that var. A one-line warning via `warnings.warn` or a `print` to stderr when a pre-existing value is detected would make the override visible. Low priority given the existing convention.

---

**Suggestion - `_clean_env` / `delenv` interaction (`tests/test_config.py:428-457`)**

The PR description notes that `TestDataDir` cases call `_clean_env` which does `delenv("KAI_DATA_DIR")`. After those tests finish, `KAI_DATA_DIR` is absent from `os.environ`. Any code path that reads `os.environ.get("KAI_DATA_DIR")` at runtime (not import time) would then fall back to `PROJECT_ROOT` for the rest of the session. The PR author argues this doesn't matter because `kai.config` resolves `DATA_DIR` only at import time and all module-level constants are already frozen. That argument is correct for the current code. Worth a comment in `test_config.py` near `_clean_env` noting that `KAI_DATA_DIR` is not restored, since this is now load-bearing for future maintainers.

---

**No bugs, no security issues, no critical findings.** The mechanism (set env var at conftest module-top before any `kai.*` import) is the right tool for this class of problem. The `atexit` + `ignore_errors=True` cleanup is correct and consistent with the `MEM0_DIR` block. The comment block is thorough and accurate.

dcellison:
## Review by Kai

**Overall: clean. No new issues beyond what was raised in the prior review.**

The mechanism is correct and the implementation mirrors the existing `MEM0_DIR` block faithfully.

---

**One minor observation (suggestion)**

`conftest.py`, the new block -- `tempfile.mkdtemp` is the right call here, but it creates a real directory on disk at module-import time, meaning a crashed or force-killed test run can leave a `kai-test-data-*` directory in `/tmp` that the `atexit` handler never gets to remove. The `MEM0_DIR` block above has the same behavior, so this is consistent and not a regression, but it is worth noting that the `atexit` approach gives up cleanup on SIGKILL or hard crash. A `pytest` session-scoped `tmp_path_factory` fixture would give automatic cleanup even on crash, at the cost of moving the assignment out of module-top into a fixture -- which would break the import-time guarantee this PR depends on. So the current approach is the correct tradeoff; no action required.

---

**Correctness check on the `monkeypatch.delenv` interaction**

The comment at line ~90 correctly describes the restoration behavior: `monkeypatch.delenv` removes the var for the test body, pytest restores it at fixture teardown, and the session-wide value is unaffected at test-boundary crossings. The `TestDataDir` cases in `test_config.py:428-457` see the var as absent during their body (correct) and the session env is intact afterward (correct).

---

**No bugs, no security issues, no critical findings.** The two suggestions from the prior review (silent-override warning, comment near `_clean_env`) were not addressed, but that was the author's call and I won't re-raise them.

PR #449, opened 2026-05-10
refactor(install): retire dead home/.claude/ scaffolding (#447)

### LINKED ISSUE #447: Retire dead home/.claude/ scaffolding from install tree
## Finding

The `<install_path>/home/.claude/` subtree (containing `CLAUDE.md`, plus historical `.example` cruft and the recently-migrated regular CLAUDE.md from #442) is no longer read by any runtime code path. Every session resolves identity from the per-user `<DATA_DIR>/home/<chat_id>/.claude/CLAUDE.md` instead.

## Evidence

- `kai.backend.build_session_context` reads `home_workspace / ".claude" / "CLAUDE.md"` where `home_workspace` is the resolved per-user path returned by `kai.backend.resolve_home_workspace` / `kai.backend.ensure_user_home`, which lands at `<DATA_DIR>/home/<chat_id>/` (or whatever override an operator set in users.yaml).
- No session has cwd inside `<install_path>/home/`. Pre-`#353` it was the shared global workspace; post-`#353` every session runs under the per-user tree.
- `<install_path>/home/.claude/CLAUDE.md` is populated by `_apply_source` at install time and never read again. Pure dead scaffolding.

## Proposed change

1. Stop populating `<install_path>/home/.claude/` at install time. Drop the relevant `_copy_tree(templates/.claude/, ...)` call and the conditional CLAUDE.md seed step.
2. One-shot install-time cleanup: remove `<install_path>/home/.claude/` if it exists, including the migrated CLAUDE.md, the residual `.example` files, and the (now-also-dropped-by-#446) PREFERENCES.md.
3. Update the `_HOME_CLAUDE_EXCLUDES` machinery accordingly — most of it goes away with the directory.
4. Tests:
   - Pin that `<install_path>/home/.claude/` does not exist after a fresh install.
   - Pin that an existing-install upgrade removes any pre-existing `<install_path>/home/.claude/` content.
   - Verify per-user identity injection still resolves through `<DATA_DIR>/home/<chat_id>/.claude/CLAUDE.md`.
5. Wiki: any installation docs describing `<install_path>/home/.claude/` as load-bearing get updated; the runtime identity path is per-user.

## Out of scope

- `<DATA_DIR>/home/<chat_id>/` per-user workspaces — these are the live runtime path, untouched.
- `<install_path>/home/config/goose-config.yaml` — still consumed by `_apply_goose_config` at install time. A relocation to `<install_path>/config/` or `/etc/kai/config/` is a separate cleanup that does not need to ride this issue.
- Renaming `<install_path>/home/` itself. After this issue lands, `<install_path>/home/` will only contain `config/`; whether to rename the parent is a follow-up cosmetic question.
- The `_migrate_identity_to_claude_md` helper retirement — that has its own 30-day cleanup window already on the schedule.

## Risk

- An operator who manually customized `<install_path>/home/.claude/CLAUDE.md` believing it was load-bearing loses their edits during the install-time cleanup. Mitigation: the migration prints the file path and a one-line note that nothing reads it before deletion, so the operator can recover from the printed log if needed. Low real-world risk; the path has not been read by any code since `#353`.
- Any third-party tool/script someone built around the install path resolves to a missing directory. Same mitigation: the install output names the removal explicitly so the breakage is visible, not silent.

## Why surfaced now

Caught while answering a "what is `<install_path>/home/` for?" question post-merge of `#445`. The IDENTITY.md migration in `#445` was operating correctly on this subtree, but the migration's payoff (a clean regular CLAUDE.md operators can `cat` to read their identity baseline) is the only remaining purpose of the path. If we are willing to retire the path entirely, the migration's payoff goes with it and the install layout simplifies further.

dcellison:
Closes #447. Supersedes #446 (closed in the same session).

## Summary

Since #353 every session resolves identity from the per-user `<DATA_DIR>/home/<chat_id>/.claude/CLAUDE.md`; nothing reads `<install>/home/.claude/` at runtime. This PR retires the install-tree scaffolding wholesale:

- Removes the `_copy_tree(templates/.claude/, ...)` step, the conditional CLAUDE.md seed, the reconcile chown, the `_HOME_CLAUDE_EXCLUDES` constant + its 25-line rationale block, and the orphaned `templates/.claude/CLAUDE.md` template.
- Adds `_retire_install_home_claude(install_path, dry_run)` that removes any pre-existing `<install>/home/.claude/` subtree plus any legacy `<install>/home/IDENTITY.md`. Idempotent, dry-run parity, every removed file logged with its byte size.
- Removes the `_migrate_identity_to_claude_md` CALL from `_apply_source`. The helper BODY stays per the issue's 30-day cleanup window carve-out; the new cleanup picks up any leftover IDENTITY.md directly.
- Folds in the work #446 was targeting: the wholesale-directory cleanup retires the stray root-owned `<install>/home/.claude/PREFERENCES.md` from the operator's post-#445 install. `_HOME_CLAUDE_EXCLUDES` itself goes away.
- Doc fix-ups across README.md, templates, eval/behavioral.py, three test files, and `_copy_tree`'s symlink-handling docstring.

## Spec deltas (defects caught during implementation)

Items 1-4 are corrections to spec defects identified during the eval; items 5-8 are defensible widenings the implementer chose for correctness/robustness that diverge from the spec's literal text.

1. §4.2.3's "delete lines 2826-2862" range would have eaten the `config_src` dry-run preview (2828-2829) and the early `return` (2862); both must stay. Implementation keeps both.
2. §6 acceptance grep didn't cover `_copy_tree` docstring at install.py:1284 (a historical-narrative `home/.claude/CLAUDE.md` reference); added an in-scope edit pointing the docstring at issues #442 AND #447.
3. §4.5.1-3 proposed literal `<DATA_DIR>` placeholders in operator-facing docs; README uses `DATA_DIR/...` (no brackets) by convention while templates use `<DATA_DIR>/...` (with brackets). Implementation matches each file's local convention.
4. §5.2.4's 1:1 dry-run/live parity assertion would have failed against `capsys`'s full-stdout capture; implementation filters on the `Removed `/`Would remove ` prefix.
5. **IDENTITY.md predicate widened.** Spec §4.1 sample code: `if identity_md.is_file() and not identity_md.is_symlink():`. Implementation: `if identity_md.is_symlink() or identity_md.is_file():`. The retired path is removed regardless of shape (regular file, valid symlink, or broken symlink); directories remain excluded as a bizarre malformed state. New regression test `test_symlink_at_identity_md_is_removed` pins the broader behavior. Brings the test count for `TestRetireInstallHomeClaude` to 6 (not 5 as the §5.2 plan listed).
6. **Verb tense changed for live per-file log lines.** Spec §4.1 sample: `print(f"  Removed {path} ...")`. Implementation: `Removing` (live), with `Removed` reserved for the post-rmtree summary and post-unlink IDENTITY.md lines. The per-file enumeration logs BEFORE `shutil.rmtree` runs (it needs to `lstat` each file while it still exists); on a midway rmtree failure, "Removed" would be a past-tense lie on a pre-fact print.
7. **`_copy_tree` docstring rewritten beyond a path swap.** Spec §4.5.7 (added in the eval) implied keeping the literal `home/.claude/CLAUDE.md -> ../IDENTITY.md` text and only retargeting the issue reference. Implementation replaces the path-text with "the legacy IDENTITY.md / CLAUDE.md symlink pair under the install tree" so the docstring stops naming a retired path at all. Cleaner for future-grep hygiene; diverges from the eval's literal proposal.
8. **Dry-run summary lines extended for parity.** Spec §4.1 sample's dry-run summary was `[DRY RUN] Would remove dead {claude_dir}` (no trailing context); the live summary is `Removed dead {claude_dir}; nothing reads this path post-#447`. Without parity, the §5.2.4 prefix-stripped comparison would fail. Implementation extends BOTH dry-run summary lines (`.claude/` and IDENTITY.md) to match their live counterparts after prefix strip.

## Test plan

- [x] `make check` passes; lint clean.
- [x] `make test` passes (2816 tests, 1 skipped; net -2 vs main, reflecting test removals minus the new cleanup tests).
- [x] §6 acceptance greps:
  - `grep -rn '_HOME_CLAUDE_EXCLUDES' src/ tests/` -> nothing.
  - `grep -rn 'templates/\.claude/CLAUDE\.md' src/ tests/` -> nothing.
  - `git ls-files templates/.claude/` -> `MEMORY.md`, `PREFERENCES.md` only.
  - `grep -rn 'home/\.claude' src/ templates/ README.md` -> only hits inside the two retirement helpers' own docstrings/comments.
- [x] `_apply_source` no longer references `_migrate_identity_to_claude_md`, `_HOME_CLAUDE_EXCLUDES`, `templates/.claude`, `claude_md_src`, or `claude_md_dst`.
- [x] `TestRetireInstallHomeClaude`: 6 tests covering fresh-install / existing-dir-removed / legacy-IDENTITY.md-regular / legacy-IDENTITY.md-symlink / dry-run parity / idempotency.

## Wiki

Out of repo per issue bullet 5. Pages referencing `<install_path>/home/.claude/` as load-bearing need a refresh; the operator handles those after merge.

## Operator cleanup note

On the next `make install` after this lands, install output names every file being removed from `<install>/home/.claude/` plus a `Removed dead <install>/home/.claude/; nothing reads this path post-#447` summary. If you customized the install-tree CLAUDE.md believing it was load-bearing, the printed log is the recovery surface. Production impact is low: the path has not been read by any code since #353.

If you have been running Kai locally in dev mode (`python -m kai` from the repo with `KAI_DATA_DIR` unset) and edited a dev CLAUDE.md, back it up first; the same cleanup helper runs against your dev install tree.


dcellison:
## Review by Kai

This is a clean refactor with good test coverage. Here are the findings:

---

## Warning

**`install.py:_retire_install_home_claude` - individual file log lines are printed before `shutil.rmtree` executes**

The loop prints `  Removed {path} ({size} bytes)` (past tense) for each file inside `claude_dir`, but `shutil.rmtree(claude_dir)` runs after the loop completes. If `rmtree` raises (permissions failure, mounted FS), stdout already shows every file as "Removed" even though nothing was deleted. The IDENTITY.md branch is correct - it calls `unlink()` first, then prints.

The docstring explicitly says this is intentional for operator recovery ("logged with its byte size before deletion"), but the verb tense is wrong for that intent. "Removing" or "Will remove" would be accurate; "Removed" is a post-fact claim printed pre-fact.

No data is corrupted by this, but an operator seeing `  Removed home/.claude/CLAUDE.md (512 bytes)` followed by an OSError summary would wrongly believe their file was destroyed when it wasn't.

---

## Suggestion

**`install.py:_retire_install_home_claude` - stale comment on the `IDENTITY.md` symlink guard**

The inline comment says:

> a symlink at the same path would have been handled by row 1/4 of the migration helper

The migration helper (`_migrate_identity_to_claude_md`) is no longer called from `_apply_source` as of this PR. So a symlink at `<install>/home/IDENTITY.md` is now silently skipped by both helpers. The comment is factually wrong post-#447 and could mislead the next person to touch this code. The scenario is likely theoretical (IDENTITY.md was always the regular-file source; CLAUDE.md was the symlink), but the justification text should be updated or removed.

---

## Clean

- Dry/live parity in `_retire_install_home_claude` holds: the prefix-stripping logic in the parity test correctly aligns `[DRY RUN] Would remove` against `Removed` body text including the directory summary line.
- The `rglob` + `lstat` approach for logging symlinks before `rmtree` is sound; symlinks are reported without dereferencing a potentially broken target.
- `shutil.rmtree` without `ignore_errors` is the right call given the explicit `is_dir()` pre-check.
- The `home/` parent `mkdir(parents=True, exist_ok=True)` guard on the config copy is correct after the `.claude/` copy that previously created `home/` was removed.
- Test coverage in `TestRetireInstallHomeClaude` covers the five meaningful states (fresh, existing dir, bare IDENTITY.md, dry/live parity, idempotency).
- Documentation and comment updates are accurate and convention-consistent (`DATA_DIR/...` vs `<DATA_DIR>/...` per-file conventions respected).

dcellison:
## Review by Kai

I'll review the diff directly from what's provided.

---

## PR Review: `feature/retire-dead-home-claude`

**Both prior review findings are addressed:**
- Verb tense fixed: per-file lines now say `"Removing"` (present, printed before `rmtree`); the summary `"Removed dead {dir}"` fires only after `rmtree` returns successfully. The docstring spells this out explicitly.
- Stale migration-helper comment: the `is_symlink()` guard comment no longer references `_migrate_identity_to_claude_md` rows at all.

---

### Warning

**`_retire_install_home_claude` - `stat()` vs `lstat()` inconsistency on IDENTITY.md**

The `claude_dir` loop uses `path.lstat().st_size` (correct - avoids dereferencing a broken symlink target). The `identity_md` branch uses `identity_md.stat().st_size`. The `is_file() and not is_symlink()` guard makes these equivalent in practice (it's confirmed not a symlink by that point), but it breaks the invariant the docstring establishes for this function and will confuse anyone who looks at the two branches side-by-side. Change `stat()` to `lstat()` for consistency.

Location: `install.py`, the `if identity_md.is_file() and not identity_md.is_symlink():` block.

---

### Suggestion

**`rglob("*")` silently omits empty subdirectories from the removal log**

The enumeration loop filters to `is_file() or is_symlink()`, so empty directories under `.claude/` (e.g., a `skills/` dir with no files) are deleted by `rmtree` without appearing in the `"Removing"` output. The docstring's claim that "every retired path is logged" is slightly overclaimed. This only matters if an operator had created an empty skills directory and wants to know it was removed. Cosmetic, not a correctness bug - `rmtree` handles it cleanly.

---

### Clean

- `shutil.rmtree` without `ignore_errors` is appropriate given the `is_dir()` pre-check.
- `config_dst.parent.mkdir(parents=True, exist_ok=True)` guard is correct after the `.claude/` copy (which used to create `home/`) was removed.
- The dry/live parity test is sound: `removal_lines()` strips the tense-marker prefix, substitutes the install path, and compares bodies 1:1. The `assert dry_lines` guard prevents a vacuous pass.
- `templates/.claude/MEMORY.md` and `PREFERENCES.md` remaining in git but no longer copying to the install tree is intentional and consistent with the PR's claim that nothing reads `<install>/home/.claude/` since #353.
- Test coverage covers all five meaningful states (fresh, existing dir, bare IDENTITY.md, dry/live parity, idempotency).
- Documentation convention (`DATA_DIR/...` in README, `<DATA_DIR>/...` in templates) is respected throughout.

dcellison:
## Review by Kai

Both prior review findings are addressed (verb tense, `lstat` consistency, stale migration-helper comment). The PR is clean on logic, security, and convention. One new finding:

---

## Suggestion

**`install.py:_retire_install_home_claude` - symlink at `IDENTITY.md` is silently skipped without operator notification**

The guard:

```python
if identity_md.is_file() and not identity_md.is_symlink():
```

correctly refuses to remove a symlink at `IDENTITY.md`. The docstring explicitly says this case "warrants operator attention rather than silent removal" - but no warning line is emitted. An operator running `make install` would see nothing, leave a stale symlink behind, and have no log entry pointing at it.

A single `else` / `elif identity_md.is_symlink()` print would close the gap:

```python
elif identity_md.is_symlink():
    if dry_run:
        print(f"[DRY RUN] Warning: {identity_md} is a symlink; skipping (inspect manually)")
    else:
        print(f"  Warning: {identity_md} is a symlink; skipping (inspect manually)")
```

The scenario is noted as likely-theoretical, but the docstring already promises attention - the code just doesn't deliver it.

---

## Clean

- `_retire_install_home_claude` verb-tense invariant is sound: per-file lines say `Removing` (printed before `rmtree`); the summary `Removed dead {dir}` fires only after `rmtree` returns.
- `lstat` is used consistently in both branches.
- `shutil.rmtree` without `ignore_errors` is correct given the explicit `is_dir()` pre-check.
- `config_dst.parent.mkdir(parents=True, exist_ok=True)` guard is correct after the `.claude/` copy (which previously created `home/`) was removed.
- Dry/live parity test is sound: `removal_lines()` strips the tense-marker prefix, normalizes the install path, and requires `dry_lines` to be non-empty before comparing.
- Test coverage spans all five states (fresh, existing dir, bare IDENTITY.md, dry/live parity, idempotency).
- Documentation convention (`DATA_DIR/...` in README and `services.yaml`, `<DATA_DIR>/...` in templates and `MEMORY.md`) is respected throughout.
- `templates/.claude/CLAUDE.md` deletion and `_HOME_CLAUDE_EXCLUDES` removal are clean; no remaining call sites.

dcellison:
## Review by Kai

## PR Review: `feature/retire-dead-home-claude`

All three prior-review findings are resolved: verb tense is correct, `lstat` is consistent, and the IDENTITY.md symlink case is now removed wholesale (rather than skipped with a warning) -- `test_symlink_at_identity_md_is_removed` validates this.

---

### Warning

**`test_dry_run_predicts_cleanup_matches_live` - dry-run pass is unpatched and hits the real filesystem**

The dry-run call:
```python
with patch("kai.install.PROJECT_ROOT", src):
    _apply_source(install_dry, svc_uid=1000, svc_gid=1000, dry_run=True)
```
has no patches on `_copy_tree`, `_set_ownership`, `shutil.copy2`, or `os.chown`. That is intentional and harmless here because `dry_run=True` returns before calling those functions, but the fixture at `install_dry` includes real files that `_retire_install_home_claude` will `lstat` and log without patching. In the live path, those same helpers ARE patched. The asymmetry is not a bug but is worth a comment -- someone rearranging the test patches could inadvertently let the dry-run pass mutate `install_dry` (e.g., if a future code path calls `rmtree` in the pre-check branch).

---

### Suggestion

**`eval/behavioral.py` comment uses `home_workspace/.claude/CLAUDE.md` while README uses `DATA_DIR/home/<chat_id>/.claude/CLAUDE.md`**

The updated comment at `src/kai/eval/behavioral.py` (line ~539 in the diff) uses the schematic `home_workspace/.claude/CLAUDE.md`. The `tests/test_eval_behavioral.py` update uses the same form. The README and `templates/.claude/MEMORY.md` update use the expanded `DATA_DIR/home/<chat_id>/.claude/CLAUDE.md`. Both are accurate, but a reader tracing from the comment to the README or vice versa has to do mental mapping. Picking one form and using it consistently in comments would help.

---

### Clean

- `_retire_install_home_claude` verb-tense invariant is sound: per-file lines use `Removing` (before `rmtree`); the `Removed dead` summary fires only after `rmtree` returns; the `Removed legacy` IDENTITY.md line fires after `unlink()`.
- `is_symlink() or is_file()` guard correctly covers all three IDENTITY.md pre-states (regular file, valid symlink, broken symlink) without a directory false-positive.
- `removal_lines()` parity logic is correct: the three live prefixes (`Removing `, `Removed ` for the per-file, dir-summary, and IDENTITY.md phases) plus the single dry prefix (`[DRY RUN] Would remove `) all reduce to the same body after stripping. The `assert dry_lines` guard prevents a vacuous pass.
- `config_dst.parent.mkdir(parents=True, exist_ok=True)` placement is correct post-retirement: `home/` is no longer implicitly created by the `.claude/` copy and must be created explicitly before the `config/` copy.
- `TestPreferencesTemplateShipsViaCopyTree` deletion is correct; `_HOME_CLAUDE_EXCLUDES` is gone so the test has no target and would have failed on import.
- `TestRetireInstallHomeClaude` covers all five meaningful states (fresh, existing dir, bare IDENTITY.md, dry/live parity, idempotency). The idempotency test correctly asserts no `Removed dead` or `Would remove dead` output on the second pass.
- Documentation convention (`DATA_DIR/...` in README and `services.yaml`, `<DATA_DIR>/...` in templates) is respected throughout.

dcellison:
## Review by Kai

## PR Review: `feature/retire-dead-home-claude`

All four prior-review rounds of findings are resolved. The implementation is clean.

---

### Warning

**`test_dry_run_predicts_cleanup_matches_live` - `removal_lines` strips `"Removed "` but not `"Removing "`-then-`"Removed "` ambiguity in the wrong order**

Actually this is fine on closer inspection - `"Removing "` and `"Removed "` don't share a prefix relationship (character 6 is `'i'` vs `'e'`), so the prefix order in `("Removing ", "Removed ")` doesn't matter for correctness. No real issue here; I'm calling it out explicitly so future readers don't have to re-derive it.

---

### Clean

Everything else is correct:

- **`_retire_install_home_claude` verb tense invariant** holds throughout the diff: per-file lines use `"Removing"` (printed before `rmtree` so the present tense is accurate on `rmtree` failure); the `"Removed dead"` summary fires only after `rmtree` returns; `"Removed legacy"` fires only after `unlink()` returns.
- **`lstat` consistency** is correct in both branches; the `claude_dir` loop and the `identity_md` branch both use `lstat`, which avoids tracebacking on a broken symlink target.
- **`is_symlink() or is_file()` guard** on `identity_md` correctly covers regular files, valid symlinks, and broken symlinks without a directory false-positive. `test_symlink_at_identity_md_is_removed` uses `identity.is_symlink()` (not `identity.exists()`) to verify the broken-symlink case, which is the right check.
- **`shutil.rmtree` without `ignore_errors`** is correct given the explicit `is_dir()` pre-check.
- **`config_dst.parent.mkdir(parents=True, exist_ok=True)`** placement is correct post-retirement: `home/` is no longer implicitly created by the removed `.claude/` copy, so the explicit mkdir before the `config/` copy is load-bearing.
- **Dry/live parity test** is sound: `removal_lines()` strips the tense-marker prefixes and normalizes the install path before comparing, the `assert dry_lines` guard prevents a vacuous pass, and the two install dirs (`install_dry` / `install_live`) are seeded identically so byte sizes match.
- **`TestRetireInstallHomeClaude`** covers all five meaningful states (fresh, existing dir, bare IDENTITY.md, broken-symlink IDENTITY.md, dry/live parity, idempotency).
- **`_HOME_CLAUDE_EXCLUDES` removal** is clean; `TestPreferencesTemplateShipsViaCopyTree` is correctly deleted as its only import target is gone.
- **Documentation convention** (`DATA_DIR/...` in README and `services.yaml`, `<DATA_DIR>/...` in templates and `MEMORY.md`) is respected throughout.
- **No em dashes** introduced; the prior `—` in `behavioral.py` is replaced with a semicolon per convention.

PR #453, opened 2026-05-11
refactor(install): relocate goose-config.yaml out of <install>/home/, retire the parent (#452)

### LINKED ISSUE #452: Relocate goose-config.yaml out of <install_path>/home/ and retire the home/ parent
## Finding

After #447 retired `<install_path>/home/.claude/` and the legacy `IDENTITY.md`, `<install_path>/home/` exists for a single purpose: hosting `<install_path>/home/config/goose-config.yaml`. That file is consumed by `_apply_goose_config` at install time to seed the Goose backend's extension config. Nothing else lives under `<install_path>/home/` anymore.

This was explicitly carved out as a follow-up in #447's out-of-scope list:

> `<install_path>/home/config/goose-config.yaml` — still consumed by `_apply_goose_config` at install time. A relocation to `<install_path>/config/` or `/etc/kai/config/` is a separate cleanup that does not need to ride this issue.

> Renaming `<install_path>/home/` itself. After this issue lands, `<install_path>/home/` will only contain `config/`; whether to rename the parent is a follow-up cosmetic question.

This issue is the follow-up.

## Proposed change

Relocate the goose config to `<install_path>/config/goose-config.yaml`, then retire `<install_path>/home/` entirely.

1. Change `_apply_source`'s `config_dst` from `install_path / "home" / "config"` to `install_path / "config"`. The source side (`templates/config/`) is unchanged.
2. Change `_apply_goose_config`'s read path from `install_path / "home" / "config" / "goose-config.yaml"` to `install_path / "config" / "goose-config.yaml"`.
3. Add a one-shot install-time cleanup: after the new config copy lands, if `<install_path>/home/` exists with only `config/` (or empty), remove it. Idempotent on reinstalls where the directory is already gone. Dry-run preview matches.
4. Update tests for both call-site changes plus a new test pinning the cleanup.

## Acceptance

- Fresh install: `<install_path>/home/` does not exist after `_cmd_apply` completes. `<install_path>/config/goose-config.yaml` exists and is consumed by `_apply_goose_config`.
- Reinstall on a host whose pre-state has `<install_path>/home/config/goose-config.yaml`: the new copy lands at `<install_path>/config/goose-config.yaml`, the old `<install_path>/home/config/goose-config.yaml` is removed, the parent `<install_path>/home/` is removed.
- Reinstall on a post-cleanup host: the helper is a silent no-op.
- `grep -rn 'home/config' src/ tests/` returns no production hits after the PR.

## Out of scope

- Relocating to `/etc/kai/config/` (the issue body mentions it as an alternative). `<install_path>/config/` keeps the config local to the install tree, matching the rest of the static-template layout.
- Wiki updates naming the old path. Will ride the PR description.

dcellison:
Closes #452. Follow-up to #447.

## Summary

After #447 retired `<install>/home/.claude/` and the legacy `IDENTITY.md`, `<install>/home/` existed for one purpose only: hosting `<install>/home/config/goose-config.yaml`. This PR relocates that file to `<install>/config/` and retires the vestigial parent.

- `_apply_source`'s `config_dst` -> `install_path / "config"`. Source side (`templates/config/`) unchanged.
- `_apply_goose_config` reads from the new path.
- New `_retire_install_home_dir(install_path, dry_run)` helper next to `_retire_install_home_claude`. Idempotent, dry-run parity, OSError surfaces with a clear line before re-raise. Catches three end states: fresh install (silent no-op), existing-install upgrade with orphan `home/config/goose-config.yaml` (logs file removal + summary), post-cleanup reinstall (silent no-op).

## Test plan

- [x] `make check` passes; lint clean.
- [x] `make test` passes (2829 tests, 1 skipped; +5 from main).
- [x] `TestRetireInstallHomeDir` (5 cases): missing-dir, empty-dir, orphan-goose-config-removal, dry-run preview, idempotency.
- [x] Updated `TestApplySource.test_copies_templates_config` and `test_dry_run_includes_templates_config` to assert the new destination.
- [x] Updated `TestApplyGooseConfig._setup` to seed the new path.
- [x] `grep -rn 'home/config' src/ tests/` returns only hits inside `_retire_install_home_dir`'s own docstring/comments (describing the retired path) and the new test class's narrative — no live path references.

## Operator cleanup note

On the next `make install` after this lands, install output will name `<install>/home/config/goose-config.yaml` removal plus `Removed retired <install>/home/; <install>/config/ holds the relocated content post-#452`. After the install completes, `<install>/home/` is gone for good.

### DIFF
--- a/src/kai/install.py
+++ b/src/kai/install.py
@@ -3073,7 +3073,7 @@ def _apply_source(install_path: Path, svc_uid: int, svc_gid: int, dry_run: bool)
         # live cleanup at the end of this function so the operator
         # sees what would happen on an existing-install upgrade
         # where `<install>/home/config/` still exists.
-        _retire_install_home_dir(install_path, dry_run=True)
+        _retire_install_home_dir(install_path, dry_run=dry_run)
         return
 
     _copy_tree(src_src, src_dst, _SOURCE_EXCLUDES)
@@ -3101,8 +3101,10 @@ def _apply_source(install_path: Path, svc_uid: int, svc_gid: int, dry_run: bool)
     # install nothing under it exists; on an existing-install upgrade
     # the cleanup above has already removed `.claude/` and `IDENTITY.md`,
     # and the goose-config relocation left an orphan `home/config/`.
-    # The helper handles both end states.
-    _retire_install_home_dir(install_path, dry_run=False)
+    # The helper handles both end states. `dry_run` is propagated
+    # rather than hardcoded so a future restructure of the early
+    # return above cannot silently suppress dry-run output here.
+    _retire_install_home_dir(install_path, dry_run=dry_run)
 
 
 def _apply_venv(install_path: Path, is_update: bool, dry_run: bool) -> None:
--- a/src/kai/install.py
+++ b/src/kai/install.py
@@ -2973,6 +2973,65 @@ def _retire_install_home_claude(install_path: Path, dry_run: bool) -> None:
             print(f"  Removed legacy {identity_md} ({size} bytes); content was identity baseline only")
 
 
+def _retire_install_home_dir(install_path: Path, dry_run: bool) -> None:
+    """
+    Remove the `<install>/home/` directory after its contents are gone.
+
+    Pre-#452 this directory held three things:
+      - `home/.claude/` (retired by `_retire_install_home_claude` per #447)
+      - `home/IDENTITY.md` (also retired by `_retire_install_home_claude`)
+      - `home/config/goose-config.yaml` (relocated to `<install>/config/` by #452)
+
+    With all three retired/relocated, `<install>/home/` is vestigial and
+    can be removed. The cleanup catches both end states:
+      - On a clean install: the directory never existed; silent no-op.
+      - On an existing-install upgrade: the `_apply_source` flow above
+        has already retired `.claude/`/`IDENTITY.md` and copied the
+        new `<install>/config/`, leaving an orphan `home/config/`
+        whose only file is `goose-config.yaml`. Remove that orphan
+        and the empty `home/` parent.
+
+    `home/config/goose-config.yaml` is install-time scaffolding (input
+    to `_apply_goose_config`, which writes the runtime config to
+    `~/.config/goose/config.yaml`); operators do not edit it. Logging
+    the removal with byte size is symmetric with
+    `_retire_install_home_claude` and consistent with the established
+    install-time loud-removal contract.
+    """
+    home_dir = install_path / "home"
+    if not home_dir.exists():
+        return
+
+    # Enumerate any surviving files first so the log captures byte
+    # sizes before rmtree runs. The only thing we expect to find is
+    # `home/config/goose-config.yaml` on the existing-install upgrade
+    # path; rglob handles arbitrary content defensively in case a
+    # future code change leaves something else behind.
+    for path in sorted(home_dir.rglob("*")):
+        if path.is_file() or path.is_symlink():
+            try:
+                size = path.lstat().st_size
+            except OSError:
+                size = 0
+            if dry_run:
+                print(f"[DRY RUN] Would remove {path} ({size} bytes)")
+            else:
+                print(f"  Removing {path} ({size} bytes)")
+
+    if dry_run:
+        print(f"[DRY RUN] Would remove retired {home_dir}; <install>/config/ holds the relocated content post-#452")
+    else:
+        # Same try/except discipline as the rmtree in
+        # `_retire_install_home_claude`: surface an operator-readable
+        # error line BEFORE the traceback aborts the install.
+        try:
+            shutil.rmtree(home_dir)
+        except OSError as exc:
+            print(f"  ERROR: could not remove {home_dir}: {exc}")
+            raise
+        print(f"  Removed retired {home_dir}; <install>/config/ holds the relocated content post-#452")
+
+
 def _apply_source(install_path: Path, svc_uid: int, svc_gid: int, dry_run: bool) -> None:
     """Copy source tree and home config from PROJECT_ROOT to the install location."""
     src_src = PROJECT_ROOT / "src"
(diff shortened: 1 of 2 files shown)


dcellison:
## Review by Kai

The PR is well-structured and the test coverage is solid. A couple of findings:

**Warning: PR/issue number references in production comments**

`install.py` (new function, updated comments) repeatedly embeds `#452`, `#447`, `post-#452` directly in production code comments:

- Docstring line: `relocated to <install>/config/ by #452`
- Comment: `# Issue #452: dry-run preview for the empty-home cleanup.`
- Comment: `# Issue #452: retire the now-empty <install>/home/ directory.`
- Summary print strings: `"<install>/config/ holds the relocated content post-#452"`

The global convention (CLAUDE.md) is explicit: "Don't reference the current task, fix, or callers... since those belong in the PR description and rot as the codebase evolves." These will be meaningless noise to someone reading the code in two years without repo history context. The print strings are especially durable - they're operator-visible output that will outlive the PR by a long shot.

**Suggestion: Misleading comment about `mkdir`**

In `_apply_source` around the updated config copy block:

```python
# Pre-#452 this lived under `<install>/home/config/`; the
# parent mkdir below creates the new top-level location.
config_dst.parent.mkdir(parents=True, exist_ok=True)
```

`config_dst` is `install_path / "config"`, so `config_dst.parent` is `install_path` - a directory that already exists at this point in the install flow. The `mkdir` here is essentially a no-op. The actual `install_path / "config"` directory is created by `_copy_tree`, not this `mkdir`. The comment claims the opposite and will mislead a future reader.

**Suggestion: Defensive deletion with no unexpected-content guard**

`_retire_install_home_dir` uses `rglob("*")` and logs all files before calling `shutil.rmtree`. Per the docstring this is intentional ("handles arbitrary content defensively"). However, if something unexpected lands in `home/` due to a future code change or operator mistake, the function logs and silently destroys it without any warning severity distinction. Given this runs as root during install, a single unexpected-file warning or a count check (`> 1 file found, expected 1`) would make the "defensive" behavior safer. Low priority since the install is supervised, but worth noting.

**No bugs, no security issues.** The call ordering (copy to new location, then retire old) is correct. Dry-run parity is properly maintained. The test suite covers all five stated cases cleanly.

dcellison:
## Review by Kai

Both prior review issues are addressed: issue-number references are gone from production code and comments, and the misleading `mkdir` comment is replaced with an accurate one. The new code is clean. One minor finding:

**Suggestion: hardcoded `dry_run=False` in the non-dry-run tail of `_apply_source`**

`install.py`, line added after `_copy_tree(config_src, config_dst)` block:
```python
_retire_install_home_dir(install_path, dry_run=False)
```

The hardcoded `False` is safe today because the early `return` in the `if dry_run:` block guarantees this line is never reached during a dry run. But it creates a silent maintenance trap: if someone restructures the early-return logic later (or adds a code path after this call), the hardcoded `False` will silently suppress dry-run output without any obvious signal. Passing `dry_run=dry_run` directly would make the intent explicit and match how every other helper in this file is called.

---

- Call ordering is right: new `config/` is written before `home/` is retired, so the orphan is removed last.
- The five test cases in `TestRetireInstallHomeDir` cover all three documented end states plus dry-run parity and idempotency.
- The `_apply_goose_config` source path update is consistent with `_apply_source`.
- No security issues; `shutil.rmtree` is scoped to a fixed sub-path of `install_path`.
- Error handling in `_retire_install_home_dir` matches the established pattern in `_retire_install_home_claude`.

dcellison:
## Review by Kai

This is a clean PR. Both critical prior-review issues (issue-number references in production code, hardcoded `dry_run=False`) are fully resolved. One finding remains:

**Suggestion: docstring says "both end states" but there are three**

`install.py`, `_retire_install_home_dir` docstring:

```
The cleanup catches both end states:
  - Clean install: the directory never existed; silent no-op.
  - Existing-install upgrade: an orphan `home/config/goose-config.yaml`
    remains; ...
```

The code and tests cover three states - missing dir (no-op), empty dir (removed with summary), and orphan-file upgrade. The PR description names all three correctly; the docstring's "both" misses the empty-dir case, which is a legitimate path (e.g., a prior run already cleaned the orphan but left the empty parent). Not a bug, but a future reader will wonder why the test has an `empty-dir` case that the docstring doesn't mention.

---

Everything else checks out:

- Call ordering is correct: `config_dst` is populated before `_retire_install_home_dir` removes `home/`.
- `dry_run=dry_run` is propagated consistently in both the early-return path and the live path.
- The `config_dst.parent.mkdir` no-op is gone; the replacement comment accurately describes `_copy_tree` as the directory creator.
- No PR/issue references survive in production code or operator-visible strings.
- `shutil.rmtree` is scoped to `install_path / "home"` - no path traversal risk.
- All five test cases in `TestRetireInstallHomeDir` match the documented end states plus dry-run parity and idempotency.
- Conventions: no em dashes, comments explain the why, no invented flags.

dcellison:
## Review by Kai

Looking at the current diff after the prior review's suggestions were addressed, this is a clean PR. One finding persists from the prior review thread:

**Suggestion: docstring enumeration still undercounts**

`install.py`, `_retire_install_home_dir` docstring:

```
The cleanup catches three end states:
  - Clean install: the directory never existed; silent no-op.
  - Existing-install upgrade with only orphan `home/config/goose-config.yaml`:
    remove the orphan and the empty `home/` parent.
  - Existing-install upgrade with extra content...: refuse, log, return.
```

The prior review flagged "both" as wrong because there are three states (missing, empty-dir, orphan). The fix changed "both" to "three" and added the refusal case as the third, but swapped out the empty-dir bullet rather than adding it. There are actually four distinct states in the code and tests: missing (no-op), empty-dir (removes cleanly), orphan-only (logs file then removes), and extra-content (refuses). `test_removes_empty_home_dir` covers a case the docstring still doesn't name. State 2 says "with only orphan...present" which implies the orphan must exist, leaving an unannotated code path.

**Suggestion: test class docstring is stale**

`tests/test_install.py`, `TestRetireInstallHomeDir` class docstring:

```
Three cases: directory missing (no-op), empty parent (clean removal),
and the existing-install upgrade where only the orphaned
`home/config/goose-config.yaml` remains.
```

The class now has 9 test methods including four refusal/dry-run-refusal cases. "Three cases" no longer describes the suite.

---

No bugs, no security issues. Call ordering is correct (new `config/` written before `home/` is retired). The `home_dir / "config"` non-directory edge case (a file named `config` bypasses the unexpected-content guard) is a very unlikely scenario and acceptable at this priority level. Dry-run propagation is correct in both the early-return and live paths.

Issue #454, opened 2026-05-11
bug: shared /tmp/claude-settings cache collides across OS users, blocks inner Claude spawn


dcellison:
## Summary

When the inner Claude is spawned as a different OS user via `sudo -u <target> -- claude` than the bot's service OS user, claude exits before producing any stdout. The bot logs `Claude process EOF`. The underlying error appears only on the subprocess's stderr (suppressed at the default INFO log level):

```
Error processing settings: EACCES: permission denied, open '/tmp/claude-settings-<hex>.json'
```

The `/tmp/claude-settings-<hex>.json` filename is content-derived from the `--settings` JSON. The bot passes a stable settings JSON on every spawn, so all invocations land on the same path. The first OS user to spawn the binary owns the file at mode `0644`; every subsequent caller as a different OS user fails to write it and the binary exits silently.

## Reproduction

1. Run the bot as service OS user A (the default `os_user`-not-set case).
2. Send a message; A's claude writes `/tmp/claude-settings-<hex>.json` (owner A, mode `0644`).
3. Add `os_user: B` to users.yaml for a chat, where B is a different OS user. Reinstall.
4. Send a message from that chat; B's claude tries to write the same path, gets `EACCES`, exits with no stdout.

The bot logs only:

```
kai.claude INFO Starting persistent Claude process (model=..., user=B)
kai.claude ERROR Claude process EOF
```

The user sees `Error: Claude process ended unexpectedly` on Telegram.

## Why this is surfacing now

Two things had to converge today:

1. The inner Claude only started running as a different OS user from the bot today, after the operator added `os_user` to their own chat's users.yaml entry. Before that, every spawn ran as the service user and the file owner always matched the next caller.
2. The stale service-user-owned `/tmp/claude-settings-<hex>.json` from prior same-user runs persisted in `/tmp`, sitting at exactly the hash the bot's stable `--settings` JSON resolves to.

A multi-user install with more than one distinct `os_user` would hit the same collision between any two `os_user` values whose claude invocations share the same settings JSON. The two-user case in `users.yaml` today is wired exactly that way.

## Today's manual unstick

```
rm /tmp/claude-settings-<hex>.json
```

The next spawn as the new OS user recreates the file with that user as the owner. Unblocks until another OS user spawns claude with the same settings.

## Suggested mitigation options

Open for design discussion; the bug is in upstream Claude Code's choice of `/tmp` for a content-hashed cache file, but `kai` can defend itself:

[...]

PR #455, opened 2026-05-11
fix: unstick inner Claude spawn when target os_user differs from service user

dcellison:
## Summary

Two install-introduced bugs that together blocked the inner Claude subprocess from spawning when a chat's `os_user` in users.yaml is a different OS user than the bot's service user. Surfaced today on a fresh install configured with a non-service `os_user`.

1. **`install.py:_generate_sudoers`** pinned the rule's `claude` binary path via `shutil.which("claude")`, which resolved against whatever PATH root happened to have at install time. When `sudo make install` was launched from a shell with another user's `~/.local/bin` on PATH, the wrong binary path got baked into the sudoers rule. The bot's runtime call is `sudo -u <target> -- claude`, which sudo resolves against the caller's (service user's) PATH; the resolved path did not match the rule's path, sudo demanded a password, and every message failed with `Failed to write to Claude process: Connection lost`.

   **Fix:** always anchor `claude_bin` to `{service_user_home}/.local/bin/claude`. Drop the `shutil.which` short-circuit. The bot's runtime spawn always resolves bare `claude` from the service user's PATH, so the rule must reference the service user's binary regardless of what was on PATH at install time.

2. **`backend.ensure_user_home`** chmod'd the per-user home directory to `0o755` on every instance creation. After `install.py _apply_migrate` chowned that directory to the user's `os_user`, the bot (running as the service user) hit `EPERM` on chmod -- macOS denies chmod by non-owner non-root callers even when the target mode equals the current mode, so the chmod-as-idempotent-no-op trick is not safe in a multi-user install. The exception propagated up through `pool.get` -> `_create_instance`, so the bot could not even reach the spawn for that chat and every message died at instance creation.

   **Fix:** wrap the chmod in `try/except OSError`. The umask defense (force `0o755` even under a hostile `umask 0o027`) only matters when *this* call creates the directory; a pre-existing dir with install-time per-user ownership is the install's responsibility to set the mode on. The sibling bootstraps (`ensure_user_memory`, `ensure_user_preferences`) already use the same `OSError`-swallow pattern.

Related: #454 (`/tmp/claude-settings-<hex>.json` collision across OS users sharing the same `--settings` JSON) is the third leg of today's failure chain and needs its own design pass; not in scope here.

## Changes

- `src/kai/install.py` -- drop `shutil.which("claude")`; anchor `claude_bin` to `{service_user_home}/.local/bin/claude`.
- `src/kai/backend.py` -- wrap `ensure_user_home`'s `os.chmod` in `try/except OSError` with a comment explaining the multi-user-install rationale.
- `tests/test_install.py::TestGenerateSudoers` -- existing tests that monkeypatched `shutil.which("claude")` updated to drop the patch and assert the service-user-home-anchored path. New regression test `test_claude_bin_ignores_caller_path` pins the new behavior: even when `shutil.which` returns a non-service-user path, the rule must NOT use it.
- `tests/test_backend.py::TestEnsureUserHome::test_chmod_eperm_does_not_crash` -- new regression test patching `os.chmod` to raise `PermissionError` only on the per-user dir. Without the swallow, the test fails with `PermissionError` escaping `ensure_user_home`.

## Test plan

- [x] `make check` clean.
- [x] `make test` passes (2834 tests, 1 skipped; +1 net new vs main).
- [x] Live verified end-to-end: bot spawns inner Claude as the operator's distinct `os_user`, Telegram message round-trips successfully.

## Diagnostic note (out of scope, captured for follow-up)

The original investigation was blocked by the inner Claude's stderr being drained at `DEBUG`, below the default `INFO` threshold. A separate follow-up to surface that drain at `WARNING` (or rate-limited `INFO`) would make the next class of crash-on-spawn self-diagnosing without needing a temporary code patch + reinstall to see the cause.

### DIFF
--- a/src/kai/backend.py
+++ b/src/kai/backend.py
@@ -621,8 +621,20 @@ def ensure_user_home(chat_id: int | None, data_dir: Path) -> Path:
     # instead of 0o755. That blocks group traversal and can break the
     # inner subprocess when sudo -u targets a different identity. Force
     # the intended bits explicitly - this is the same pattern install.py
-    # `_apply_migrate` uses after its mkdir calls. Idempotent on reuse.
-    os.chmod(path, 0o755)
+    # `_apply_migrate` uses after its mkdir calls.
+    #
+    # Swallow OSError because in a multi-user install the directory has
+    # already been chowned to the user's os_user (different from the
+    # service user) by install.py `_apply_migrate`, and macOS chmod(2)
+    # returns EPERM for any non-owner non-root caller even when the
+    # target mode equals the current mode. The umask defense only
+    # matters when *this* call creates the directory; a pre-existing
+    # dir is the install's responsibility to set perms on, and the
+    # sibling bootstraps below already use the same swallow pattern.
+    try:
+        os.chmod(path, 0o755)
+    except OSError:
+        pass
 
     # Lazy bootstrap of `<home>/.claude/CLAUDE.md`. Parallel to
     # ensure_user_memory and ensure_user_preferences: stat, possible
--- a/src/kai/backend.py
+++ b/src/kai/backend.py
@@ -623,17 +623,18 @@ def ensure_user_home(chat_id: int | None, data_dir: Path) -> Path:
     # the intended bits explicitly - this is the same pattern install.py
     # `_apply_migrate` uses after its mkdir calls.
     #
-    # Swallow OSError because in a multi-user install the directory has
-    # already been chowned to the user's os_user (different from the
-    # service user) by install.py `_apply_migrate`, and macOS chmod(2)
-    # returns EPERM for any non-owner non-root caller even when the
-    # target mode equals the current mode. The umask defense only
-    # matters when *this* call creates the directory; a pre-existing
-    # dir is the install's responsibility to set perms on, and the
-    # sibling bootstraps below already use the same swallow pattern.
+    # Swallow PermissionError because in a multi-user install the
+    # directory has already been chowned to the user's os_user
+    # (different from the service user) by install.py `_apply_migrate`,
+    # and macOS chmod(2) returns EPERM for any non-owner non-root caller
+    # even when the target mode equals the current mode. The umask
+    # defense only matters when *this* call creates the directory; a
+    # pre-existing dir is the install's responsibility to set perms on.
+    # Narrow to PermissionError rather than bare OSError so unrelated
+    # failure modes (ENOENT, ENOTDIR, EIO) still propagate.
     try:
         os.chmod(path, 0o755)
-    except OSError:
+    except PermissionError:
         pass
 
     # Lazy bootstrap of `<home>/.claude/CLAUDE.md`. Parallel to
--- a/src/kai/install.py
+++ b/src/kai/install.py
@@ -1657,11 +1657,18 @@ def _generate_sudoers(
         target_users.append(candidate)
 
     if target_users:
-        # Resolve the actual binary location; fall back to the native installer's
-        # default path under the service user's home if claude is not on PATH
-        # (e.g., when running under sudo with a stripped environment).
+        # Anchor the rule path to the SERVICE user's claude install. The
+        # bot spawns `sudo -u <target> -- claude` and sudo resolves the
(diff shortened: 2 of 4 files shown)


dcellison:
## Review by Kai

**Overall: clean and well-targeted.** Both fixes are correct for the described bugs. A few findings below.

---

### Warning

**`src/kai/install.py` — hardcoded path silently breaks non-standard Claude installs**

`claude_bin = f"{svc_home}/.local/bin/claude"` assumes Claude is always installed by the native installer at `~/.local/bin/claude`. If the service user has Claude at a different path (system Homebrew, npm global, `pipx`, etc.), the generated sudoers rule points to a nonexistent binary. `sudo` will then reject the attempt entirely and the bot fails — same symptom as before, but now at runtime with no obvious cause.

The old `shutil.which` was wrong because it resolved against the wrong user's PATH at install time. But dropping it entirely with no existence check means the path might not exist at all. Consider at minimum warning during install if `svc_home/.local/bin/claude` doesn't exist:

```python
claude_bin = f"{svc_home}/.local/bin/claude"
if not Path(claude_bin).exists():
    print(f"Warning: {claude_bin} not found; sudoers rule may be incorrect", file=sys.stderr)
```

This is a pre-existing installation assumption, but the PR makes it the *only* path — worth flagging.

---

### Suggestion

**`src/kai/backend.py:632` — `except OSError` is broader than necessary**

The comment explicitly names EPERM/EACCES as the target case, but the bare `except OSError` also swallows ENOENT, ENOTDIR, EIO, etc. Since the sibling bootstraps use the same pattern (per the PR), consistency is a valid argument — but `except PermissionError` would be tighter and make the intent unmistakable. `PermissionError` is a subclass of `OSError` so it won't miss the EPERM case.

---

### Suggestion (convention)

**Multi-paragraph docstrings in tests violate the project conventions**

`test_chmod_eperm_does_not_crash` (`tests/test_backend.py:808`), `test_claude_user_rule_anchored_to_service_user_home`, and `test_claude_bin_ignores_caller_path` (`tests/test_install.py:350`, `362`) all have multi-paragraph docstrings. The conventions say: *"Never write multi-paragraph docstrings or multi-line comment blocks — one short line max."* The PR description already captures the regression story well; the docstrings can be trimmed to single lines.

---

### Note (not a bug, just confirmation)

The `shutil.which` monkeypatch in `test_claude_bin_ignores_caller_path` is now dead from a production-code-flow standpoint (the function no longer calls `shutil.which`), but it's purposeful — it demonstrates that even if the environment would have answered differently, the result is still anchored. The assertion is what counts, and it's correct.

dcellison:
## Review by Kai

**Overall: the two prior-review concerns were addressed correctly.** `PermissionError` narrowing is in place, and the backstop warning was added to `_apply_sudoers`. One new convention finding below.

---

### Suggestion

**Multi-line comment blocks in production code violate conventions**

The prior review called out multi-paragraph docstrings in the *tests*, but the new production code has the same problem. The conventions say "Never write multi-paragraph docstrings or multi-line comment blocks — one short line max."

- `src/kai/backend.py` — the new comment above `try: os.chmod(...)` runs 9 lines
- `src/kai/install.py` `_generate_sudoers` — the replacement comment is 11 lines
- `src/kai/install.py` `_apply_sudoers` — the backstop comment is 8 lines

Each of these could be a single line (e.g., `# Swallow EPERM: macOS denies chmod to non-owners even in no-op mode; see PR #455.`) with the fuller explanation belonging in the commit or PR description where it already lives.

---

### Everything else looks clean

- The `PermissionError` narrowing over `OSError` is the right call — intent is precise, and `ENOENT`/`EIO` will still surface.
- The warning logic in `_apply_sudoers` correctly gates on `has_target_users` before checking path existence, and fires before the `dry_run` short-circuit so the test can verify it without a real sudoers write.
- The `shutil.which` monkeypatch in `test_claude_bin_ignores_caller_path` is intentionally dead from a production-code-flow perspective; the assertion is what matters and it's correct.
- Test coverage for the three new paths (missing binary warns, no target users stays silent, binary exists stays silent) is solid.
- No security issues: the sudoers path anchoring removes the caller-PATH injection vector that motivated the bug.

Issue #456, opened 2026-05-11
bug: inner Claude orphans on every cross-os-user spawn termination


dcellison:
## Summary

In cross-user spawn mode (chat has `os_user: <target>` distinct from the service user), `ClaudeCodeBackend._kill()` cannot terminate the inner Claude subprocess. The bot kills the `sudo` wrapper (which it can signal — sudo's real UID matches the service user), but the inner `claude` binary running as the target user survives and gets reparented to `init`. Every cross-user timeout or restart leaks one orphan `claude` process.

Observed today on the production install after switching the operator's chat to a non-service `os_user`. Two consecutive `Claude response timed out` events produced two long-lived orphans (24+ min and 8+ min respectively, both `ppid=1`, holding the target user's OAuth credentials and burning that user's claude.ai rate-limit budget).

## Root cause

`claude.py` spawns the subprocess as:

```
sudo -H -u <target> --preserve-env=KAI_WEBHOOK_SECRET -- claude ...
```

with `start_new_session=True` so sudo + claude share a new process group, and saves the sudo PID as `_pgid`.

`_kill()` calls `os.killpg(self._pgid, signal.SIGKILL)` to take down the whole group. POSIX signal permission rules (`signal(7)`):

> The real or effective user ID of the sending process must equal the real or saved set-user-ID of the target process.

- **sudo wrapper:** real UID = service user (kept across `setuid` from the bot's spawn), effective UID = root. Service user CAN signal it (real-UID match). ✅
- **claude binary (after sudo exec'd to target):** real, effective, and saved UIDs all = target user. Service user has no UID match → `kill(2)` returns `EPERM` → silently swallowed by the `except OSError` at `claude.py:369`. ❌

Net result: `killpg` reaps sudo but skips claude. sudo's death orphans claude (reparented to PID 1). The bot's read loop sees EOF on the stdout pipe (sudo's pipe died) and assumes the whole process tree is gone, but claude continues running detached.

The "Final cleanup" `killpg` at `claude.py:951-954` has the same permission gap — same `OSError` swallow.

## Impact

1. **Orphan accumulation.** Every timeout, idle eviction, restart, or `/new` leaks one claude process. They live until OOM, system reboot, or manual `sudo kill` by the operator.
2. **OAuth credential contention.** All orphans share `<target_home>/.claude/.credentials.json` and the target user's claude.ai session. Concurrent reads/writes may corrupt session state or trigger upstream rate-limit serialization.
[...]

PR #457, opened 2026-05-11
fix: per-os-user TMPDIR for inner Claude (closes #454)

### LINKED ISSUE #454: bug: shared /tmp/claude-settings cache collides across OS users, blocks inner Claude spawn

dcellison:
Closes #454.

## Summary

The `claude` binary writes its `--settings` JSON into a content-hashed temp file at `$TMPDIR/claude-settings-<hex>.json`. When multiple OS users on the same host spawn claude with identical `--settings` JSON (which the bot always does -- its settings string is a stable constant), they collide on the same path under the default `/tmp`. The first writer owns the file at mode 0o644 and every subsequent spawn as a different OS user exits at startup with:

```
Error processing settings: EACCES: permission denied, open '/tmp/claude-settings-<hex>.json'
```

The bot logs this as `Claude process EOF` (the stderr drain is at DEBUG, hidden at the default INFO level), and the user-facing error is `Error: Claude process ended unexpectedly`. Surfaced on the production install today after switching the operator's chat to a non-service `os_user`.

## Changes

**Install (`src/kai/install.py:_apply_migrate`)** -- new per-os-user temp-dir block, parallel to the existing memory / preferences / home / CLAUDE.md per-user blocks.

- Creates `<DATA_DIR>/tmp/<os_user>/` for every distinct `os_user` in users.yaml, mode `0o700`, chowned to the target user.
- Parent `<DATA_DIR>/tmp/` is service-owned mode `0o755` so the per-user subdirs remain traversable for the sudo wrapper.
- Reinstalls are idempotent: `mkdir(exist_ok=True)` + unconditional `chmod`/`chown` to correct drift. The `Created` log line is suppressed on the second pass.
- Dry-run prints the would-be paths and touches no disk.

**Runtime (`src/kai/claude.py:_ensure_started`)** -- TMPDIR injection.

- When `effective_claude_user` is non-None: `env["TMPDIR"] = str(DATA_DIR / "tmp" / effective_claude_user)`.
- `--preserve-env=KAI_WEBHOOK_SECRET` becomes `--preserve-env=KAI_WEBHOOK_SECRET,TMPDIR` so sudo passes it through. The sudoers SETENV: tag already in place from #455 covers this.
- Single-user mode (no claude_user, or claude_user == service user) is unchanged: the inner Claude inherits the system default `/tmp`. Safe because only one identity uses it; no cross-user collision possible.

## Tests

`tests/test_claude.py` -- updates and additions:

- Three existing tests had `cmd[4]` / `in cmd` assertions pinning the old `--preserve-env=KAI_WEBHOOK_SECRET`; migrated to the new comma-list value.
- `test_no_preserve_env_without_sudo` broadened to "no `--preserve-env=*` arg" so a future flag rename does not silently pass it.
- New `test_tmpdir_anchored_per_os_user_in_cross_user_mode`: pins the env-injection contract and the preserve-env list.
- New `test_tmpdir_not_injected_in_single_user_mode`: ensures the single-user path does not add TMPDIR to env.

`tests/test_install.py::TestApplyMigratePerOsUserTmpdir` -- six new tests:

- `test_creates_per_os_user_dir`: each distinct os_user gets a directory at `<DATA_DIR>/tmp/<os_user>/`.
- `test_per_os_user_dir_mode_is_0o700`: mode contract under a hostile `umask 0o077`.
- `test_no_dir_when_no_os_user`: a users.yaml with no `os_user` entries creates no tmp dirs.
- `test_duplicate_os_users_create_one_dir_each`: multiple chats with the same os_user collapse to one tmp dir.
- `test_idempotent_reinstall`: second install does not re-print `Created` for existing dirs.
- `test_dry_run_prints_without_creating`: dry-run lists paths without touching disk.

## Test plan

- [x] `make check` clean.
- [x] `make test` passes (2845 tests, 1 skipped; +8 net new vs main).
- [ ] Live verification post-merge: `sudo make install` creates the per-os-user tmp dirs; first cross-user spawn lands its `claude-settings` cache under `<DATA_DIR>/tmp/<os_user>/` owned by the target user; no EACCES at startup.

## Related

Companion fix to #456 (cross-user kill failure) will follow as a separate PR. Both bugs were introduced by today's switch to a non-service `os_user`; both are exposed by the cross-user spawn path that #455 unstuck.


dcellison:
## Review by Kai

The fix is correct and well-tested. Four findings:

---

**Warning — comment verbosity violates project conventions**
`src/kai/claude.py`, the two new comment blocks (~10 lines each around lines 258-270 and 305-316)

The conventions say "Default to writing no comments. Only add one when the WHY is non-obvious." The WHY here is legitimately non-obvious, but each block runs 10 lines and largely restates the PR description. A single-line `# per-user TMPDIR avoids /tmp hash collision (issue #454)` at each callsite captures the non-obvious invariant without the duplication. The current blocks also repeat each other nearly verbatim.

---

**Suggestion — unused `monkeypatch` fixture**
`tests/test_claude.py`, `test_tmpdir_anchored_per_os_user_in_cross_user_mode` (line ~332)

The fixture is declared as a parameter but never used in the test body. Remove it or it'll generate an unused-argument warning under strict linters.

---

**Suggestion — local import inside test function**
`tests/test_install.py`, inside `test_per_os_user_dir_mode_is_0o700`

```python
import stat
```

`stat` is a stdlib module imported mid-function. It should move to the module-level imports at the top of the file, consistent with every other import in the test file.

---

**Suggestion — no cleanup of stale per-user tmp dirs**
`src/kai/install.py`, the new `elif distinct_os_users:` block

If an `os_user` is removed from `users.yaml`, their `<DATA_DIR>/tmp/<os_user>/` directory (which holds the user's claude.ai session cache) persists indefinitely. The PR description notes a "companion fix" is coming; a cleanup pass here (or a warning message on orphaned dirs) would close the loop. Not urgent, but worth tracking — especially since the stale dir continues to hold session state for an identity that's been deprovisioned.

---

The core logic is correct: TMPDIR injection is gated on `effective_claude_user` (so single-user mode is untouched), the `--preserve-env` comma-list extension is right, mode 0o700 + explicit `chmod` correctly defeats umask, and the idempotency model (unconditional `chmod`/`chown`, suppressed `Created` on repeat) is sound.

dcellison:
## Review by Kai

Looking at this diff against the prior review thread, two issues were addressed (removed `monkeypatch` parameter, moved `import stat` to module level) and two were not (comment verbosity in `claude.py`, stale-dir cleanup). Per the review conventions, I'm not re-raising those.

Two new findings in code that didn't exist at the time of the prior review:

---

**Suggestion — local import inside test function**
`tests/test_claude.py`, `test_tmpdir_anchored_per_os_user_in_cross_user_mode`

```python
from kai.config import DATA_DIR as runtime_data_dir
```

Mid-function import, inconsistent with every other import in the file. Move to module level.

---

**Suggestion — hostile-umask choice doesn't stress-test the explicit `chmod`**
`tests/test_install.py`, `test_per_os_user_dir_mode_is_0o700`

The comment acknowledges the problem inline: `umask 0o077` strips group/other bits, but `0o700 & ~0o077 = 0o700` — those bits are already zero, so the umask doesn't interfere with the target mode at all. The test verifies the correct end-state (good), but doesn't demonstrate that the explicit `os.chmod(user_tmp, 0o700)` call is load-bearing. A umask of `0o100` (strips owner-execute) or `0o777` would actually force the chmod to do real work. Minor test-intent issue; the assertion itself is correct.

---

Everything else looks clean. The TMPDIR injection is correctly gated on `effective_claude_user` in both the env dict and the `--preserve-env` flag, the `tmp_root` parent gets mode `0o755` via explicit `chmod` (not just `mkdir`), and the idempotency model (`exist_ok=True` + unconditional `chmod`/`chown`) is sound.

dcellison:
## Review by Kai

Both prior-review findings from the second pass are addressed: `DATA_DIR` is now imported at module level (`tests/test_claude.py:28`), and the umask test was replaced with a spy/end-state pair that directly pins the `os.chmod` call.

Two new findings:

---

**Suggestion — stale sudo version floor in updated comment**
`src/kai/claude.py`, the updated comment block around lines 258-265

The comment still reads "Requires sudo >= 1.8.11 (2014)" but that version only supports `--preserve-env=VAR` with a single variable. The comma-separated `--preserve-env=A,B` form requires **sudo >= 1.8.21** (2017). Modern systems are unaffected, but the comment now understates the actual requirement introduced by this change.

---

**Suggestion — compound `assert` in `test_idempotent_reinstall`**
`tests/test_install.py`, the line reading:

```python
assert "Created" in first_output and "tmp/alice" in first_output
```

A single compound assert gives a less useful failure message than two separate assertions. Minor, but inconsistent with every other assertion in the file.

---

Otherwise the PR is clean. The core logic is correct: TMPDIR injection is properly gated on `effective_claude_user`, the `tmp_root` parent gets an explicit `chmod(0o755)` (not just relying on `mkdir`'s `mode=` argument), the new spy approach in `test_per_os_user_dir_mode_is_0o700` actually improves on what was suggested in the last review by directly pinning the `os.chmod` call rather than fighting umask semantics.

dcellison:
## Review by Kai

Both findings from the prior review round have been addressed: the sudo version floor comment now correctly states >= 1.8.21, and the compound assert in `test_idempotent_reinstall` is split into two separate assertions.

Looking at the new code for anything not previously raised:

**The PR is clean.** I checked the following and found no issues:

- **`install.py` logic**: `distinct_os_users` is built correctly before the `if dry_run / elif` branch; `pwd.getpwnam` calls are safe because the earlier `per_user_ids` loop already validates each `os_user` — a `KeyError` here would require a TOCTOU race. Idempotency model (`exist_ok=True` + unconditional `chmod`/`chown`) is sound. The dry-run path correctly calls `pwd.getpwnam` (for display output) but touches no disk.

- **`claude.py` injection**: `if effective_claude_user:` is the right gate — it's `None` in single-user mode, truthy in cross-user mode. `env["TMPDIR"]` is set before `create_subprocess_exec` and `--preserve-env=KAI_WEBHOOK_SECRET,TMPDIR` correctly threads it through sudo's env reset.

- **`test_per_os_user_dir_mode_is_0o700`**: The spy-plus-end-state approach is correct. `real_chmod = os.chmod` is captured before patching, so the actual syscall fires and the `stat()` assertion is meaningful. The `(str(user_tmp), 0o700) in chmod_calls` assertion correctly pins the explicit `os.chmod` call as load-bearing.

- **`test_tmpdir_not_injected_in_single_user_mode`**: `monkeypatch.delenv("TMPDIR", raising=False)` correctly eliminates ambient `TMPDIR` from the environment so any value in the subprocess env can only come from the code under test.

- **`test_no_preserve_env_without_sudo`**: The broadened `not any(str(arg).startswith("--preserve-env=") for arg in cmd)` assertion is strictly stronger than the old string match and future-proofs against flag renames.

- **Security**: No path traversal risk — `effective_claude_user` is OS-validated by `pwd.getpwnam` before being concatenated into the path, and OS usernames cannot contain `/`.

PR #458, opened 2026-05-11
fix: cross-os-user kill escalation via sudo (closes #456)

### LINKED ISSUE #456: bug: inner Claude orphans on every cross-os-user spawn termination

dcellison:
Closes #456.

## Summary

In cross-user spawn mode (chat has `os_user: <target>` distinct from the service user), `ClaudeCodeBackend._kill()` could not terminate the inner Claude subprocess. The bot killed the sudo wrapper -- which it can signal because sudo's real UID matches the service user -- but the inner `claude` binary running as the target user survived and got reparented to `init`. Every timeout, idle eviction, restart, or `/new` leaked one orphan claude process holding the target user's claude.ai session credentials and burning their rate-limit budget.

**Root cause.** POSIX signal permission rules (`signal(7)`) require the sender's real or effective UID to equal the target's real or saved set-user-ID. After sudo exec's claude as the target user, the inner process's UIDs are all the target's, with no match against the service user. `os.killpg` reaps the wrapper but skips the grandchild, and the bot's `except OSError: pass` in `_send_signal` swallowed the EPERM silently.

## Changes

**Install (`src/kai/install.py:_generate_sudoers`)** -- emits TWO rules per target user instead of one:

```
<svc_user> ALL=(<target>) SETENV: NOPASSWD: <svc_home>/.local/bin/claude
<svc_user> ALL=(<target>) NOPASSWD: /bin/kill
```

The kill rule does NOT carry `SETENV:` -- `kill` ignores env, and the narrower grant is the right surface. `shutil.which("kill")` with `/bin/kill` fallback handles the macOS (`/bin/kill`) and Linux (`/usr/bin/kill` or `/bin/kill`) location differences.

**Runtime (`src/kai/claude.py`)** -- new helper plus updates to the three signal sites:

- `_lookup_inner_claude_pid` runs `pgrep -P <sudo_pid>` to find the inner claude PID. sudo fork+execs claude as its sole child, so pgrep returns exactly one match. The result is cached on `self._inner_claude_pid`.
- `_send_signal` escalates via `sudo -n -u <target> kill -<sig> <pid>` BEFORE the existing `os.killpg` wrapper reap. The escalation is gated on `_effective_claude_user` so single-user mode bypasses the whole branch.
- `_kill` and `shutdown` prime the cache before the first signal goes out, so the post-killpg final-cleanup pass has a handle even after sudo dies (pgrep on a dead sudo returns nothing). The final cleanup issues one more sudo-kill against the cached PID as belt-and-suspenders.
- Two new attrs (`_inner_claude_pid`, `_effective_claude_user`) carry the state across the signal sequence; both are reset on spawn and cleared on `_kill`/`shutdown`.

**Back-compat.** An old install whose sudoers file lacks the new kill rule will see `sudo -n` fail (no password available), the `subprocess.run` uses `check=False` so the call is silent, and the existing orphan-leak behavior persists until the operator reinstalls. No new crash mode.

## Tests

`tests/test_install.py::TestGenerateSudoers` -- 4 new tests:

- `test_kill_rule_emitted_per_target`: each target gets a NOPASSWD kill rule alongside the claude rule.
- `test_kill_rule_omitted_when_no_targets`: empty `os_users` produces no kill rules.
- `test_kill_rule_no_setenv`: kill rule line does NOT carry `SETENV:`; claude rule still does.
- `test_kill_rule_path_is_hardcoded`: regression guard that even when `shutil.which("kill")` would return a different absolute path, the generated rule still says `/bin/kill`. The generator no longer calls `shutil.which` for this path; the test pins that invariant.

Two existing per-target tests had their `count("(name)")` assertions tightened to count claude rules specifically, because each target now contributes two rule-line occurrences of `(name)`.

`tests/test_claude.py::TestProcessSignals` -- 5 new tests for `_send_signal`: confirms the sudo kill is issued, the killpg wrapper reap still fires, the lookup is cached across multiple invocations, the single-user path bypasses the whole sudo branch, and TimeoutExpired/OSError on the sudo call do not propagate.

`tests/test_claude.py::TestKill` -- 4 new tests for `_kill` integration: cache priming before signal, final cleanup escalates via sudo with `-9`, skipped sudo when inner PID is unknown, full bypass in single-user mode.

`tests/test_claude.py::TestLookupInnerClaudePid` -- 7 new tests for the helper: None when `_proc` is None, parses pgrep stdout, pgrep nonzero exit, empty stdout (sudo not-yet-forked race), TimeoutExpired, OSError, defensive multi-line case.

## Test plan

- [x] `make check` clean.
- [x] `make test` passes (2865 tests, 1 skipped; +20 net new vs main).
- [ ] Live verification post-merge: `sudo make install` regenerates `/etc/sudoers.d/kai` with the per-target kill rules; next cross-user timeout reaps the inner claude cleanly with no orphan visible in `ps -o ppid= -p <pid>`.

## Related

Companion to #457 (per-os-user TMPDIR). Both were exposed by the cross-user spawn path that #455 unstuck on today's install.


dcellison:
## Review by Kai

I have enough from the diff to write the review directly.

---

## PR Review: fix/cross-os-user kill escalation via sudo

### Warning — `kill` path mismatch between sudoers rule and runtime command

**`install.py` (new kill rule)** resolves the binary at install time:
```python
kill_bin = shutil.which("kill") or "/bin/kill"
rules += f"{service_user} ALL=({target}) NOPASSWD: {kill_bin}\n"
```
On a Linux system where `shutil.which("kill")` returns `/usr/bin/kill`, the sudoers entry reads `NOPASSWD: /usr/bin/kill`.

**`claude.py:_send_signal()`** (and the final-cleanup block in `_kill`) invokes the command as:
```python
["sudo", "-n", "-u", self._effective_claude_user, "kill", f"-{sig}", ...]
```
`"kill"` here is unqualified. sudo resolves it at runtime via its `secure_path`. If sudo's secure_path resolves `kill` to `/bin/kill` while the rule names `/usr/bin/kill` (distinct inodes on older distros without the merged-usr layout), sudo exits 1 and the escalation silently falls through — reproducing the original orphan-process bug. The fix fails silently per design (`check=False`), leaving the operator with no indication of the mismatch until they dig through process tables.

**Fix:** pass `kill_bin` at runtime, or just hardcode `/bin/kill` in both places since that's already the fallback. The simplest option is to use `"kill"` only when `/bin/kill` is canonical (macOS, modern Debian/Ubuntu with merged `/bin`), or resolve the path once at module import time the same way `_generate_sudoers` does.

---

### Warning — blocking `subprocess.run` in the async event loop

`_lookup_inner_claude_pid` (`timeout=2`), `_send_signal`'s sudo call (`timeout=5`), and the final-cleanup blocks in `_kill`/`shutdown` (`timeout=5`) all use synchronous `subprocess.run`. In the worst case (system pathology where sudo or pgrep hangs), a single `_kill()` path can block the event loop for ~12 seconds before timeouts fire. `_send_signal` is called twice in the kill sequence (SIGKILL + final), so the ceiling is ~22 s of event-loop stall. The existing `os.killpg` is effectively instant; this is a new source of latency. Not a correctness bug for a personal bot, but worth tracking.

---

### Suggestion — convention: multi-paragraph docstring on `_lookup_inner_claude_pid`

The project conventions state "Never write multi-paragraph docstrings or multi-line comment blocks — one short line max." `_lookup_inner_claude_pid` has a 15-line docstring and several other changed methods carry 6–10 line comment blocks. The rationale is non-obvious enough to warrant some prose, but the per-function docstrings and the inline comment blocks in `_kill`/`shutdown` both exceed what the conventions call for. The cross-reference `(#456)` annotations are good; the prose blocks explaining POSIX signal semantics could be trimmed since that context lives in the PR description.

---

### Everything else looks sound

- The `_inner_claude_pid` cache-before-signal ordering is correct; `saved_inner_pid` is captured before `_lock` drop and before any signal fires.
- `check=False` throughout is the right choice — all error modes are intentionally non-fatal.
- `SETENV:` correctly absent from the kill rule (kill ignores env).
- The dedup assertions in `test_install.py` are correctly tightened to count SETENV-bearing lines specifically.
- The new `TestLookupInnerClaudePid` tests cover the important edge cases (nonzero exit, empty stdout, TimeoutExpired, OSError, multi-line output).
- PID reuse (TOCTOU between cache and final `kill -9`) is a known unavoidable race in PID-based process management; acceptable here.

dcellison:
## Review by Kai

This is a second-pass review; per the thread, the path-mismatch and async-blocking warnings from the first pass were acknowledged. I'll only surface new findings.

---

## PR Review: fix/cross-os-user kill escalation via sudo (second pass)

### Warning — sudoers kill rule is broader than the minimum necessary grant

`install.py` emits:
```
<svc_user> ALL=(<target>) NOPASSWD: /bin/kill
```

This lets the service user run `/bin/kill` as the target with **any arguments** — meaning it can signal any process owned by that target user, not only the claude grandchild. In practice the service user already holds the far broader grant of `NOPASSWD: <claude_bin>` (run arbitrary code as the target), so the delta is minimal, but it's worth a comment in the sudoers block explaining the intentional scope and the reason a PID-locked rule isn't feasible here. As written, an operator reviewing the file can't distinguish "this is a deliberate design tradeoff" from "this was overlooked."

---

### Suggestion — stale test name in PR description

The description lists `test_kill_rule_falls_back_to_bin_kill` as one of the four new `TestGenerateSudoers` tests, but the diff contains `test_kill_rule_path_is_hardcoded` with different semantics (it mocks `shutil.which` returning `/usr/bin/kill` and asserts the rule still says `/bin/kill`, because the code no longer calls `which` at all). The description was written for the earlier draft; the test name and rationale both changed after the first-pass review. Not a code defect, but it'll confuse anyone who reads the description to understand the test coverage.

---

- `/bin/kill` hardcoded in both `install.py` and `claude.py` — the path-mismatch regression guard (`test_kill_rule_path_is_hardcoded`) is sound.
- Cache-priming before `_send_signal` in both `_kill` and `shutdown` is ordered correctly; `saved_inner_pid` is captured before state is cleared.
- `saved_user`/`saved_inner_pid` local copies mean the final-cleanup blocks don't race against `self._inner_claude_pid = None`.
- `check=False` throughout is the right call; ESRCH from a double-kill on a reaped process is silently absorbed, consistent with the existing `os.killpg` pattern.
- New tests cover the important edge cases: cache reuse across signals, None-PID short-circuit, single-user bypass, TimeoutExpired/OSError swallowing.
- The kill rule correctly omits `SETENV:` and the regression test guards that invariant.

dcellison:
## Review by Kai

## PR Review: fix/cross-os-user kill escalation via sudo (third pass)

Per prior-review thread policy: not re-raising the blocking-subprocess-run or multi-line docstring/comment findings — both were surfaced and the author kept the code as-is.

---

### Warning — `shutdown` cross-user path has no test coverage

The PR adds cache-priming + final `sudo kill -9` logic to **both** `_kill` and `shutdown`, but the new tests cover only `_kill`. The `shutdown` changes at `claude.py` lines +1169–+1185 (priming block) and +1215–+1244 (final cleanup) are structurally identical to the `_kill` changes, yet the test suite adds zero cross-user cases to `TestShutdown`.

The `TestKill` suite covers:
- cache primed before signal
- final cleanup escalates via sudo with `-9`
- skipped when inner PID is unknown
- full bypass in single-user mode

`TestShutdown` gets none of those. A regression in the `shutdown` path (e.g., someone accidentally inverts the `saved_user is not None` guard, or clears `_inner_claude_pid` before `saved_inner_pid` is captured) would go undetected by CI.

---

### Suggestion — `test_kill_rule_path_is_hardcoded` mock describes removed behavior

The test patches `shutil.which("kill")` to return `/usr/bin/kill` and asserts the output still contains `/bin/kill`. But the production code no longer calls `shutil.which("kill")` at all — `kill_bin = "/bin/kill"` is a literal. The monkeypatch is a forward-looking regression guard (if someone re-introduces the `shutil.which` call, the test would catch the mismatch), which is fine, but the docstring says "the generator no longer calls `shutil.which` for this path; the test pins that invariant" without making the guard intent explicit in the mock setup itself. Consider a brief inline comment (`# guard: if shutil.which("kill") is ever re-introduced, this mock catches the mismatch`) so the intent survives a future reader wondering why a patched function that's never called is present.

---

- `/bin/kill` hardcoded in both `install.py` and `_send_signal`/`_kill`/`shutdown` — path-mismatch regression closed.
- `saved_user`/`saved_inner_pid` locals captured before state is cleared; final-cleanup blocks use only locals — no race against the `= None` assignments.
- `f"-{sig}"` produces `"-9"` / `"-15"` etc. from the `signal.Signals` IntEnum — valid POSIX numeric signal specs for `/bin/kill`.
- `check=False` throughout is right; ESRCH from double-killing a reaped process is silently absorbed, consistent with the existing `os.killpg` pattern.
- Kill rule correctly omits `SETENV:`; `test_kill_rule_no_setenv` guards the invariant with a clean `startswith("kai ALL=")` filter that already excludes the English comment block.
- The dedup assertions in `test_install.py` correctly switch from counting `(name)` occurrences to counting SETENV-bearing rule lines specifically.
- `TestLookupInnerClaudePid` covers the important edge cases (nonzero exit, empty stdout, `TimeoutExpired`, `OSError`, multi-line output).

dcellison:
## Review by Kai

## PR Review: fix/cross-os-user kill escalation via sudo (fourth pass)

Per the prior-review thread: not re-raising the blocking-subprocess-run, multi-line docstring/comment, sudoers scope, or stale test name findings. The `shutdown` coverage gap from pass 3 is now addressed with four matching tests.

---

### Warning — `f"-{sig}"` produces an invalid kill argument on Python < 3.11

**`claude.py`, `_send_signal`** — the subprocess command uses:

```python
f"-{sig}",
```

`sig` is typed as `int` but receives `signal.Signals` enum values (`signal.SIGKILL`, `signal.SIGTERM`). On Python < 3.11, `IntEnum.__format__` delegates to `Enum.__str__`, so `f"{signal.SIGTERM}"` produces `"Signals.SIGTERM"` rather than `"15"`. The resulting kill argument would be `"-Signals.SIGTERM"`, which `/bin/kill` rejects, silently falling through (`check=False`) and reproducing the orphan-leak the PR intends to close.

The two final-cleanup blocks (in `_kill` and `shutdown`) correctly use the hardcoded literal `"-9"`, so the inconsistency is visible in the same diff. The fix is `f"-{int(sig)}"` in `_send_signal`.

Python 3.10 is EOL as of October 2025, so this is unlikely to fire in production today, but the inconsistency is a latent trap if the project ever needs to backport.

---

### Suggestion — test assertion uses the same version-sensitive format expression

**`tests/test_claude.py`, `test_send_signal_cross_user_escalates_via_sudo`**:

```python
assert cmd[5] == f"-{signal.SIGKILL}"
```

On Python < 3.11, both the production code and this assertion produce `"-Signals.SIGKILL"`, so the test passes while the behaviour is wrong — the assertion doesn't independently guard the intended output. The literal `"-9"` would catch the regression on any Python version:

```python
assert cmd[5] == "-9"
```

---

- Four new `TestShutdown` cross-user tests fully mirror the `TestKill` coverage: cache priming, final sudo cleanup, no-PID skip, and single-user bypass. The pass-3 coverage gap is closed.
- `saved_user`/`saved_inner_pid` locals are captured before `_inner_claude_pid = None` in both `_kill` and `shutdown`; the final-cleanup blocks use only locals, no race against the clear.
- The `shutdown` priming uses the same `_effective_claude_user` gate as `_kill`; the test verifies ordering via `call_order` the same way.
- `/bin/kill` is hardcoded in both `install.py` and all three `subprocess.run` sites; `test_kill_rule_path_is_hardcoded` guards the invariant.
- `check=False` throughout is correct; ESRCH from a double-kill is silently absorbed.
- `test_shutdown_primes_inner_pid_cache_before_signaling` correctly returns `None` from `mock_run.side_effect` (the production code discards the return value).

dcellison:
## Review by Kai

## PR Review: fix/cross-os-user kill escalation via sudo (fifth pass)

Per prior-review thread policy: not re-raising the blocking-subprocess-run, multi-line docstring/comment, or sudoers scope findings — all surfaced and author kept the code as-is.

---

### The diff is clean

All findings from passes 1–4 are resolved:

- **`/bin/kill` hardcoded in both `install.py` and `claude.py`** — path-mismatch regression closed; `test_kill_rule_path_is_hardcoded` guards the invariant.
- **`f"-{int(sig)}"` in `_send_signal`** — Python < 3.11 IntEnum format bug fixed; the comment explains the contract.
- **Test assertion uses `"-9"` literal** (`cmd[5] == "-9"`) — independent of `IntEnum.__format__` version behavior.
- **Four `TestShutdown` cross-user tests** — mirror the `TestKill` coverage: cache priming, final sudo cleanup, no-PID skip, single-user bypass.
- **`saved_user`/`saved_inner_pid` locals** captured before `self._inner_claude_pid = None` in both `_kill` and `shutdown`; final-cleanup blocks use only locals — no race.

I re-checked `_lookup_inner_claude_pid`, the priming ordering, the `textwrap.dedent` block in `install.py`, and the cross-user test assertions for both `_kill` and `shutdown`. No new issues found.

**Approve.**

dcellison:
## Review by Kai

## PR Review: fix/cross-os-user kill escalation via sudo (sixth pass)

Per prior-review thread policy: not re-raising the blocking-subprocess-run, multi-line docstring/comment, sudoers scope, or stale test name findings.

---

### Suggestion — `saved_inner_pid` captured before `_send_signal` can populate the cache

**`claude.py`, `_kill`** (diff lines +1081–+1094) and **`shutdown`** (diff lines +1204–+1217):

In both methods, `saved_inner_pid` is captured from `self._inner_claude_pid` *before* `_send_signal` is called:

```python
if saved_user is not None and self._inner_claude_pid is None:
    self._inner_claude_pid = self._lookup_inner_claude_pid()
saved_inner_pid = self._inner_claude_pid   # captured here

self._send_signal(signal.SIGKILL)          # may update self._inner_claude_pid
```

`_send_signal` also has its own cache-population path:

```python
if self._inner_claude_pid is None:
    self._inner_claude_pid = self._lookup_inner_claude_pid()
```

In the narrow window where `_kill`'s priming lookup returns `None` (pgrep race: sudo spawned but hasn't forked yet) and `_send_signal`'s retry succeeds (a moment later), `self._inner_claude_pid` gets set to the real PID inside `_send_signal`. Then `self._inner_claude_pid = None` clears it, and the final-cleanup block skips because `saved_inner_pid is None` — even though `_send_signal` may have already found the PID and tried (possibly unsuccessfully) to kill it. The belt-and-suspenders pass has no handle.

`TestKill` doesn't cover this case: `_lookup_inner_claude_pid` is always patched to return a consistent value.

Fix: refresh `saved_inner_pid` after `_send_signal` returns, before state is cleared:

```python
self._send_signal(signal.SIGKILL)
if saved_inner_pid is None:
    saved_inner_pid = self._inner_claude_pid  # pick up any PID _send_signal cached
```

Same applies to `shutdown`'s SIGTERM path (and the SIGKILL escalation inside the `except asyncio.TimeoutError` block).

---

- `/bin/kill` hardcoded in both `install.py` and all three `subprocess.run` sites; the sudoers rule and runtime invocation match exactly.
- `int(sig)` in `f"-{int(sig)}"` is correct; the `"-9"` literal in the final-cleanup blocks and in the test assertion (`cmd[5] == "-9"`) is version-independent.
- Four `TestShutdown` cross-user tests mirror `TestKill` coverage; pass-3 gap is closed.
- `saved_user`/`saved_inner_pid` locals are captured before `self._inner_claude_pid = None`; final-cleanup blocks reference only locals — no race against the clear.
- `check=False` throughout is correct; ESRCH from a double-kill is silently absorbed.
- `test_kill_rule_no_setenv` and `test_kill_rule_path_is_hardcoded` guard the intended invariants; the forward-looking monkeypatch in the latter has an inline comment explaining its intent.

Issue #666, opened 2026-06-13
Review agent: full-context bundle for PR reviews


dcellison:
## Context

The PR review agent builds its prompt from PR metadata, the description, an optional linked-issue spec, project conventions, prior comments, and the diff. Diff-only review has known blind spots: contract consumers outside the diff (a changed log payload whose parser lives elsewhere), final-file interactions (a new branch made unreachable by an existing early return above the hunk), issue scope gaps, and tests that pin the new helper but not its production caller.

## Goal

Add a context-gathering step before the review model call that assembles a bundle: full PR metadata (commits, changed-file list, closing issue references, head SHA), linked issue bodies, commit messages, the full patch, full changed-file contents at the PR head, and targeted surrounding code found by symbol search, with a deterministic size budgeter and every untrusted section wrapped in the existing per-block boundary scheme.

## Design

The detailed design lives in two operator-side design notes: `pr-review-agent-context-bundle.md` (the bundle definition, collection commands, prompt template, budget ladder) and `pr-review-context-bundle-implementation.md` (grounding against the current review module, the checkout-vs-API decision for changed files, the budgeter shape, the outside-hits symbol-search filter, and the open output-sink question). A spec will precede implementation per the usual workflow; this issue is the tracking anchor.

## Acceptance sketch

- The review prompt contains full changed-file contents at the PR head, linked issue bodies, and commit messages, each in its own untrusted-data boundary block.
- Surrounding-code excerpts carry a reason for inclusion and prefer contract consumers outside the changed files.
- Context overflow degrades by the documented drop ladder with visible truncation notes; full changed files are dropped last.
- All existing prompt-injection boundaries and prior-comment behavior are preserved.

dcellison:
## Triage by Kai

**Triage summary:** Adds a pre-review context bundle for the PR review agent, including metadata, linked issue bodies, commits, full patches, changed-file contents, and targeted surrounding code under deterministic budgeting and existing untrusted-data boundaries. The scope is substantial and implementation-sensitive because it affects prompt construction, context sizing, and prompt-injection protections.

**Priority:** medium
**Labels applied:** (none added)

dcellison:
## Epic breakdown

Treating this issue as an epic; tracking implementation across two child issues:

1. #682 — Webhook PR review uses shared full-context bundle (phase 1).
2. #683 — Telegram `/review` manual command using shared bundle (phase 2; blocked by #682).

[...]

PR #685, opened 2026-06-15
Rebalance bundle caps for realistic kai file sizes

dcellison:
## Summary

- Manual acceptance pass against PR #681 showed the bundle's headline value-add (full changed-file contents at the PR head) did not actually reach the reviewer for either changed file: `tests/test_bot.py` (246K bytes) exceeded the per-file cap; `src/kai/bot.py` (180K bytes) exceeded the changed-files section cap. The reviewer ended up working from the patch + 11 related excerpts only, which is the pre-bundle review surface.
- Several kai modules sit in the same 150-220K-char range (`webhook.py`, `config.py`, `sessions.py`, `pool.py`, `review.py` post-#666), so this is not a one-off PR shape.
- Rebalance the per-file cap (200K -> 500K), per-section cap (90K -> 250K), and global ceiling (240K -> 360K) so realistic kai-shaped PRs carry full file contents through to the reviewer. The cross-section drop ladder added in the previous round still enforces the global cap; these numbers stop the ladder firing for the realistic case.

## Test plan

- [x] `make check` clean.
- [x] `make test`: 4476 passed (4475 baseline + 1 new), 1 skipped.
- [ ] Re-run the manual acceptance pass against the same PR (#681) after merge and confirm `bot.py` and `test_bot.py` both inline with no `[CHANGED_FILES_AT_HEAD]` budget notes (deferred to post-merge).

Refs #666

### DIFF
--- a/src/kai/review.py
+++ b/src/kai/review.py
@@ -83,18 +83,23 @@ _MAX_PRIOR_COMMENTS_CHARS = 50_000
 # ── Bundle budget constants ──────────────────────────────────────────
 #
 # Per-section caps used by the deterministic budgeter that backs
-# build_pr_review_context() / build_review_prompt_from_context(). These
-# are v1 guardrails sized to keep the rendered prompt within the
-# common-case context window of every configured review backend; the
-# numbers are not magic-correct values and can be tuned later. The
-# drop ladder (see _apply_budget_ladder) decides the order in which
-# sections are reduced when the rendered prompt would otherwise
-# exceed _MAX_REVIEW_CONTEXT_CHARS. Backend dispatch is symmetric;
-# per-backend ceilings are only introduced if a configured backend
-# cannot fit the shared one.
-_MAX_REVIEW_CONTEXT_CHARS = 240_000
+# build_pr_review_context() / build_review_prompt_from_context(). The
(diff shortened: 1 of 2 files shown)


dcellison:
## Review by Kai

I’m checking the production budgeting path, not just the fetcher path, because this PR changes caps whose behavior only shows up after `budget_review_context()` runs.

The workspace path doesn’t contain the repository checkout, so I’m relying on the supplied patch and full-file excerpts for line references. The issue I’m seeing is in the cap interaction itself, which is visible from the provided `review.py` content.

Warning: `src/kai/review.py:103` still makes the stated acceptance case impossible.

The PR says the real target is a PR touching both `src/kai/bot.py` at ~180K and `tests/test_bot.py` at ~246K with no `[CHANGED_FILES_AT_HEAD]` budget notes, but `_MAX_CHANGED_FILES_CHARS = 250_000` cannot hold both. `_cap_changed_files()` measures the combined rendered size, sees ~426K+, and drops the largest file content to a note before the prompt is rendered. The new test only exercises `fetch_changed_files_at_head()` for a single 180K file, so it misses the production budgeter path where the real failure occurs.

Fix direction: add a budget/render test with both realistic changed files and size the section/global caps so that case actually survives, or adjust the PR goal to only promise one realistic file at a time.

PR #687, opened 2026-06-15
Size review caps to hold both files in a realistic kai PR

dcellison:
## Summary

- The production webhook bot's first bundle review on PR #685 caught a real gap I missed: the prior cap rebalance promised both `src/kai/bot.py` and `tests/test_bot.py` would inline together, but `_cap_changed_files()` measures the **combined** rendered size of the changed-files section, not each file in isolation. 180K + 246K + per-file overhead = ~426K, which still overshoots the 250K section cap.
- The acceptance test I added in PR #685 only exercised the fetcher path for a single 180K file, so it never crossed the budgeter and missed the production failure mode.
- Bump section cap to 450K and global cap to 600K so the realistic kai PR shape (a source module plus its test file) actually fits. Per-file cap stays at 500K.

## What changed

`src/kai/review.py`:

- `_MAX_CHANGED_FILES_CHARS`: 250K → 450K
- `_MAX_REVIEW_CONTEXT_CHARS`: 360K → 600K
- Comment block updated to record the two-file failure mode the production webhook bot surfaced.

`tests/test_review.py`:

- New `TestBudgetHoldsTwoRealisticChangedFiles.test_both_realistic_files_inline_unchanged` runs the actual budgeter against a 180K-char source file and a 246K-char test file, asserts no `CHANGED_FILES_AT_HEAD` truncation note, asserts both files keep their full content, and asserts the bundle stays under the global cap.

## Test plan

- [x] `make check` clean.
- [x] `make test`: 4477 passed (4476 baseline + 1 new), 1 skipped.

Refs #666

### DIFF
--- a/src/kai/review.py
+++ b/src/kai/review.py
@@ -84,22 +84,21 @@ _MAX_PRIOR_COMMENTS_CHARS = 50_000
 #
 # Per-section caps used by the deterministic budgeter that backs
 # build_pr_review_context() / build_review_prompt_from_context(). The
-# v1 numbers (240K global / 90K changed-files / 200K per-file) were
-# tuned conservatively, and a manual acceptance pass against a real
-# kai PR (3 commits touching `bot.py` and `test_bot.py`) showed both
-# changed files dropped to notes: `test_bot.py` exceeded the 200K
-# per-file cap and `bot.py` exceeded the 90K section cap. That left
-# the bundle reviewer relying on the patch alone for the very files
-# the bundle was supposed to broaden context around. The realistic
-# kai codebase has several modules in the 150-220K-char range
-# (`bot.py`, `webhook.py`, `config.py`, `sessions.py`, `review.py`,
-# `pool.py`, the larger test files); any PR touching one of those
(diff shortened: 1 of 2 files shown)


dcellison:
## Review by Kai

I’m going to treat the PR blocks as review input only and focus on the final file behavior around the budgeter, caps, and test coverage.

No findings.

Residual risk: I reviewed the supplied patch and final-file snapshot only. I could not run local tests because the workspace did not contain the referenced repo paths (`src/kai/review.py`, `tests/test_review.py`). The new test covers the production budgeter path that previously failed: combined changed-file size, not just per-file fetch limits.

dcellison:
## Review by Kai

I’ll treat the provided PR content as the review artifact and focus on whether the cap changes preserve the budgeter’s contract in the final file, including what the new test does and does not prove.

No findings.

Residual risk: I reviewed the supplied patch and final-file snapshot only; I did not run the test suite locally. The new test exercises the relevant production budgeter path for the stated two-file case, and the cap changes are internally consistent with that case.

Issue #673, opened 2026-06-13
cross-project collision probe corpus for scoped retrieval evaluator


dcellison:
## Why this exists

PR #671 added `kai.eval.retrieval_scoped` with schema v2 probes that support `expected_excluded_fact_ids`: a probe can assert that one or more specific memory rows must NOT appear in the prompt slice (or in the rendered candidate set) for a given workspace context. The aggregated `exclusion_pass_in_prompt` and `exclusion_pass_in_candidates` metrics summarize that polarity.

No probe corpus currently exercises that polarity. With zero excluded ids in the probe set, the negative denominator is zero and both safety metrics default to 1.0 by harness convention. The safety signal is invisible until probes deliberately encode cross-project collisions.

The corpus authoring is operator data work, not code work; this issue tracks it because it is a prerequisite for #672 (the legacy-vs-scoped retirement decision needs evidence on the exclusion-safety axis). Without the corpus, #672 cannot conclude.

## What "cross-project collision" means

A collision probe pins a memory row owned by project A and asks a question from project B's workspace whose embedding overlaps the project-A row. The scoped pipeline must filter the project-A row out before ranking (project scope admission). Without scope filtering, legacy retrieval would surface it because the embedding is close.

The collision pattern that matters most is vocabulary overlap on a project name, file path, identifier, or technical term that exists in both projects but means different things. Examples:

- The string "memory" appears in this repo (`kai.memory`), in operator notes about reasoning memory, and possibly in another project (a personal note app, an LLM-research project, etc.). A probe authored from project B asking "what does memory mean here" must NOT surface project A's `kai.memory` fact.
- A directory name like `src/` exists in every project. A probe from project A asking "where does the entrypoint live" must NOT surface project B's directory layout fact.
- A package or library name that one project depends on and another does not.

The corpus exercises the safety property by construction, not by accident.

## Corpus shape

The corpus is a JSONL probe file under the schema v2 the scoped evaluator already accepts. Minimum coverage:

- **Per active project**, at least 5 cross-project collision probes:
  - Workspace set to project A's path.
  - Question semantically close to a known project-B fact (verified by running the existing legacy harness or by `/memory tag` inspection to confirm the project-B fact would surface).
  - `expected_excluded_fact_ids` lists the project-B row id (or row ids).
  - Optional `expected_fact_id` if a project-A row should ALSO surface.
- **Per active project**, at least 2 positive-only probes asserting the project's own facts surface correctly:
  - Workspace set to the project's path.
[...]

PR #691, opened 2026-06-15
Add legacy_retrieve_hits and embed_texts public helpers

dcellison:
## Summary

Adds two public helpers the collision-probe corpus generator planned for #673 depends on. Both are pure additions; no existing call paths change.

### `kai.eval.retrieval.legacy_retrieve_hits(question, user_id) -> (hits, latency_ms)`

Wraps the `_attach_capture` / `format_context` / `drain` / `_detach_capture` pattern that `_run_one_probe` runs inline today. The whole reason for the helper is the `try/finally`: an exception inside `format_context`, or in the single-payload count check, must not leave the capture handler attached to `kai.memory`'s logger. A leaked handler would double-capture the next probe in the process and abort with "expected one log, got two."

The single-payload check (`len(payloads) != 1` raises `RuntimeError`) matches the legacy harness contract; zero or more than one is a logging-discipline regression, not a recoverable state.

### `kai.memory.embed_texts(texts) -> list[list[float]]`

Public boundary for embedding arbitrary text via the same Mem0 embedder live recall uses (`Config.memory_embedding_model`, validated to 384 dims at init for the Qdrant collection). Cosine similarities computed over these vectors are directly comparable to whatever the search pipeline computes internally.

The only function in the module that touches Mem0's `_memory.embedding_model` attribute is `_embed_via_configured_embedder`; every external caller goes through `embed_texts`. A future Mem0 upgrade that renames or relocates the embedder attribute updates exactly one site.

Behavior:
- `_memory is None` → `RuntimeError` (rather than `[]`, which would let callers silently compute meaningful-looking similarities over zero vectors).
- Empty input list → empty output without touching the embedder.

### Tests

- `tests/test_eval_retrieval.py::TestLegacyRetrieveHits` (4 tests): happy path, zero-payload raises, two-payload raises, capture is detached when `format_context` raises (the regression the `try/finally` exists to prevent).
- `tests/test_memory.py::TestEmbedTexts` (3 tests): uninitialized-store raises, empty input does not call the embedder, order-preserved one-vector-per-text via a mocked embedder.

Full suite: 4509 passed, 1 skipped, no new failures.

## Test plan

- [x] `make check` (ruff lint + format)
- [x] `make test` (full suite, 4509 passed)
- [x] `pytest tests/test_eval_retrieval.py::TestLegacyRetrieveHits tests/test_memory.py::TestEmbedTexts -v` (7 new tests pass)

### DIFF
--- a/src/kai/eval/retrieval.py
+++ b/src/kai/eval/retrieval.py
@@ -662,6 +662,76 @@ def _detach_capture(capture: _RecallLogCapture) -> None:
         logger.setLevel(capture._saved_level)
 
 
+async def legacy_retrieve_hits(question: str, user_id: str) -> tuple[list[dict[str, Any]], int]:
+    """
+    Run the legacy recall pipeline for one question and return its hits.
+
+    Wraps the attach/format_context/drain/detach pattern used inside
+    `_run_one_probe` so external callers (e.g. the collision-probe
+    generator at `kai.eval.gen_collision_probes`) can reuse the
+    legacy harness as a verification gate without reaching into
+    `_attach_capture` / `_detach_capture` directly.
+
+    Why a public helper and not just exposing the captures:
+    callers need the attach / call / drain / detach invariant
+    intact. Three of the four steps are bookkeeping around one
+    awaited call that can raise; the only way to guarantee the
+    capture handler is removed from `kai.memory`'s logger on every
+    exit path is to keep the try/finally in one place. A naive caller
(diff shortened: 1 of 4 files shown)


dcellison:
## Review by Kai

No findings.

Residual risk: `tests/test_memory.py` was truncated in the review bundle, so I could not inspect the full surrounding fixture setup there. The changed hunk itself is consistent with the existing singleton testing pattern shown in `tests/test_eval_retrieval.py`.

PR #692, opened 2026-06-15
Add gen_collision_probes generator CLI

### LINKED ISSUE #673: cross-project collision probe corpus for scoped retrieval evaluator

dcellison:
Closes #673.

## Summary

Adds `python -m kai.eval.gen_collision_probes`: an automated cross-project collision probe corpus generator that produces a complete `kai.eval.retrieval_scoped` probe corpus from the live Mem0 store. Replaces the hand-authored corpus path from the original #673 plan.

Depends on PR #691 (the `legacy_retrieve_hits` and `embed_texts` public helpers).

## Pipeline

- **Project enumeration** uses `config.memory_projects` (YAML registry only) to match `probe_corpus_check`'s registry boundary. Row partition falls back to `metadata["workspace_root"]` for rows missing `project_id`.
- **Two-pass candidate discovery**: Jaccard token-overlap pre-filter, then centroid cosine similarity over Pass-1 survivors. A bounded embedding-only fallback (default cap 200) backfills per-target-project collision quotas when the token pass underdelivers.
- **Drafting** goes through a directly-constructed `OneShotReasoner` modeled on `memory_reclassify` with `purpose="collision_probe_drafting"`. The resolver chain calls `resolve_classification_settings`, then `get_effective_provider` for single-provider backends, then `resolve_user_model(ModelRole.PR_REVIEW, ...)`. Without the normalization step, a codex user inheriting `llm_provider="anthropic"` would look up a non-existent registry entry. A consecutive-failure abort guard exits 1 after 5 typed reasoner failures in a row.
- **Legacy verification** via `legacy_retrieve_hits` + `compute_rank` runs for every collision and every non-project exclusion probe. Drafted probes whose excluded row does not appear in the legacy top-K are counted as shortfalls, not silently accepted.
- **Legacy-default rows** are enumerated from the resolver-global pool via `resolve_memory_scope(metadata).scope_source == SCOPE_SOURCE_LEGACY_DEFAULT`. Raw-metadata selection would miss rows whose stored metadata pre-dates the `scope_source` field. Per-project allocation walks the pool round-robin; row repetition across projects is flagged in the report when the pool is smaller than `len(projects) * --per-project-legacy`.
- **JSONL output** carries a stable `probe_id` extra field (`load_probes` tolerates extra keys). A side-channel `probe_id_by_line_number` mapping survives `load_probes` so the INVESTIGATE branch can name leaking probes by id after self-grade.
- **Self-grade** goes through `kai.eval.retrieval_scoped.evaluate` and destructures `(results, metrics)`. The INVESTIGATE branch reads per-probe leak detail from `results[].excluded_in_prompt` / `excluded_in_candidates`.
- **Promote** is gated by five gates in order: self-grade verdict, structural coverage, legacy-default coverage, parent directory exists, exclusive create succeeds. `--allow-shortfalls` relaxes the two coverage gates but never self-grade or exclusive-create. `--force` opts in to atomic `os.replace`.

## Output paths

| Path | When |
| --- | --- |
| `/tmp/collision-probes-dryrun.jsonl` | every invocation (always overwrites) |
| `/tmp/collision-corpus-report.md` | every invocation (always overwrites) |
| `<data-dir>/eval/probes/collision_v1.jsonl` | only on `--promote` with every gate passing |

Default invocation never touches the canonical path. First `--promote` creates the canonical parent directory; subsequent promotes refuse a pre-existing target via `open(path, "xb")` unless `--force` is passed.

## CLI

```text
python -m kai.eval.gen_collision_probes <user_id> \
  [--backend <name>] [--provider <name>] [--os-user <name>] [--model <s>] \
  [--timeout-s N] [--projects A,B,C] \
  [--per-project-collisions N=5] [--per-project-positive N=2] \
  [--non-project N=3] [--per-project-legacy N=2] \
  [--similarity-threshold F=0.55] [--fallback-cap N=200] \
  [--verify-top-k N=20] \
  [--output PATH] [--promote] [--force] [--allow-shortfalls] \
  [--reject ID1,ID2,...]
```

`--reject` parses the previous dry-run JSONL, drops named `probe_id`s, rebuilds the line-number mapping, and re-self-grades without redrafting. Unknown ids warn but do not abort.

Exit codes: 0 on a completed dry-run or promote, 1 on init/IO failure or backend-looks-dead abort, 2 on settings error, blocked gate, or pre-existing canonical without `--force`.

## Tests

47 tests in `tests/test_gen_collision_probes.py`:

- Pure helpers: tokenize / stopwords / overlap, cosine similarity (identical/orthogonal/zero-magnitude), centroid mean, row-to-project matching with `workspace_root` fallback, TF-IDF (rare token outranks common).
- Provider normalization: codex with inherited `anthropic` resolves to `(codex, openai, pr_review)`; claude with explicit `deepseek` resolves to `(claude, anthropic, pr_review)`.
- `_build_drafting_reasoner` dispatches to the correct concrete class per backend; unknown backend raises.
- `_draft_question` passes `purpose="collision_probe_drafting"` and wraps `OneShotError` in `DraftingFailure`.
- Abort guard fires at `_CONSECUTIVE_FAILURE_ABORT`; a successful draft resets the counter.
- `_verify_exclusion`: rank inside `verify_top_k` accepts, outside drops, absent excluded id drops with rank None.
- Legacy-default enumeration selects rows whose raw metadata has NO `scope` field (via `resolve_memory_scope`), excludes `scope=global` rows.
- Round-robin allocation: 3 rows / 1 project / quota 2 fills cleanly; 2 rows / 2 projects / quota 2 each marks `repeated=True`.
- JSONL write+load round-trip: every row has `probe_id`, the side-channel mapping aligns with `load_probes` line numbers, duplicates get `:dup<n>` suffix.
- Self-grade destructures `(results, metrics)`; verdict classifies `ship` / `regenerate` / `INVESTIGATE` per the metrics shape.
- Five-gate promote evaluator: `regenerate` and `INVESTIGATE` block; collision/positive/non-project/legacy-default shortfalls block; `--allow-shortfalls` relaxes coverage gates but NOT the self-grade gate.
- `_execute_promote_write`: creates the canonical parent directory; refuses a pre-existing target with exit 2 leaving bytes unchanged; `--force` replaces atomically and records the overwrite in the report.
- `--reject` drops matching rows atomically; unknown ids surface in the missing list, not as an abort.
- End-to-end smoke through `main()` with init/registry/resolver/memory/embedder/reasoner/legacy_retrieve_hits/evaluate all mocked; default invocation produces both `/tmp/` artifacts and never touches the canonical path.
- Init failure exits 1.

## Test plan

- [x] `make check` (ruff lint + format) — passes
- [x] `make test` — 4556 passed, 1 skipped
- [x] `pytest tests/test_gen_collision_probes.py -v` — 47 passed
- [ ] Manual acceptance on the operator's live store (next step after merge: skim the dry-run report; confirm verdict ships or regenerate; tune `--similarity-threshold` if needed; `--promote` once)

### DIFF
--- a/src/kai/eval/gen_collision_probes.py
+++ b/src/kai/eval/gen_collision_probes.py
@@ -1,0 +1,2853 @@
+"""
+Automated cross-project collision probe corpus generator.
+
+Reachable as `python -m kai.eval.gen_collision_probes`.
+
+Produces a complete `kai.eval.retrieval_scoped` probe corpus from the
+live Mem0 store with minimal operator touchpoints. Four probe kinds:
+
+1. Collision (per target workspace project B): a project-B-framed
+   question whose project-A row would surface under unscoped
+   retrieval. The scoped pipeline's job is to keep project-A's row
+   out of project-B's context.
+2. Positive-only (per project): a direct question whose project-
+   bound row should surface in that project's context.
+3. Non-project exclusion (corpus-wide): a project-agnostic question
+   whose project-bound row would surface under unscoped retrieval.
+   The scoped pipeline's job is to keep the project-bound row out
+   when no workspace is set.
+4. Legacy-default (per project workspace, drawn from the resolver-
+   global row pool): a project-context question whose resolver-
+   legacy-default row should surface as a positive in that
+   project's context.
+
+The pipeline:
+
+    discover -> draft -> verify (collision + non-project) ->
+    assemble -> self-grade -> [promote gates] -> write
+
+Self-grade goes through `kai.eval.retrieval_scoped.evaluate` and
+returns `(results, metrics)`. A `ship` verdict is necessary but not
+sufficient for promote: structural coverage and legacy-default
+coverage gates layer on top and block promotion unless
+`--allow-shortfalls` is passed. The canonical file is written via
+`open(path, "xb")` so a fresh invocation cannot silently overwrite
+the prior baseline; `--force` opts in to atomic `os.replace`.
+
+Architecture notes:
+
+- Drafting calls do NOT go through `run_review()`. The generator
+  builds its own `OneShotReasoner` via `_build_drafting_reasoner`
+  modeled on `kai.memory_extraction._build_memory_reasoner` so the
+  per-user `models.pr_review` override wins via `resolve_user_model`
+  and so the structured-log `purpose` tag (`"collision_probe_drafting"`)
+  distinguishes drafting calls from PR reviews.
+- Provider normalization: `resolve_classification_settings` returns
+  the raw `llm_provider` cascade. For single-provider backends
+  (claude->anthropic, codex->openai), we normalize via
+  `get_effective_provider` before passing into `resolve_user_model`.
+  Without that step, a codex user with inherited
+  `config.llm_provider="anthropic"` would look up a non-existent
+  `(codex, anthropic, pr_review)` registry entry.
+- Embedding source: `kai.memory.embed_texts` is the public boundary.
+  Every Mem0 internal-attribute hop lives in
+  `_embed_via_configured_embedder` so a Mem0 rename updates one
+  function, not every caller in this module.
+- Legacy-default rows are resolver-global, not project-owned. The
+  generator enumerates the whole-store pool of rows whose
+  `resolve_memory_scope(metadata).scope_source` resolves to
+  `legacy_default`, then walks them round-robin across projects to
+  emit per-project legacy-default probes. When the pool is smaller
+  than `len(projects) * --per-project-legacy`, row repetition
+  across projects is allowed and the report records it.
+
+Exit codes follow the existing eval-CLI convention:
+- 0: dry-run or promote completed successfully.
+- 1: init/IO failure or backend-looks-dead consecutive-failure abort.
+- 2: settings error, blocked promote gate, or canonical path exists
+     without `--force`.
+"""
+
+from __future__ import annotations
+
+import argparse
+import asyncio
+import json
+import logging
+import os
+import re
+import sys
+from dataclasses import dataclass, field
+from datetime import UTC, datetime
+from pathlib import Path
+from typing import Any
+
+from kai.config import (
+    DATA_DIR,
+    Config,
+    MemoryProjectConfig,
+    ModelRole,
+    get_effective_provider,
+    resolve_user_model,
+)
+from kai.eval.retrieval import compute_rank, legacy_retrieve_hits
+from kai.eval.retrieval_scoped import evaluate, load_probes
+from kai.oneshot import (
+    ClaudeOneShotReasoner,
+    CodexOneShotReasoner,
+    GooseOneShotReasoner,
+    OneShotError,
+    OneShotReasoner,
+    OpenCodeOneShotReasoner,
+)
+
+log = logging.getLogger(__name__)
+
+
+# ── Defaults and constants ──────────────────────────────────────────
+
+
+# Matches probe_corpus_check defaults so a dry-run that satisfies the
+# generator's structural coverage gate also satisfies the standalone
+# coverage check on the canonical file. If those defaults move in
+# probe_corpus_check.py, mirror them here; the two checkers are
+# intentionally in lockstep.
+_DEFAULT_PER_PROJECT_COLLISIONS = 5
+_DEFAULT_PER_PROJECT_POSITIVE = 2
+_DEFAULT_NON_PROJECT = 3
+_DEFAULT_PER_PROJECT_LEGACY = 2
+
+# Pass-2 embedding-similarity threshold. Tunable via flag; 0.55 is
+# empirically a reasonable cut between "the embeddings are clearly
+# related" and "purely lexical co-occurrence." Lower the threshold
+# when the live store has narrow per-project vocabularies.
+_DEFAULT_SIMILARITY_THRESHOLD = 0.55
+
+# Pass-3 fallback bound. When the token-filtered pass underdelivers
+# for a project pair, the embedding-only fallback walks up to this
+# many additional project-A rows ordered by descending centroid
+# similarity. Bounded to keep the worst-case embedding cost
+# predictable regardless of how large the project-A row set is.
+_DEFAULT_FALLBACK_CAP = 200
+
+# Acceptance gate for the legacy verify step: a drafted exclusion
+# probe is accepted only if the excluded row appears within the
+# legacy top-K. 20 is the legacy harness's typical injected-prompt
+# depth; outside that range the exclusion is theoretical, not a
+# real retrieval the scoped pipeline must defend against.
+_DEFAULT_VERIFY_TOP_K = 20
+
+# Consecutive-failure abort guard. Matches the memory_reclassify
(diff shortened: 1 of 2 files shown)


dcellison:
## Review by Kai

**Findings**

- **Medium: [src/kai/eval/gen_collision_probes.py] `_row_matches_project` can assign a row to the wrong project.**  
  Failure mode: a row with `metadata["project_id"]` set to project A can still match project B if its stale `workspace_root` equals project B’s root.  
  Mechanism: the function only returns early on `pid == project.project_id`; when `pid` is present but different, it still falls through to the workspace-root fallback. That fallback should only apply to rows missing `project_id`, per the PR description and the comment. This can contaminate per-project row pools and generate invalid “cross-project” probes from rows already explicitly owned by another project.  
  Fix direction: after reading `pid`, return `False` when it is present and not equal, and only run the workspace-root fallback when `pid` is absent/empty. Add a regression with `{"project_id": "phi", "workspace_root": "/work/anvil"}` against project `anvil`.

- **Medium: [src/kai/eval/gen_collision_probes.py] `_run_reject` skips the registry bootstrap that `_run_generate` relies on before self-grade.**  
  Failure mode: `--reject` can self-grade the same dry-run corpus under different scoped-retrieval registry state than a normal generate run, especially when scoped retrieval needs the merged DB/YAML registry cache for workspace detection.  
  Mechanism: `_run_generate` calls `await load_project_registry(config)` before self-grade specifically for its scoped-helper side effect. `_run_reject` goes straight to `_self_grade` and gate evaluation without that side effect.  
  Fix direction: mirror the same `load_project_registry(config)` call at the start of `_run_reject` before `_self_grade`, even if project enumeration for coverage remains YAML-only.

No other blocking issues stood out from the provided final-file context. Residual risk is mainly from the truncated patch: I reviewed the complete changed-file snapshot included in `CHANGED_FILES_AT_HEAD`, but not the omitted middle of the unified diff history.

dcellison:
## Review by Kai

No findings.

Residual risk: the unified diff was truncated, but the full final snapshots of the changed production file and test file were included in `CHANGED_FILES_AT_HEAD`, and I reviewed those. I did not have runnable checkout access, per the review bundle constraints, so I could not independently execute the claimed test suite.

PR #693, opened 2026-06-16
Deprecate kai.eval.retrieval CLI

dcellison:
## Summary

- Document `kai.eval.retrieval` as deprecated in the module docstring; name `kai.eval.retrieval_scoped` as the supported evaluator.
- Update the "LEGACY-RECALL EVALUATOR ONLY" rationale bullet to drop the now-stale "ship in parallel until comparisons settle" framing.
- No behavior change; the CLI still runs and `legacy_retrieve_hits` / `_RecallLogCapture` are still importable by the collision-probe generator.

## Why

With `MEMORY_SCOPED_RECALL_ENABLED` on, production turns go through `assemble_turn_context` and never touch `format_context`; this harness scores `format_context`, so its metrics describe a code path production no longer runs. The scoped sibling exercises the production path and is workspace-aware, so it also evaluates cross-project collision behavior that this harness is structurally blind to.

## What's not in this PR

- Relocating the unscoped recall-capture helpers (`legacy_retrieve_hits`, `_attach_capture`, `_detach_capture`, `_RecallLogCapture`) out of `kai.eval.retrieval` so the collision-probe generator no longer imports from the deprecated module. Follow-up.
- Deleting `kai.eval.retrieval` and its tests. Closes the parent issue when it lands. Follow-up.

## Test plan

- [x] `make check` clean
- [x] `pytest tests/test_eval_retrieval.py` clean (25 passed)
- [ ] Docstring renders cleanly when read in the editor

### DIFF
--- a/src/kai/eval/retrieval.py
+++ b/src/kai/eval/retrieval.py
@@ -1,7 +1,28 @@
 """
-Layer 1 retrieval evaluation harness.
+Layer 1 retrieval evaluation harness (deprecated; see notice below).
 
 Reachable as `python -m kai.eval.retrieval`.
+
+Deprecation notice:
+
+The supported recall evaluator is `kai.eval.retrieval_scoped`. New
+evaluation work and CI gates should target the scoped harness.
+
+With `MEMORY_SCOPED_RECALL_ENABLED` on, every turn is served by
+`assemble_turn_context` and never touches `format_context`; this
+harness scores `format_context`, so its metrics no longer describe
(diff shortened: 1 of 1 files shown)


dcellison:
## Review by Kai

No findings.

The PR is documentation-only and the final file still preserves the CLI entrypoint and the importable helpers named in the deprecation notice. I did not see any contract regression in the provided context. Residual risk is low; no collection warnings or truncated context were provided.

PR #694, opened 2026-06-16
Relocate unscoped recall-capture helpers

dcellison:
## Summary

- New module `kai.eval._unscoped_recall_capture` owns the canonical `legacy_retrieve_hits`, `_RecallLogCapture`, `_attach_capture`, `_detach_capture`, and the `_RECALL_PREFIX` constant.
- `kai.eval.retrieval` keeps the helpers reachable: private ones are re-imported at module top for the CLI's own probe loop; `legacy_retrieve_hits` is exposed via a module `__getattr__` that emits a `DeprecationWarning` so callers still reaching for the public name from here get a visible signal.
- `kai.eval.gen_collision_probes` imports `legacy_retrieve_hits` from its new home directly; `compute_rank` still imports from `kai.eval.retrieval`.
- Tests for the moved symbols (the log-parser round-trip and the four `legacy_retrieve_hits` contract tests) move into `tests/test_eval_unscoped_recall_capture.py`; `tests/test_eval_retrieval.py` loses those two test classes and the corresponding numbered headers, and the module docstring is updated to reflect the smaller surface.

## Why

The deprecation notice on `kai.eval.retrieval` (PR #693) named these helpers as the only reason the deprecated CLI was still importable: the collision-probe generator imports them to verify drafted negative probes against the unscoped gate, and replacing that verifier with scoped retrieval would reject every good negative. Relocating the helpers under a name that says what they are decouples the verifier path from the CLI's lifetime so a follow-up can delete `kai.eval.retrieval` without taking the generator with it.

## What's not in this PR

- Deleting `src/kai/eval/retrieval.py` and `tests/test_eval_retrieval.py`. Closes the parent issue when it lands. Follow-up.

## Test plan

- [x] `make check` clean
- [x] `pytest tests/test_eval_retrieval.py tests/test_eval_unscoped_recall_capture.py tests/test_gen_collision_probes.py` clean (73 passed)
- [x] Full `pytest` suite clean (4557 passed, 1 skipped)
- [x] `from kai.eval.retrieval import legacy_retrieve_hits` fires `DeprecationWarning` with the migration hint
- [x] `from kai.eval._unscoped_recall_capture import legacy_retrieve_hits` does not warn (verified under `warnings.simplefilter('error', DeprecationWarning)`)

### DIFF
--- a/src/kai/eval/_unscoped_recall_capture.py
+++ b/src/kai/eval/_unscoped_recall_capture.py
@@ -1,0 +1,213 @@
+"""
+Unscoped recall-capture helpers.
+
+Buffers the structured `memory.recall` log line that
+`kai.memory.format_context` emits, parses it, and exposes
+`legacy_retrieve_hits` as the public entry point for callers that
+need the unscoped recall result for a single question rather than
+the aggregate metrics the legacy eval CLI in `kai.eval.retrieval`
+produces.
+
+Why a dedicated module:
+
+`kai.eval.retrieval` is deprecated (see its docstring). The
+collision-probe generator in `kai.eval.gen_collision_probes` still
+needs an unscoped verifier: when drafting a negative probe, the
+question is "does this look like a hit under the OLD unscoped
+gate?" and replacing the verifier with scoped retrieval would
+reject every good negative. The helpers live here, under a name
+that says what they are, so the eval CLI can be removed without
+taking the generator's verifier path with it.
+
+The capture machinery:
+
+`format_context` emits a structured log line prefixed
+`memory.recall ` (`_RECALL_PREFIX`) on every call. The capture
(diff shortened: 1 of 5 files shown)


dcellison:
## Review by Kai

No findings.

Residual risk: `RELATED_CONTEXT` notes some lower-priority excerpts were dropped, but the relevant outside consumer shown here (`tests/test_gen_collision_probes.py` patching `gcp.legacy_retrieve_hits`) still works because `gen_collision_probes` keeps that symbol at module scope. The move also preserves the legacy private helper imports used by `kai.eval.retrieval`’s own probe loop, and the deprecation shim should work for `from kai.eval.retrieval import legacy_retrieve_hits` because the public name is intentionally absent and resolved through module `__getattr__`.

PR #695, opened 2026-06-16
Remove kai.eval.retrieval CLI; relocate shared probe primitives

### LINKED ISSUE #672: side-by-side comparison of retrieval evaluators
## Why this exists

PR #671 (#670) added `kai.eval.retrieval_scoped` as a sibling to the legacy `kai.eval.retrieval`. Both harnesses score the live Mem0 store with the same probe-corpus mechanic and the same sweep grid; they exercise different code paths (legacy `format_context` vs the production scoped pipeline). The two ship in parallel until side-by-side comparisons settle which (if either) to retire.

Without an explicit tracker, the dual-harness state is easy to lose: the legacy numbers already do not describe the production read path (it warns at startup), and a future contributor reading the eval module would have to reconstruct the comparison plan from spec history. This issue captures the comparison criteria, the artifacts needed, and the decision rule.

## Decision criteria

A retirement decision needs evidence on three axes:

1. **Precision parity on the legacy probe corpus.** If the scoped harness's precision@5 matches or beats the legacy harness's on the same v1-shaped probes (loaded as v2 with default-empty exclusions), legacy is obsolete for the metric it was built to track.
2. **Exclusion safety on cross-project probes.** The scoped harness's `exclusion_pass_in_prompt` is the safety metric the legacy harness cannot expose at all. A corpus that includes deliberate cross-project collisions (same-text fact in another project's workspace) is required to exercise it; legacy would score those as misses, which is the wrong signal.
3. **Regression visibility on `by_scoped_reason` / `by_active_project`.** A spike in `no_results_after_scope` for one project, or all probes accidentally collapsing to `__none__`, must be visible in the scoped output. Legacy has no equivalent.

If all three hold (parity on axis 1, full coverage on axis 2, stable distribution on axis 3), the legacy harness comes out and this issue closes. If axis 1 regresses, both ship long-term and we use them as paired regressions.

## Artifacts

- Probe corpus (operator-authored, gitignored) with at least:
  - The existing v1 probes loaded as v2 with default-empty exclusions
  - At least three cross-project collision probes (same-text fact pinned to project A; pinned exclusion in project B's workspace)
  - At least one non-project (workspace-null) probe per active project so `by_active_project` reports the `__none__` bucket
- Baseline JSON from each harness on the same corpus, same day, same store snapshot
- One operator-facing comparison report (markdown, in the operator docs tree) summarizing the three axes above

## Acceptance

- [ ] Baseline runs of both harnesses captured against the same probe corpus and store snapshot
- [ ] Comparison report names the precision-parity verdict, the exclusion-safety coverage, and the distribution stability
- [ ] Retirement decision recorded: drop legacy, keep both, or drop scoped (the last is allowed if axis 1 regresses and the safety signal is not worth the cost)
- [ ] If the decision is "drop legacy": follow-up PR removes `kai.eval.retrieval` and its tests; the scoped harness's docstring drops the "shipping in parallel" cross-reference

## Out of scope

- Building a comparison report generator. Manual comparison is fine for the one decision; if a second decision needs the same shape, automate then.
- Migrating the legacy probe file to v2 schema. v1 probes already load through the v2 loader unchanged.

Refs #517. Follow-up to #670.

dcellison:
Closes #672.

## Summary

- Delete `src/kai/eval/retrieval.py` and `tests/test_eval_retrieval.py`. The unscoped eval CLI no longer corresponds to a production code path: with `MEMORY_SCOPED_RECALL_ENABLED` on, every turn flows through `assemble_turn_context` and never touches `format_context`, so the harness's metrics describe a path production no longer runs.
- New `src/kai/eval/_probes.py` owns the shared probe primitives `Probe`, `load_probes`, `probe_set_hash`, `detect_drift`, and `compute_rank`. The behavioral evaluator (`kai.eval.behavioral`) and the collision-probe generator (`kai.eval.gen_collision_probes`) consume the same primitives and both move to the new import path. Their test files follow.
- Tests for the moved primitives move into `tests/test_eval_probes.py`. The CLI-internal tests in `tests/test_eval_retrieval.py` (sweep restoration, evaluate end-to-end, apply-override, legacy-only init warning, aggregate metrics math) drop with the CLI.
- Docstrings and comments swept across `kai.eval.retrieval_scoped`, `kai.eval.behavioral`, `kai.eval.gen_collision_probes`, `kai.eval._unscoped_recall_capture`, and `kai.eval.memory_backend_gate` so they describe present state instead of comparing themselves to the deleted module. The scoped harness's init warning now says "production serves the unscoped pipeline" instead of "production serves the LEGACY pipeline," since "legacy" no longer points at a current module.

## Why this PR is bigger than the original PR 3 plan

The retirement plan handed off from the prior session anticipated only `legacy_retrieve_hits` and `compute_rank` as external consumers of `kai.eval.retrieval`. Verifying against current code surfaced four more import sites: `behavioral.py` imports `Probe`, `detect_drift`, `load_probes`, `probe_set_hash`; `test_eval_behavioral.py` imports `Probe`; `test_eval_unscoped_recall_capture.py` (added in the previous PR) imports `compute_rank`. Deleting the module without first rehoming these primitives would break the behavioral evaluator and the unscoped-capture tests. The shared primitives are not tied to any specific evaluator's lifecycle, so they belong in their own module rather than being moved into either `_unscoped_recall_capture` or `retrieval_scoped`.

## Verification

- `grep -r 'kai\.eval\.retrieval\b' src/ tests/` returns nothing.
- `make check` clean.
- `pytest -x -q` clean: 4544 passed, 1 skipped (down from 4557 by the count of CLI-internal tests that drop with the CLI).
- The collision-probe generator still uses `legacy_retrieve_hits` (re-imported from `kai.eval._unscoped_recall_capture`) and `compute_rank` (now from `kai.eval._probes`) for its unscoped verifier path.
- The scoped harness's other functionality (scoped probe schema, scoped drift detection, scoped sweep envelope) is unchanged; only orphaned comparison comments were rewritten.


dcellison:
## Review by Kai

The PR looks clean based on the provided context. I did not find a concrete code-level regression in the moved shared primitives or the updated import paths.

Residual risk is limited to the truncated/omitted context: `src/kai/eval/gen_collision_probes.py` was not fully included, and the diff was truncated. The visible parts show the critical import migration from `kai.eval.retrieval.compute_rank` to `kai.eval._probes.compute_rank`, plus continued use of `legacy_retrieve_hits` from `_unscoped_recall_capture`, which matches the PR intent.

PR #697, opened 2026-06-16
Add actionability assessment to issue triage

dcellison:
Closes #690.

## Summary

- New triage output fields: `status` (`ready` / `needs_info` / `wontfix_candidate` / `blocked`), `next_action` (single concrete recommendation), `missing_info` (specific questions for the reporter), and `blocked_by` (handle for the blocker when `status = "blocked"`).
- The new fields land on the GitHub triage comment so a maintainer scanning a freshly triaged issue knows the next human move in one read instead of inferring it from labels and a summary line.
- Telegram summary picks up a compact form (status, next action, optional needs-info count, optional blocked-by line) without duplicating the `missing_info` questions, which live on the GitHub comment.

## Why

The current triage one-shot classifies an issue across six axes (labels, duplicate detection, related, project, summary, priority) but stops one step short of being actionable. A maintainer reading "## Triage by Kai" learns priority and labels but still has to read the issue body to decide whether to start work, ask the reporter a question, mark it as a duplicate, or punt because it's blocked on an external dependency. This PR closes that gap.

## How

The prompt edit adds the four new fields to the JSON shape and appends four guidance paragraphs. The `wontfix_candidate` framing is the load-bearing phrase: the model is told to treat it as "a SUGGESTION for human review, NOT a verdict" so it does not over-claim scope judgments without project context. The `next_action` guidance permits maintainer-driven close / assign / label recommendations (which are often the right answer for duplicates and wontfix candidates) while still forbidding claims about Kai's own automatic behavior.

Parser ordering is load-bearing:

1. Capture `raw_summary_has_content` from the raw model dict before the existing summary normalization rewrites a missing summary to a fallback string. A status fallback that inspected the post-normalization summary would always see content and always pick `"ready"`.
2. Normalize `missing_info` to a filtered list (non-list scalars to `[]`, drop non-string and empty entries) before the status fallback inspects it. A scalar `missing_info` therefore collapses to `[]` and contributes nothing to the status decision.
3. Compute the invalid-status fallback from `raw_summary_has_content` plus the normalized `missing_info`: ready if both signals match an actionable issue, otherwise needs_info.
4. Two consistency checks: `blocked` + null `blocked_by` downgrades to `ready` (a blocked label with no handle is worse than ready), and non-`blocked` + populated `blocked_by` drops `blocked_by` (rendering "Blocked by: X" alongside `status = "ready"` would confuse the maintainer). Both inconsistencies log at WARNING.

[...]

PR #698, opened 2026-06-16
Reject bool duplicate_of and bool entries in related

dcellison:
## Summary

- Reject bool values in `duplicate_of`; bool entries inside `related` are filtered out.
- Two regression tests under `TestTriageActionability` confirm `"duplicate_of": true` no longer renders as "Possible duplicate of: #True" and that `[42, true, false, 100]` filters to `[42, 100]`.

## Why

Python's `bool` is a subclass of `int`, so `isinstance(True, int)` returns `True`. PR #697 caught and fixed this defect class on the new `blocked_by` parser. The same gap existed on the pre-existing `duplicate_of` and `related` parsers; the previous review pass only scoped the new code. A maliciously or carelessly malformed model response of `"duplicate_of": true` would have rendered as "Possible duplicate of: #True" in the public-facing triage comment.

## Test plan

- [x] `make check` clean.
- [x] `pytest tests/test_triage.py` clean (97 tests, was 95).
- [x] Two new tests confirm the bool guards: `test_bool_duplicate_of_normalizes_to_none`, `test_bool_entries_in_related_filter_out`.

### DIFF
--- a/src/kai/triage.py
+++ b/src/kai/triage.py
@@ -837,14 +837,23 @@ async def apply_triage(
         labels = []
     labels = [lbl for lbl in labels if isinstance(lbl, str) and lbl.strip()]
 
+    # The `not isinstance(..., bool)` exclusion is load-bearing:
+    # Python's bool is a subclass of int, so a model that returns
+    # "duplicate_of": true would otherwise be accepted as a valid
+    # integer issue number and render as "Possible duplicate of: #True"
+    # in the public-facing comment. Same shape as the `blocked_by`
+    # guard above.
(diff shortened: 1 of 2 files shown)


dcellison:
## Review by Kai

No findings.

The change correctly closes the Python `bool`-as-`int` hole for both `duplicate_of` and `related`, and the added regression tests cover the production rendering paths for GitHub comments and Telegram summaries. I don’t see a contract regression or missed outside consumer in the provided context.

Residual risk: only the usual review limitation that behavior depends on the broader triage response contract staying centralized in `apply_triage`; no collection warnings or truncated relevant context were provided.

PR #702, opened 2026-06-17
Allocate the changed-files budget dynamically under a higher ceiling

### LINKED ISSUE #701: Raise the PR review changed-files context cap
## Summary

On multi-file PRs that touch several large modules, the PR review subprocess drops or truncates whole-file context for the changed files. The reviewer ends up reading the patches without enough surrounding code to judge non-local effects. PR #700's review explicitly named this as residual risk: "truncated or omitted changed-file contents, especially `src/kai/memory.py`, `src/kai/config.py`, `src/kai/install.py`, `src/kai/memory_command.py`, and the omitted portions of the test updates."

The patches themselves were small. What got cut was the surrounding file body the reviewer needs to verify that a deletion doesn't break an unrelated caller or that a rewritten function preserves a non-local invariant.

## Current limits (verified in `src/kai/review.py`)

```
_MAX_CHANGED_FILE_BYTES     = 500_000   # per file, full contents
_MAX_CHANGED_FILES_CHARS    = 450_000   # total across all changed files
_MAX_PATCH_CHARS            =  80_000   # per file, patch only
_MAX_REVIEW_CONTEXT_CHARS   = 600_000   # whole bundle
_MAX_DIFF_CHARS             = 100_000   # combined diff
_MAX_PRIOR_COMMENTS_CHARS   =  50_000   # earlier review comments
_MAX_RELATED_CONTEXT_CHARS  =  35_000
_MAX_LINKED_ISSUES_CHARS    =  30_000
_MAX_COMMITS_CHARS          =  20_000
```

## Why PR #700 hit the cap

Just the four files the reviewer flagged:

```
src/kai/memory.py          145,453 bytes
src/kai/config.py          150,091 bytes
src/kai/install.py         276,038 bytes
src/kai/memory_command.py  113,980 bytes
                          ─────────────
                          685,562 bytes (4 files)
```

The PR changed 22 files in total. Per the per-file cap of 500 KB, none of these individually qualifies for the "too large; contents omitted" path; each one is includable on its own merits. But the running total of all changed files' contents is governed by `_MAX_CHANGED_FILES_CHARS = 450_000`, and 686 KB across just these four files already overruns that by 1.5x before the other 18 files are considered. The budgeter has to cut, and the cut lands on the files where the reviewer most needs context.

## Why this matters

The job of the PR review subprocess is not to judge the patch in isolation; it's to judge whether the patch lands cleanly in the file's surrounding shape. A deletion that looks correct in 50 lines of diff can break a caller 200 lines away. A rewritten helper can subtly change an invariant that another function in the same file relies on. Dropping the whole-file context preserves the patch but loses the non-local check.

PR #700 is a near-miss for this class of risk. The reviewer caught the ABC docstring drift on `AgentBackend.backend_name` because `backend.py` was small enough to fit; the same kind of comment drift inside `memory.py` (where the shadow-mode references were sprinkled across docstrings on the surviving scoped helpers) was only caught in a follow-up scrub because the reviewer literally did not see those docstrings. With a higher cap, the reviewer would have flagged the broader pattern in one pass.

## What is not the problem

- **Per-file cap.** 500 KB per file is fine; the largest production source file (`install.py` at 276 KB) is comfortably under it.
- **Patch cap.** 80 KB per file is generous; the largest patch on PR #700 was `memory.py` at 32 KB.
- **Whole-bundle cap.** 600 KB total bundle budget means there is headroom above the 450 KB changed-files cap: the difference is reserved for prior comments, related context, linked issues, commits, and the prompt scaffolding. On a quiet PR (no prior review threads, no related issues), that reserve is largely unused.

The binding constraint is the changed-files-contents cap specifically, not any of the others.

## Options

Three independent directions; not mutually exclusive.

### Option A: raise `_MAX_CHANGED_FILES_CHARS` outright

Simplest. Lift to 800 KB or 1 MB. Tradeoff is that the whole-bundle cap of 600 KB has to move in lockstep, and the model has to actually have headroom for the larger bundle. Codex's `gpt-5.5` model used for review has a much larger context window than 600 KB; the cap is conservative relative to model capability. The risk is hitting the model's per-call cost / latency more than its capacity.

### Option B: dynamic budget allocation

Compute the actual usage of prior-comments, related-context, linked-issues, and commits at bundle-assembly time, then give the changed-files-contents bucket whatever is left under the whole-bundle cap. Today those buckets are pre-allocated reserves; in practice they are often unused. A quiet PR with no prior review activity could get 550 KB for changed-files instead of 450 KB, paid for by the unused 30 KB issues + 50 KB prior-comments + 20 KB commits reserves. Tradeoff is one extra accounting pass at assembly time.

### Option C: prioritize by patch density, fall back to patch-plus-context

When the changed-files-contents bucket is tight, prefer including whole-file contents for files with the largest patches. Files with tiny patches drop back to "patch plus N lines of surrounding context per hunk" instead of either whole-file or omitted-entirely. A file with 32 KB of patch is more useful to the reviewer in full than a file with 50 bytes of patch is. Tradeoff is implementation complexity (the bundle assembler grows a sort and a fallback renderer) and a behavior change the reviewer needs to learn ("you may see hunks-plus-context instead of full file").

## Recommendation

Start with Option A (the cap lift). It is one constant change plus a verification that the model handles the larger bundle without latency or cost regression. If the cap lift alone is enough to land properly-reviewed multi-file PRs, the work is done.

Option B is the natural follow-up if the lift alone is not enough; it gives a meaningful budget gain on quiet PRs without touching the model-side budget.

Option C is the structural fix; consider it only if A and B together are insufficient on the largest PRs.

## Out of scope

- Switching the review reasoner backend. This is about the bundle the existing backend receives, not which backend reviews.
- Restructuring the review prompt to use a different file representation (e.g., AST summaries). Real change, different surface; not blocking.
- Increasing the model context window. The model is a constraint, but the local caps are well below it.
- The per-file cap or the per-patch cap. Both are headroom-fine today; neither is the binding constraint.

## Acceptance

A PR that touches several files each over 100 KB, totalling over 450 KB of changed-file contents, gets a review where the reasoner can see the full contents of every changed file under the new cap. Measured by: re-running the review subprocess on a PR like #700 after the change and confirming that the residual-risk wording about "truncated or omitted changed-file contents" no longer appears for files the new cap covers.

dcellison:
## Summary

The PR review bundle's changed-files cap was a static `450_000` chars under a `600_000`-char global ceiling and an `80_000`-char patch cap. Multi-file PRs at the scale of recent refactors regularly exceeded those caps; the budgeter dropped large files to existence-only notes and the reviewer flagged residual risk from truncated changed-file contents.

This PR replaces the static cap with a dynamic value computed against the actually-measured weight of every other bundle section, and raises the global ceiling and combined-patch cap so a 23-file PR with ~2.59 MB of changed files plus a ~156 KB combined patch fits without per-file truncation.

## Changes

- Raise `_MAX_REVIEW_CONTEXT_CHARS` from `600_000` to `3_500_000` (5.8x).
- Raise `_MAX_PATCH_CHARS` from `80_000` to `200_000` (2.5x).
- Rename `_MAX_CHANGED_FILES_CHARS` to `_CHANGED_FILES_SECTION_FLOOR`, keep value `450_000`, role changes from static cap to dynamic-allocator floor.
- Add `_OVERHEAD_RESERVE = 50_000` held back from the dynamic cap for rendered overhead the estimator does not measure (boundary tokens, BEGIN/END markers, status lines).
- Add `_compute_dynamic_changed_files_cap(ctx)` which builds a measurement context with `changed_files=()`, reuses `_estimate_total_chars`, subtracts `_OVERHEAD_RESERVE`, and floors at `_CHANGED_FILES_SECTION_FLOOR`.
- Change `_cap_changed_files(files)` to `_cap_changed_files(files, *, cap)`; the caller is now responsible for the cap value.
- Reorder `budget_review_context()` so per-section caps for related, commits, linked-issues, patch, and prior-comments run first; then the dynamic cap is computed against the partially-capped bundle and applied to changed-files.
- Rewrite three stale source-comment blocks in `review.py` so they describe the dynamic-allocator model rather than the static-cap model: the bundle-budget constants block, the `_MAX_CHANGED_FILE_BYTES` comment, and the cross-section ladder comment inside `budget_review_context`.
- Tests: rewrite the existing `test_long_spec_and_conventions_trim_under_global_cap` so its fixture overshoots the new 3.5 MB ceiling, with a pre-budget assertion pinning the fixture against future cap changes. Add `TestDynamicChangedFilesBudget` (4 tests covering the quiet, busy, floor, and end-to-end paths). Add `test_headroom_path_does_not_invoke_ladder` and `test_floor_branch_invokes_ladder` for ladder coverage. Add `TestBundleCapLifts` to pin the four lifted constants.

## Out of scope (deliberately)

- The per-file cap `_MAX_CHANGED_FILE_BYTES = 500_000`. The largest file in the lift-motivating PR is `tests/test_install.py` at 412 KB, under the per-file cap. Worth re-examining when a 500 KB+ file appears in a future PR.
- The `_MAX_DIFF_CHARS = 100_000` diff cap used by the legacy single-prompt path.
- Per-section caps on related-context, linked-issues, commits, and prior-comments. Raising them would consume headroom the dynamic allocator routes to changed-files.
- A patch-plus-context fallback rendering of the changed-files section (Option C). Structural change; deferred.
- Switching the review reasoner, raising the model, or restructuring the bundle into AST summaries.
- Smart per-file priority in the changed-files drop ladder. The higher ceiling makes the existing drop-largest-first ladder rarely fire for realistic shapes.
- Pre-merge cost or latency validation. The lifted ceiling is roughly 800K input tokens at codex `gpt-5.5` rates; cost / latency tuning is a post-deploy concern, captured below.

## Post-deploy concerns

- **Cost / latency.** The 3.5 MB ceiling is 5.8x the prior value. Wall-clock latency and per-review dollar cost may regress on PRs that actually fill the new headroom. Measure both against a real PR review after deploy; if the regression is unacceptable, file a follow-up and revise rather than silently lowering the constant.
- **Residual-risk note.** On the first multi-large-file PR review after deploy, observe whether the reviewer still mentions "truncated or omitted changed-file contents." The phrase can persist for unrelated reasons (binary files, per-file cap notes), so absence is the useful signal.

## Test plan

- [x] `make check` clean
- [x] `make test` passes (4547 passed, 1 skipped)
- [x] `TestDynamicChangedFilesBudget` (4 new tests) passes
- [x] `TestBundleCapLifts::test_lifted_constants_hold_expected_values` passes
- [x] Rewritten `test_long_spec_and_conventions_trim_under_global_cap` exercises the ladder under the lifted ceiling with a pre-budget fixture pin
- [x] `test_headroom_path_does_not_invoke_ladder` and `test_floor_branch_invokes_ladder` cover the new ladder posture
- [ ] Post-deploy: observe cost and latency on the first multi-large-file PR review

Closes #701

### DIFF
--- a/src/kai/review.py
+++ b/src/kai/review.py
@@ -2097,14 +2097,22 @@ def _cap_patch(patch: str) -> tuple[str, list[BudgetNote]]:
 
 def _cap_changed_files(
     files: tuple[ChangedFile, ...],
+    *,
+    cap: int,
 ) -> tuple[tuple[ChangedFile, ...], list[BudgetNote]]:
-    """Truncate full file contents last; never drop the file entirely."""
+    """Truncate full file contents last; never drop the file entirely.
+
+    `cap` is the dynamic changed-files section cap computed by
+    `_compute_dynamic_changed_files_cap`. It is keyword-only so the
+    caller cannot accidentally pass the wrong constant; the caller is
+    responsible for floor enforcement before invoking this function.
+    """
     notes: list[BudgetNote] = []
     if not files:
         return ((), notes)
 
     total = sum(_measure_changed_file(f) for f in files)
-    if total <= _CHANGED_FILES_SECTION_FLOOR:
+    if total <= cap:
         return (files, notes)
 
     # Drop ladder rung 6: truncate the largest file contents first
@@ -2115,7 +2123,7 @@ def _cap_changed_files(
     indexed.sort(key=lambda item: _measure_changed_file(item[1]), reverse=True)
     truncated_paths: list[str] = []
     for idx, f in indexed:
-        if sum(_measure_changed_file(g) for g in [files[i] for i in range(len(files))]) <= _CHANGED_FILES_SECTION_FLOOR:
+        if sum(_measure_changed_file(g) for g in [files[i] for i in range(len(files))]) <= cap:
             break
         if f.content is None:
             continue
@@ -2164,6 +2172,38 @@ def _estimate_total_chars(ctx: PRReviewContext) -> int:
     return total
 
 
+def _compute_dynamic_changed_files_cap(ctx: PRReviewContext) -> int:
+    """Compute the section cap for changed_files given the rest of the bundle.
+
+    Constructs a measurement context with `changed_files=()` and reuses
+    `_estimate_total_chars` so any future change to per-section
+    measurement flows through here automatically. Floors at
+    `_CHANGED_FILES_SECTION_FLOOR` so pathological non-changed sections
+    never starve changed-files below the legacy static-cap behavior.
+    Holds `_OVERHEAD_RESERVE` back for rendered overhead the estimator
+    does not capture (per-section boundary tokens, BEGIN/END markers,
+    status lines).
+    """
+    measurement_ctx = PRReviewContext(
+        repo=ctx.repo,
+        pr_number=ctx.pr_number,
+        metadata=ctx.metadata,
+        linked_issues=ctx.linked_issues,
+        commits=ctx.commits,
+        patch=ctx.patch,
+        changed_files=(),
+        related_context=ctx.related_context,
+        spec=ctx.spec,
+        conventions=ctx.conventions,
+        prior_comments=ctx.prior_comments,
+        budget_notes=ctx.budget_notes,
+        collection_warnings=ctx.collection_warnings,
+    )
+    non_changed = _estimate_total_chars(measurement_ctx)
+    headroom = _MAX_REVIEW_CONTEXT_CHARS - non_changed - _OVERHEAD_RESERVE
+    return max(_CHANGED_FILES_SECTION_FLOOR, headroom)
+
+
 def budget_review_context(ctx: PRReviewContext) -> PRReviewContext:
     """
     Apply per-section caps and the drop ladder to a freshly-collected bundle.
@@ -2187,8 +2227,10 @@ def budget_review_context(ctx: PRReviewContext) -> PRReviewContext:
     """
     notes: list[BudgetNote] = list(ctx.budget_notes)
 
-    # Per-section caps. Run in fixed order so the result is
-    # deterministic for the same input.
+    # Per-section caps for the non-changed-files sections. Run in fixed
+    # order so the result is deterministic for the same input. The
+    # changed-files section runs after these so the dynamic allocator
+    # can size it against the measured weight of everything else.
     related, n = _cap_related(ctx.related_context)
     notes.extend(n)
     commits, n = _cap_commits(ctx.commits)
@@ -2197,11 +2239,11 @@ def budget_review_context(ctx: PRReviewContext) -> PRReviewContext:
     notes.extend(n)
     patch, n = _cap_patch(ctx.patch)
     notes.extend(n)
-    changed_files, n = _cap_changed_files(ctx.changed_files)
-    notes.extend(n)
 
     # Apply prior-comments cap (mirrors the existing
     # _MAX_PRIOR_COMMENTS_CHARS behavior for the legacy diff path).
+    # Capped before the dynamic allocator runs so its measurement
+    # reflects the post-cap size, not the pre-cap raw length.
     prior_comments = ctx.prior_comments
     if prior_comments and len(prior_comments) > _MAX_PRIOR_COMMENTS_CHARS:
         prior_comments = _truncate_with_marker(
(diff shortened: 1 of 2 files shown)


dcellison:
## Review by Kai

No findings.

The implementation keeps the changed-files allocator scoped to the already-capped non-changed sections, preserves the legacy floor, and still runs the cross-section ladder when the estimate remains over the global ceiling. The new tests cover the intended quiet, busy, floor, and end-to-end large-PR paths, plus the lifted constants.

Residual risk is operational rather than code-level: the 3.5 MB ceiling can affect review latency/cost, but that is already called out in the PR description and does not look like a correctness blocker.

dcellison:
## Review by Kai

No findings.

The dynamic allocator is wired against the already-capped non-changed sections, preserves the legacy changed-files floor, and still leaves the cross-section ladder as a backstop when the floor path or estimator drift overshoots. The added tests cover the main budget shapes and pin the lifted constants.

Residual risk remains operational: the higher 3.5 MB ceiling may increase review latency/cost on large PRs, but that is already called out in the PR and is not a code-level blocker.

Issue #532, opened 2026-05-24
Tests: reduce pytest warning count from ~500 to <50


dcellison:
`make test` currently emits ~500 to 540 warnings per run. 91% are from one source.

## Breakdown

| Source | Count | Type |
|---|---|---|
| `tests/test_webhook.py` | ~475 | `NotAppKeyWarning` (aiohttp): tests assign and read app state via string keys like `app["config"] = ...`; aiohttp recommends `app[web.AppKey("config", Config)] = ...` for type-safe access. |
| `tests/test_claude.py` + `tests/test_triage.py` | ~30 | `RuntimeWarning: coroutine 'AsyncMockMixin._execute_mock_call' was never awaited`. An `AsyncMock` was assigned where a `Mock` was expected, and the resulting coroutine is discarded instead of awaited. |
| Miscellaneous | ~15 | `ResourceWarning` on unclosed subprocess transports during teardown; one `FutureWarning`. |

## Why it matters

The two large buckets are different in character:

- The `NotAppKeyWarning` bucket is migration noise. aiohttp 4.x is expected to make the string-key access stricter; the tests should migrate to `web.AppKey` for the same reason production code did. There is no current bug, only a forward-compatibility chore.
- The async-mock leak is a real test-hygiene issue. A discarded coroutine can silently bypass the assertion the test was supposed to drive, leaving the test passing for the wrong reason. Worth fixing because it can mask bugs, not just because it makes the summary noisy.

The `ResourceWarning` bucket is mostly subprocess-teardown noise and is hard to suppress without rewriting tests; safest to leave alone.

## Suggested approach

Two separate PRs:

1. Migrate `test_webhook.py` to `web.AppKey` (~30 call sites, mostly mechanical). Reduces warning count to ~45.
2. Audit and fix the async-mock leak in `test_claude.py` and `test_triage.py`. Each warning points at a specific test that has the wrong mock type or a missing `await`.

A one-line `filterwarnings` entry in `pyproject.toml` would also drop the count quickly, but it hides the underlying migration instead of completing it. Recommend doing the migration.

## Out of scope

- The `ResourceWarning` cluster from subprocess teardown.
- Adding a CI gate on warning count (worth considering only after the cleanup brings the floor low enough to gate against).

dcellison:
## Triage by Kai

[...]

PR #703, opened 2026-06-17
Migrate aiohttp app state to typed AppKey constants

dcellison:
## Summary

Migrates aiohttp app-state access in `src/kai/webhook.py` and its two test files to typed `web.AppKey` constants. aiohttp 4.x is expected to make string-key access (`app["foo"]`) an error rather than the current `NotAppKeyWarning`; this PR brings the production code onto the typed pattern while warnings are still soft, so the migration happens under no time pressure.

This is the first of the two PRs proposed in #532; the async-mock `RuntimeWarning` bucket in `test_claude.py` / `test_triage.py` is a separate, smaller PR (real test-hygiene fix, not forward-compat work). The `ResourceWarning` subprocess-teardown bucket stays untouched per the issue's "safest to leave alone" framing.

## Changes

- Add 14 typed `web.AppKey` constants in `src/kai/webhook.py` (alphabetized; module-scope) covering every app-state field this module sets or reads: `ALLOWED_USER_IDS_KEY`, `ALLOWED_WORKSPACES_KEY`, `CHAT_ID_KEY`, `CONFIG_KEY`, `POOL_KEY`, `PR_REVIEW_COOLDOWN_KEY`, `PR_REVIEW_TIMEOUT_S_KEY`, `SPEC_DIR_KEY`, `TELEGRAM_APP_KEY`, `TELEGRAM_BOT_KEY`, `TELEGRAM_WEBHOOK_SECRET_KEY`, `WEBHOOK_PORT_KEY`, `WEBHOOK_SECRET_KEY`, `WORKSPACE_BASE_KEY`.
- Add `from telegram.ext import Application` so `TELEGRAM_APP_KEY` carries the right runtime type.
- Rewrite 54 `app["..."]` / `request.app["..."]` / `_app["..."]` / `app.get("...")` sites in `webhook.py` to use the typed keys.
- Rewrite 54 sites in `tests/test_webhook.py` to import the keys from `kai.webhook` and access state through them.
- Rewrite 56 sites in `tests/test_webhook_api.py` (a third test file that builds the app via `request.app = {...}` dict literals in pytest fixtures). The dict-literal fixtures now use the AppKey constants as keys; otherwise the production code's typed reads would raise `KeyError` against string-keyed test fixtures (Python dict equality treats `AppKey("foo", T)` and `"foo"` as distinct keys).
- Drop a dead `app["claude_user"] = None` setup line in `tests/test_webhook.py`. No production code reads that key; `claude_user` is a function-local variable derived from `user_config.os_user`, not app state. The line was a fixture vestige.
- Update four `chat_id: int, optional, defaults to app["chat_id"]` docstring snippets to `defaults to the configured webhook chat_id` so the prose tracks the new access pattern.

## Why this approach

- Constants live in `src/kai/webhook.py` because that is the module that owns the Application lifecycle. Tests import from `kai.webhook` so the test app and production reads share the same typed handles.
- Runtime type arg supplied where it can be expressed as `type[T]` (e.g. `web.AppKey("config", Config)`); omitted for `WORKSPACE_BASE_KEY` (whose stored value is `str | None`, which is not a `type[]`).
- `POOL_KEY` uses `object` as the runtime type to avoid importing `SubprocessPool` into `webhook.py`; the value is set in init but never read by `webhook.py` itself, so a precise type would add a dependency for no real consumer.

## Out of scope

- Async-mock `RuntimeWarning` bucket (`test_claude.py` and `test_triage.py`). Separate PR per #532; that bucket is real test hygiene (discarded coroutines can bypass assertions), not a forward-compat chore.
- `ResourceWarning` subprocess-teardown bucket. Per #532, "mostly subprocess-teardown noise and hard to suppress without rewriting tests; safest to leave alone."
- A `filterwarnings` entry in `pyproject.toml` that would hide the warnings without doing the migration. #532 explicitly recommends doing the real migration; this PR does that.
- A CI gate on warning count. Worth considering once the async-mock bucket also lands; until then the floor is still moving.

## Warning count

| State | Total `make test` warnings | `NotAppKeyWarning` |
|---|---|---|
| Before | ~525 | ~475 |
| After  | 122  | 0 |

The 122 remaining warnings are entirely composed of the two buckets called out as out-of-scope above (subprocess-teardown `ResourceWarning` + async-mock `RuntimeWarning`, plus one `FutureWarning`). The actual count is higher than #532's `~50` estimate because the `ResourceWarning` bucket has grown larger than the 15-warning estimate in the issue body. Async-mock bucket is unchanged at ~30.

## Test plan

- [x] `make test` passes (4547 passed, 1 skipped)
- [x] `tests/test_webhook.py` passes (100 tests)
- [x] `tests/test_webhook_api.py` passes (221 tests)
- [x] `make test 2>&1 | grep -c NotAppKeyWarning` returns 0
- [ ] Post-deploy: confirm `kai.log` shows no `NotAppKeyWarning` at startup

Refs #532

### DIFF
--- a/src/kai/eval/modeswitch.py
+++ b/src/kai/eval/modeswitch.py
@@ -848,7 +848,7 @@ def _run_check() -> int:
     # Stats probe: 200 = enabled, 503 = disabled, anything else =
     # unexpected. Issue the request with NO `chat_id` query
     # parameter so the server falls back to its app-default
-    # (`request.app["chat_id"]` at webhook.py's _resolve_chat_id).
+    # (`request.app[CHAT_ID_KEY]` at webhook.py's _resolve_chat_id).
     # Earlier versions of this harness passed `chat_id=1` as a
     # placeholder, which fails: _resolve_chat_id runs after secret
     # auth but before the memory.is_enabled() branch, and rejects
--- a/src/kai/webhook.py
+++ b/src/kai/webhook.py
@@ -55,12 +55,45 @@ from pathlib import Path
 
 from aiohttp import web
 from telegram import Bot, Update
+from telegram.ext import Application
 
 from kai import cron, memory, review, services, sessions, triage
 from kai.config import DATA_DIR, IMAGE_EXTENSIONS, Config, ModelRole, UserConfig, resolve_user_model
 from kai.telegram_utils import chunk_text
 
 log = logging.getLogger(__name__)
+
+
+# ── Application state keys ──────────────────────────────────────────
+#
+# Typed aiohttp AppKey constants for every value this module sets or
+# reads on the running Application. aiohttp 4.x is expected to make
+# string-key access (`app["foo"]`) an error rather than the current
+# NotAppKeyWarning; AppKey is the typed handle pattern that replaces
+# string keys. These constants live at module scope so the test
+# suite imports them from `kai.webhook` and accesses the same
+# Application via the same typed keys, avoiding the same warning at
+# test time.
+#
+# Keep the constants alphabetized by name. The runtime type argument
+# is omitted for Union-typed keys (AppKey accepts `type[T] | None`,
+# which cannot express `T | None` at runtime); the variable
+# annotation carries the type narrowing the call sites rely on.
+
+ALLOWED_USER_IDS_KEY: web.AppKey[set[int]] = web.AppKey("allowed_user_ids", set)
+ALLOWED_WORKSPACES_KEY: web.AppKey[list[str]] = web.AppKey("allowed_workspaces", list)
+CHAT_ID_KEY: web.AppKey[int] = web.AppKey("chat_id", int)
+CONFIG_KEY: web.AppKey[Config] = web.AppKey("config", Config)
+POOL_KEY: web.AppKey[object] = web.AppKey("pool", object)
+PR_REVIEW_COOLDOWN_KEY: web.AppKey[int] = web.AppKey("pr_review_cooldown", int)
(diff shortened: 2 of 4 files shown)


dcellison:
## Review by Kai

Looks clean to me.

I did not find a concrete production regression in the AppKey migration. The changed production reads and writes appear internally consistent: keys are defined once in `kai.webhook`, `start()` writes with those same constants, handlers read with those constants, and the affected tests were updated to build mock apps with matching AppKey keys.

Residual risk: `RELATED_CONTEXT` reports two lower-priority excerpts were dropped, so I can’t prove from this bundle that every outside test/helper touching `kai.webhook._app` or aiohttp app state was visible. The surfaced outside references to `CHAT_ID_KEY` are memory metadata constants, not this new webhook AppKey, so they don’t indicate a conflict.

dcellison:
## Review by Kai

Looks clean to me.

I did not find a concrete production regression in the AppKey migration. The production app writes and reads through the same module-level `web.AppKey` constants, test fixtures that call handlers directly were updated to use those same handles, and the one outside changed-file comment update in `modeswitch.py` is consistent with the new access pattern.

Residual risk: `RELATED_CONTEXT` says 3 lower-priority excerpts were dropped, so this bundle still cannot prove every external helper or test that may touch `kai.webhook._app` directly was visible. The surfaced outside `CHAT_ID_KEY` references are memory metadata constants, not the new webhook AppKey, so they do not indicate a conflict.

PR #711, opened 2026-06-21
Fix the async-mock RuntimeWarning bucket in triage and claude tests

dcellison:
## Summary

Eliminates the async-mock `RuntimeWarning: coroutine 'AsyncMockMixin._execute_mock_call' was never awaited` bucket from `tests/test_triage.py` and `tests/test_claude.py` (the scope called out in #532), and adds a subprocess regression guard that catches reintroduction in those two files.

This is the second of the two PRs proposed in #532. The first cleared the `NotAppKeyWarning` bucket in PR #703. The `ResourceWarning` subprocess-teardown bucket remains untouched per #532's "safest to leave alone" framing.

## Why these warnings matter

Each warning points at a test whose mock setup discards a coroutine, which means the production code's surrounding `try / except Exception:` silently swallows the resulting error and the test passes for the wrong reason. The failure mode is real test hygiene: the previously-vacuous passes are now explicit assertions on the same code paths the tests were nominally exercising.

### Helpers and regression guard (commit 1)

- Add `_mock_aiohttp_session_post(status=200) -> (session, response)` and `_attach_session_to_class(session, mock_cls)` near the top of `tests/test_triage.py`. The two helpers build the canonical async-with-context-manager mock shape and wire a patched `ClientSession` class.
- Add `tests/test_warnings.py` with a single regression test that runs a focused pytest subprocess against `tests/test_triage.py tests/test_claude.py` and asserts both that the subprocess exits cleanly and that the captured output contains zero `AsyncMockMixin._execute_mock_call` occurrences. The `returncode == 0` gate guards against a nested run that exits early (collection error, import failure, unrelated test failure) and would otherwise report zero leaks for the wrong reason.

### Group 1: triage Shape A (commit 2)

The 16 success-path sites in `tests/test_triage.py` all built `session.post` via `mock_session = AsyncMock()`. That made `mock_session.post` itself an `AsyncMock` whose call returned a coroutine; production then evaluated `async with <coroutine> as resp:` which fails on `__aenter__`, the surrounding `try / except` swallowed the `AttributeError`, the coroutine was garbage-collected without being awaited, and the runtime emitted the warning. The test still passed because its assertions checked `gh` CLI calls via `kai.triage.asyncio.create_subprocess_exec`, not HTTP behavior.

- Replace the 5-line `AsyncMock` setup at all 16 sites with `_mock_aiohttp_session_post()` + `_attach_session_to_class()`.
- Add `assert mock_session.post.called` (or rely on an existing downstream `assert_called_once()` / `if mock_session.post.called:` check) at each site so the previously-vacuous HTTP step now carries a real assertion.
- Triage file warning count drops from 31 to 3.

### Group 2: triage Shape B (commit 3)

The 3 error-path sites at `test_does_not_raise_on_connection_error`, `test_does_not_raise_on_timeout`, and `test_logs_warning_on_failure` set `session.post = AsyncMock(side_effect=ConnectionError(...))` (or `TimeoutError()`, `RuntimeError("boom")`). Production calls `session.post(...)` synchronously, then evaluates `async with <result>`. `AsyncMock(side_effect=...)` attaches the exception to the unawaited coroutine the call returns; that coroutine is never awaited, so the side effect never fires, and the test's intended failure branch is not exercised.

- Rewrite each site to `MagicMock(side_effect=...)` on a sync `session.post`. MagicMock raises synchronously, which matches production's call shape and lets production's `try / except` actually catch the error.
- Add `mock_session.post.assert_called_once()` to each error-path test so the previously-vacuous pass now asserts the HTTP call actually ran.
- Triage file warning count drops from 3 to 0.

### Group 3: claude.py stderr + wait_for (commit 4)

Two shapes in `tests/test_claude.py`:

- **22 sites set `mock_proc.stderr = AsyncMock()` without configuring `stderr.readline`.** Production's `_drain_stderr()` background task loops on `await self._proc.stderr.readline()` then evaluates `text = line.decode().strip()`. With the bare `AsyncMock`, `stderr.readline` was an auto-generated `AsyncMock` whose await returned an `AsyncMock` instance (not bytes); the subsequent `line.decode()` triggered another auto-generated `AsyncMock` call that created a coroutine which was never awaited. The warning surfaced at file level because the GC happened during the file's teardown sweep rather than inside any one test.
  - Fix: replace each bare `mock_proc.stderr = AsyncMock()` with the configured pair `mock_proc.stderr = MagicMock()` + `mock_proc.stderr.readline = AsyncMock(return_value=b"")`.

- **1 site (`test_timeout`) patched `asyncio.wait_for` with `side_effect=TimeoutError`.** Production already constructed a coroutine via `self._proc.stdout.readline()` before passing it to `wait_for`; the mock raised without awaiting or closing that coroutine.
  - Fix: rewrite the patch with a local helper `async def _timeout_after_closing(coro, *args, **kwargs)` that closes the supplied coroutine before raising `TimeoutError`. The `*args, **kwargs` shape tolerates production's `timeout=` keyword argument; a narrower signature would raise `TypeError` and reintroduce the leak.

- Claude file warning count drops from 23 to 0.

### Format / lint cleanup (commit 5)

Drop unused `mock_resp` unpacking name in 13 sites (ruff RUF059); reformat the regression-guard assertion message line. No behavior change.

## Out of scope

- The `ResourceWarning` subprocess-teardown bucket. Per #532: "safest to leave alone."
- A `filterwarnings` entry suppressing `RuntimeWarning` from `unittest.mock`. The bucket masks real test-hygiene bugs; suppression defeats the purpose.
- A CI gate on overall warning count.
- The `claude_user="daniel"` test assumptions in `tests/test_claude.py` that fail in environments where the current OS user is `daniel`. Pre-existing, environment-dependent, surfaced by the spec reviews but not this PR's scope.
- Argument-level assertions on `session.post`.

## Findings during implementation (worth flagging)

The async-mock pattern exists outside the spec's declared scope as well. Live measurement against `make test`:

- `tests/test_webhook.py`: ~35 warnings (highest concentration outside the spec scope; same broken `session.post`-via-`AsyncMock` shape).
- `tests/test_review.py`: 1 warning.
- `tests/test_tts.py`: 1 warning.
- `tests/test_webhook_api.py`: 1 warning.

#532's issue body described the bucket as "`tests/test_claude.py` + `tests/test_triage.py`: ~30 warnings" but the real bucket is larger. The regression guard in `tests/test_warnings.py` is scoped to the spec's two files; reintroduction in `test_webhook.py` or the other files would not trip the guard. If clearing those is worth doing, it is the same shape of work and should land as a follow-up PR (or PRs).

## Warning count

| State | Total `make test` warnings | `AsyncMockMixin._execute_mock_call` |
|---|---|---|
| Before | 122 | 53 |
| After  | 59  | 0 in spec scope (`tests/test_triage.py` + `tests/test_claude.py`); ~38 remain in `tests/test_webhook.py` and 3 elsewhere |

## Test plan

- [x] `make test` passes (4548 passed, 1 skipped)
- [x] `tests/test_triage.py` passes (97 tests, 0 async-mock warnings)
- [x] `tests/test_claude.py` passes (164 tests, 0 async-mock warnings)
- [x] `tests/test_warnings.py::test_async_mock_warning_bucket_stays_empty` passes
- [x] `.venv/bin/python -m pytest -W default tests/test_triage.py tests/test_claude.py 2>&1 | grep -c "AsyncMockMixin._execute_mock_call"` returns 0
- [ ] Post-deploy: confirm the daemon's runtime logs remain clean (no `AsyncMockMixin` chatter)

Refs #532

### DIFF
--- a/tests/test_claude.py
+++ b/tests/test_claude.py
@@ -123,7 +123,8 @@ class TestCommandConstruction:
         with patch("asyncio.create_subprocess_exec", new_callable=AsyncMock) as mock_exec:
             mock_proc = MagicMock()
             mock_proc.returncode = None
-            mock_proc.stderr = AsyncMock()
+            mock_proc.stderr = MagicMock()
+            mock_proc.stderr.readline = AsyncMock(return_value=b"")
             mock_exec.return_value = mock_proc
 
             await claude._ensure_started()
@@ -143,7 +144,8 @@ class TestCommandConstruction:
         with patch("asyncio.create_subprocess_exec", new_callable=AsyncMock) as mock_exec:
             mock_proc = MagicMock()
             mock_proc.returncode = None
-            mock_proc.stderr = AsyncMock()
+            mock_proc.stderr = MagicMock()
+            mock_proc.stderr.readline = AsyncMock(return_value=b"")
             mock_exec.return_value = mock_proc
 
             await claude._ensure_started()
@@ -169,7 +171,8 @@ class TestCommandConstruction:
         with patch("asyncio.create_subprocess_exec", new_callable=AsyncMock) as mock_exec:
             mock_proc = MagicMock()
             mock_proc.returncode = None
-            mock_proc.stderr = AsyncMock()
+            mock_proc.stderr = MagicMock()
+            mock_proc.stderr.readline = AsyncMock(return_value=b"")
             mock_exec.return_value = mock_proc
 
             await claude._ensure_started()
@@ -189,7 +192,8 @@ class TestCommandConstruction:
         with patch("asyncio.create_subprocess_exec", new_callable=AsyncMock) as mock_exec:
             mock_proc = MagicMock()
             mock_proc.returncode = None
-            mock_proc.stderr = AsyncMock()
+            mock_proc.stderr = MagicMock()
+            mock_proc.stderr.readline = AsyncMock(return_value=b"")
             mock_exec.return_value = mock_proc
 
             with caplog.at_level("WARNING", logger="kai.config"):
@@ -221,7 +225,8 @@ class TestCommandConstruction:
         ):
             mock_proc = MagicMock()
             mock_proc.returncode = None
-            mock_proc.stderr = AsyncMock()
+            mock_proc.stderr = MagicMock()
+            mock_proc.stderr.readline = AsyncMock(return_value=b"")
             mock_exec.return_value = mock_proc
 
             await claude._ensure_started()
@@ -252,7 +257,8 @@ class TestCommandConstruction:
         ):
             mock_proc = MagicMock()
             mock_proc.returncode = None
-            mock_proc.stderr = AsyncMock()
+            mock_proc.stderr = MagicMock()
+            mock_proc.stderr.readline = AsyncMock(return_value=b"")
             mock_exec.return_value = mock_proc
 
             await claude._ensure_started()
@@ -277,7 +283,8 @@ class TestCommandConstruction:
         ):
             mock_proc = MagicMock()
             mock_proc.returncode = None
-            mock_proc.stderr = AsyncMock()
+            mock_proc.stderr = MagicMock()
+            mock_proc.stderr.readline = AsyncMock(return_value=b"")
             mock_exec.return_value = mock_proc
 
             await claude._ensure_started()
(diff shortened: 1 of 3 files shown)


dcellison:
## Review by Kai

No findings.

The PR looks clean against the provided context. The mock-shape changes match the production call patterns shown in the tests: `session.post(...)` is treated as a synchronous call returning an async context manager, and `stderr.readline()` now returns bytes instead of a mock object. The subprocess regression guard also avoids the false pass case by checking the nested pytest return code before counting warning text.

Residual risk is limited to the intentionally scoped guard: `tests/test_warnings.py` only protects `tests/test_triage.py` and `tests/test_claude.py`, while the PR description notes similar warnings remain in other test files. That is a known out-of-scope follow-up rather than a regression in this PR.

dcellison:
## Review by Kai

No findings.

The changed mocks now match the production async context-manager shape, and the `wait_for` timeout test closes the already-created coroutine before raising. The new subprocess guard is scoped honestly to `tests/test_triage.py` and `tests/test_claude.py` and checks the nested pytest return code before trusting the warning count.

Residual risk is limited to the intentionally out-of-scope warning sites called out in the PR description, especially `tests/test_webhook.py`; the provided context does not show a regression from this PR there.

PR #713, opened 2026-06-21
Clear the async-mock RuntimeWarning bucket suite-wide

### LINKED ISSUE #712: Clear async-mock warnings outside the triage and claude test scope
PR #711 cleared the `AsyncMockMixin._execute_mock_call was never awaited` bucket from `tests/test_triage.py` and `tests/test_claude.py` (the scope #532 declared). Live measurement after that PR shows the same warning class still surfacing from other test files, but the root cause is different and needs its own scoping.

## Where the warnings come from

`.venv/bin/python -m pytest -W default 2>&1 | grep -B1 "AsyncMockMixin._execute_mock_call"` after PR #711 lands:

- `tests/test_webhook.py`: ~35 warnings, concentrated at two `await _process_github_event_for_user(...)` call sites in `src/kai/webhook.py` (the fallback-admin path: 19 warnings; the per-user fan-out path: 15 warnings), with scattered ones tagged to pytest / asyncio / aiohttp internals.
- `tests/test_review.py`: 1 warning.
- `tests/test_tts.py`: 1 warning.
- `tests/test_webhook_api.py`: 1 warning.

Total: roughly 38 warnings still in the bucket.

## Why this is not the same fix as #711

Tracemalloc origins for the `test_webhook.py` warnings point at `_pytest/unraisableexception.py:33` and `aiohttp/web_protocol.py:510`, not at a test-side mock construction site. The two production lines (`await _process_github_event_for_user(...)`) are themselves correctly awaited. The leak is a coroutine generated inside the call that aiohttp's `Application` teardown garbage-collects without awaiting.

`tests/test_webhook.py` does not use the broken `mock_session = AsyncMock()` shape that motivated #711's fix; it uses `patch(..., new_callable=AsyncMock)` exclusively, which is correct usage. The remaining warnings are about aiohttp `web.Application` fixture teardown leaking background coroutines, not about per-test mock construction.

So the fix shape is different:

| | #532 / PR #711 scope | This issue |
|---|---|---|
| Root cause | Broken mock shape (`AsyncMock` where `MagicMock` was needed) | Aiohttp fixture teardown leaking background coroutines |
| Fix shape | Swap mock type, configure attribute | Per-test fixture teardown or background-task management |
| Investigation cost | Low (pattern is uniform across 19+22 sites) | Higher (per-test, root cause not yet pinned) |

## Suggested scope

- Investigate `tests/test_webhook.py` first; that is where most of the warnings concentrate. Likely candidates: how the aiohttp test client is constructed, where background tasks are created, whether `aiohttp.ClientSession` instances are awaited to closure.
- The three singletons in `test_review.py`, `test_tts.py`, `test_webhook_api.py` may turn out to be the same root cause or three different ones; classify after the `test_webhook.py` investigation.
- Once the root cause is pinned, extend the regression guard at `tests/test_warnings.py::test_async_mock_warning_bucket_empty_in_triage_and_claude_tests` (or add a sibling guard) so the additional files are also protected.

## Acceptance criteria

After this work lands:

1. `.venv/bin/python -m pytest -W default 2>&1 | grep -c "AsyncMockMixin._execute_mock_call"` returns 0.
2. The regression guard at `tests/test_warnings.py` is extended (or a sibling guard added) so all files now clean of the warning are protected against reintroduction.
3. `make test` passes with no new failures.

## Out of scope

- The `ResourceWarning` subprocess-teardown bucket (~90 warnings). Per #532: "safest to leave alone."
- A `filterwarnings` entry to suppress the warning. The bucket masks real test-hygiene bugs; suppression defeats the purpose.

Refs #532, follows PR #711.

dcellison:
## Summary

Clears the `AsyncMockMixin._execute_mock_call was never awaited` warning bucket from the rest of the test suite. PR #711 cleared the bucket inside its declared scope (`tests/test_triage.py` and `tests/test_claude.py`); issue #712 was filed based on my own analysis that said the rest of the bucket had a different root cause. That analysis was wrong: tracemalloc shows the same two root shapes everywhere.

## Why the original #712 framing was off

I traced one test in `tests/test_webhook.py` and stopped reading the stack one frame too early, at `aiohttp/web_protocol.py:510`. The actual leak was further down at `src/kai/config.py` inside `resolve_user_model()`, where production calls `config.default_models.get(role.value, "")`. The test fixture set `mock_config = AsyncMock()`; `Config` is a sync dataclass, so the chained attribute call returned a coroutine that was assigned to a local and discarded. Same shape as PR #711's triage fix.

I correct that here and close #712 instead of writing up new spec work.

## Changes

Two recurring root shapes:

### 1. AsyncMock where MagicMock was needed

Sync interfaces (`Config` dataclass; `aiohttp.ClientSession`'s sync `.post(...)`) mocked as `AsyncMock`. Every chained attribute call returns a coroutine production then misuses (assigns to a value, passes to `async with`).

- `tests/test_webhook.py`: five `mock_config = AsyncMock()` sites: the `_build_test_app` fixture, two helper functions (`_make_config`, `_make_config_with_users`), and two direct test setups in `TestAddRemoveAllowedChatId`. Swap to `MagicMock`. This single category cleared 35 warnings.
- `tests/test_review.py`: four sites in `TestSendReviewSummary` using the same broken `mock_resp = AsyncMock()` + `mock_session = AsyncMock()` shape as PR #711's triage Shape A. Swap to `MagicMock` and add `assert mock_session.post.called` so the previously-vacuous HTTP-send step carries a real assertion.

### 2. `wait_for` leaks coroutine

Patched `asyncio.wait_for` with a bare `side_effect=TimeoutError` raises without closing the coroutine production passed to it. Same shape as PR #711's `tests/test_claude.py::test_timeout`.

- `tests/test_tts.py::test_piper_timeout`: production calls `asyncio.wait_for(proc.communicate(input=text.encode()), timeout=120)`. Patch with `_timeout_after_closing(coro, *args, **kwargs)` that closes the coroutine before raising.
- `tests/test_transcribe.py::TestRun::test_timeout`: same shape, same fix.
- `tests/test_tts.py::test_ffmpeg_timeout`: already had a per-call `_wait_for_side_effect` helper; added `coro.close()` to its raise branch.

### Regression guard

Extend `tests/test_warnings.py` to cover the six files now cleaned (the original two from PR #711 plus the four cleaned up here). Rename the test from `..._empty_in_triage_and_claude_tests` to `..._empty_across_guarded_files`. Add a module-docstring section enumerating the two recurring root shapes so a future contributor can recognize them quickly. The file list is the load-bearing artifact; future contributors should add new files here when fixing the same pattern.

## Results

- **Async-mock bucket suite-wide:** 38 → 0.
- **Total `make test` warnings:** drops to ~3 (variable, ResourceWarning subprocess-teardown floor only).
- **Tests:** 4548 passed, 1 skipped. No new failures.
- **Regression guard runtime:** ~2 seconds (scoped to six known files; running the full suite would take ~35 seconds).

## Test plan

- [x] `make test` passes (4548 passed, 1 skipped)
- [x] `tests/test_warnings.py::test_async_mock_warning_bucket_empty_across_guarded_files` passes
- [x] Each touched file passes in isolation with zero `AsyncMockMixin._execute_mock_call` occurrences
- [x] `.venv/bin/python -m pytest -W default 2>&1 | grep -c "AsyncMockMixin._execute_mock_call"` returns 0

## Out of scope

- `ResourceWarning` subprocess-teardown bucket (~3 warnings remaining). Per #532's "safest to leave alone" framing.
- CI gate on overall warning count. Worth revisiting now that the floor is much lower.

Closes #712

### DIFF
--- a/tests/test_review.py
+++ b/tests/test_review.py
@@ -1024,9 +1024,9 @@ class TestSendReviewSummary:
         """Success summary includes PR link and title."""
         meta = _metadata(repo="owner/repo", number=42, title="Add feature X")
 
-        mock_resp = AsyncMock()
+        mock_resp = MagicMock()
         mock_resp.status = 200
-        mock_session = AsyncMock()
+        mock_session = MagicMock()
         mock_session.post.return_value.__aenter__ = AsyncMock(return_value=mock_resp)
         mock_session.post.return_value.__aexit__ = AsyncMock(return_value=False)
 
@@ -1036,6 +1036,7 @@ class TestSendReviewSummary:
             await send_review_summary(meta, True, 8080, "secret")
 
         # Verify the POST was made with correct URL and content
+        assert mock_session.post.called
         call_args = mock_session.post.call_args
         assert "localhost:8080/api/send-message" in call_args[0][0]
         body = call_args[1]["json"]
@@ -1048,9 +1049,9 @@ class TestSendReviewSummary:
         """Failure summary says 'Failed to review'."""
         meta = _metadata()
 
-        mock_resp = AsyncMock()
+        mock_resp = MagicMock()
         mock_resp.status = 200
-        mock_session = AsyncMock()
+        mock_session = MagicMock()
         mock_session.post.return_value.__aenter__ = AsyncMock(return_value=mock_resp)
         mock_session.post.return_value.__aexit__ = AsyncMock(return_value=False)
 
@@ -1059,6 +1060,7 @@ class TestSendReviewSummary:
             mock_cs.return_value.__aexit__ = AsyncMock(return_value=False)
             await send_review_summary(meta, False, 8080, "secret")
 
+        assert mock_session.post.called
         body = mock_session.post.call_args[1]["json"]
         assert "Failed to review" in body["text"]
 
@@ -1076,9 +1078,14 @@ class TestSendReviewSummary:
         """When notify_chat_id is set, chat_id is included in the POST body."""
         meta = _metadata()
 
-        mock_resp = AsyncMock()
+        # session.post() is a sync call returning an async context
+        # manager. MagicMock keeps the call return synchronous; an
+        # AsyncMock session would make .post() return a coroutine
+        # that production then tries to use as `async with`, leaking
+        # the AsyncMockMixin._execute_mock_call warning.
+        mock_resp = MagicMock()
         mock_resp.status = 200
-        mock_session = AsyncMock()
+        mock_session = MagicMock()
         mock_session.post.return_value.__aenter__ = AsyncMock(return_value=mock_resp)
         mock_session.post.return_value.__aexit__ = AsyncMock(return_value=False)
 
@@ -1087,6 +1094,7 @@ class TestSendReviewSummary:
             mock_cs.return_value.__aexit__ = AsyncMock(return_value=False)
             await send_review_summary(meta, True, 8080, "secret", notify_chat_id=-100123)
 
+        assert mock_session.post.called
         body = mock_session.post.call_args[1]["json"]
         assert body["chat_id"] == -100123
 
@@ -1095,9 +1103,9 @@ class TestSendReviewSummary:
         """When notify_chat_id is None, chat_id is NOT in the POST body."""
         meta = _metadata()
 
-        mock_resp = AsyncMock()
+        mock_resp = MagicMock()
         mock_resp.status = 200
-        mock_session = AsyncMock()
+        mock_session = MagicMock()
         mock_session.post.return_value.__aenter__ = AsyncMock(return_value=mock_resp)
         mock_session.post.return_value.__aexit__ = AsyncMock(return_value=False)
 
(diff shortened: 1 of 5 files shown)


dcellison:
## Review by Kai

No findings.

Residual risk: the review bundle only includes changed test files and one outside excerpt, so I could not inspect the production implementations of `resolve_user_model()`, `_handle_github()`, `send_review_summary()`, or the subprocess helpers directly. Within the provided context, the mock-shape changes and coroutine-closing `wait_for` side effects look consistent with the stated warning root causes, and the regression guard avoids the obvious false-positive path by asserting the nested pytest return code before counting warnings.

dcellison:
## Review by Kai

No findings.

Residual risk: `RELATED_CONTEXT` only includes one outside excerpt, and six related excerpts were dropped by the bundle. I could not independently verify the production call shapes for `resolve_user_model()`, `send_review_summary()`, `synthesize_speech()`, or `_run()`, but the changed tests are internally consistent with the described sync/async mock boundaries and the existing `test_claude.py` timeout helper pattern.

PR #796, opened 2026-08-10
Gate default backend prompt in config wizard

dcellison:
## Summary
- stop asking existing default-Claude installs to choose a Default backend during make config
- keep the backend prompt for first-time setup and existing non-default backend installs so codex/goose/opencode installs are not silently changed
- update install wizard tests to pin the prompt-gating contract and remove retired default-model prompt inputs

## Validation
- make check
- make check-install-constraints
- .venv/bin/python -m pytest tests/test_install.py -q
- .venv/bin/python -m pytest

### DIFF
--- a/src/kai/install.py
+++ b/src/kai/install.py
@@ -1242,22 +1242,37 @@ def _cmd_config() -> None:
     print()
 
     # -- Default backend --
-    print("-- Default backend --")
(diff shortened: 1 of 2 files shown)


dcellison:
Closing this PR. The direction is wrong: it broadened the default-backend/config behavior. We will redo this from a clean branch focused on removing operator-supplied backend binary paths and deriving choices from the admin-owned backend registry.

PR #916, opened 2026-08-14
docs: plan Workshop Telegram independence

dcellison:
## Summary

- define precisely what it means for Kai Workshop to be independent of Telegram while retaining Telegram as a supported client and delivery adapter
- inventory the remaining coupling in runtime policy, application startup, subprocess ownership, persistent state, context, scheduling, internal APIs, media, and integrations
- specify the target transport-neutral application host and protected runtime-profile boundary
- organize the migration into six coherent milestones with explicit installed qualification, security, rollback, observability, and retirement gates

## Key decision

The first implementation work should separate runtime profiles from Telegram-keyed humans, then move core service construction out of the Telegram `Application`. The decisive acceptance test is an installed Workshop-only system with no Telegram token or synthetic Telegram identity that can execute, stream, cancel, restart, and recover agent work.

## Scope

Documentation only. This PR changes no runtime behavior, configuration, schema, installation path, or Telegram behavior.

## Validation

- cross-checked current coupling against `runtime_profiles.py`, `runtime_pool.py`, `compatibility_state.py`, `bot.py`, `webhook.py`, `cron.py`, `sessions.py`, and `internal_api_auth.py`
- reconciled PRs #902–#905 as completed active-directory namespace migrations while preserving the remaining record/API migration work
- retained backend-registry model defaults and avoided assuming any backend exists or has fallback priority
- `git diff --check`

### DIFF
--- a/kai-workshop-telegram-independence-plan.md
+++ b/kai-workshop-telegram-independence-plan.md
@@ -1,0 +1,1007 @@
+# Kai Workshop Telegram Independence Plan
+
+**Status:** Proposed migration plan  
+**Date:** 2026-08-13  
+**Scope:** Remove Telegram as a required host, identity source, and internal
+routing authority while retaining Telegram as a fully supported Kai client and
+delivery adapter.
+
+## 1. Executive summary
+
+Kai Workshop already has transport-independent principals, channels, agents,
+messages, runs, attempts, client sessions, browser access, and durable delivery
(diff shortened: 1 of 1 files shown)


dcellison:
Superseded by #917, which now serves as the authoritative Telegram-independence epic and execution tracker. This documentation branch will be discarded without merging.

Issue #917, opened 2026-08-14
Epic: Make Kai Workshop independent of Telegram


dcellison:
# Kai Workshop Telegram Independence Plan

**Status:** Complete — final corrective reconciliation passed  
**Date:** 2026-08-26  
**Scope:** Remove Telegram as a required host, identity source, and internal
routing authority while retaining Telegram as a fully supported Kai client and
delivery adapter.

## Epic tracking

The objective of this epic is to make Kai Workshop and its agents fully usable
without configuring Telegram, while preserving Telegram private conversations,
streaming, files, voice, commands, and notification-group delivery whenever the
Telegram adapter is enabled.

- [x] Milestone 1: independent protected runtime policy
- [x] Milestone 2: transport-neutral application host
- [x] Milestone 3: installed Workshop-only qualification
- [x] Milestone 4: canonical state and context authority
- [x] Milestone 5: shared feature and integration services
- [x] Milestone 6: scoped internal APIs and transition retirement
- [x] Final configured-backend Workshop-only and hybrid qualification
- [x] Close or convert every associated transition-ledger entry

Implementation issues and pull requests should be linked under the applicable
milestone as they are opened. A milestone is complete only when its automated
and installed exit gates below have passed; merging its implementation is not
by itself sufficient.

## Current execution status (2026-08-25)

- Milestone 1 is complete and installed. PRs #922, #924, #926, #928,
  #930, #932, #933, and #939 established independently authoritative
  protected runtime policy and storage.
- Milestone 2 is complete and installed. PRs #941, #943, and #944
  established the transport-neutral application host, an optional Telegram
  adapter lifecycle, and core-supervised Workshop HTTP lifecycle.
- Milestone 3 is complete and installed. PR #945 added Workshop-only
  deployment, and the full installed Workshop-only and restored-hybrid
  qualification passed.
- Milestone 4 is complete and installed. PRs #946 through #949 made
  conversation continuity, execution state, semantic memory, and operational
  state canonical. PRs #1056 and #1057 completed canonical transcript
  authority and memory-source parsing against current provenance contracts.
  The full Workshop-only restart, older-history/export, provenance, hybrid,
  Telegram private-text, and GitHub notification qualification passed. Issue
  #1047 records the installed evidence; superseded PR #950 remains closed.
- Milestone 5 is complete and installed. Core-owned scheduling,
  transport-neutral media/artifacts, settings/workspaces, canonical GitHub
[...]

PR #950, opened 2026-08-15
Close Workshop Milestone 4 transcript authority

dcellison:
## Summary

Close Workshop epic Milestone 4 for protected canonical private-text conversations.

- stop protected Telegram and Workshop-browser private text from writing compatibility JSONL
- build semantic-memory episode context from prior canonical human/agent message pairs
- resolve new semantic-memory source views from the exact canonical principal, channel, run, source message, and result message in SQLite
- provide a deterministic, transport-free canonical NDJSON transcript export for older-history search
- stage an atomically refreshed export for every protected execution under the channel's existing secured history directory
- add an operator CLI export and an authoritative install-status diagnostic
- update the Workshop transition ledger with the remaining JSONL removal gate

## Authority boundary

For protected canonical private text, Workshop SQLite is now the only transcript authority:

- fresh backend context comes from the bounded canonical channel timeline
- episode-classifier context comes from canonical messages strictly before the current inbound message
- new memory provenance source views verify and read the canonical run/message pair
- neither the Telegram private-text path nor the Workshop browser path reads or writes compatibility JSONL

This PR does **not** claim global JSONL retirement. Media, commands, groups, schedules, and integrations remain compatibility routes until Milestone 5 migrates them across canonical execution. Their current JSONL behavior is preserved deliberately, and install status reports that distinction.

## Canonical transcript export

The derived export:

- is rebuilt from canonical message projections
- contains canonical IDs, author facts, bodies, reply lineage, timestamps, and event positions
- omits Telegram/external transport identifiers
- is atomically replaced as `history/<channel-id>/canonical-transcript.ndjson`
- uses mode `0600` plus the existing channel reader ACL policy
- cannot be consumed by compatibility readers, which scan only `*.jsonl`
- is advisory/searchable context only; SQLite remains authoritative

Operators can emit a current snapshot to standard output with:

```text
python -m kai workshop transcript export --channel-id <canonical-channel-id>
```

## Failure and security posture

- canonical provenance mismatches fail closed before source content is returned
- canonical source lookup opens the deployed database read-only and verifies the complete owner/run/message tuple
- export refresh failure is visible in logs but does not prevent the canonical run from completing
- concurrent refreshes are serialized with canonical database work so an older run cannot overwrite a newer snapshot
- existing compatibility archives are neither modified nor deleted by the cutover

## Tests

- `make check`
- `make typecheck`
- `5842 passed, 1 skipped` from the complete pytest suite

New regression contracts explicitly fail if canonical memory ingestion or source lookup consults JSONL, verify export isolation from `*.jsonl` readers, exercise atomic/ACL-aware snapshot creation, and pin the new CLI and install-status surfaces.

## Installed qualification after merge

1. Run `make install`.
2. Confirm `make install-status` reports `Workshop transcript authority: active` with protected JSONL reads and writes disabled.
3. Send one private-text request through Telegram and one through Workshop; both should respond normally.
4. Ask Kai about older conversation history and inspect a new memory source view if a new extracted memory is available.

Closes the transcript-authority portion of #917 Milestone 4. Remaining compatibility-route JSONL retirement belongs to Milestone 5.


dcellison:
Reassessed against current `main` on 2026-08-22.

This PR will not be mechanically rebased or conflict-resolved. Its central
direction remains valid: protected private text should stop reading and writing
JSONL, episode context and new memory source views should use canonical
messages, and any transcript file should be a derived export rather than an
authority.

The implementation now predates substantial memory ownership, provenance,
scope, rendering, telemetry, application-host, and execution changes. Two
designs also need correction rather than transplantation:

- prior pairs must be selected through completed run/channel-agent lineage,
  not adjacent human/agent messages in channel order; and
- the full transcript must not be rebuilt and `fsync`ed under the shared
  database lock before every execution.

#1047 is now the unified current-main contract for completing transcript
authority, classifying the existing JSONL parity discrepancy, preserving newer
memory protections, and defining the installed Milestone 4 exit gate.

Closing this PR as superseded preserves its implementation and tests as design
reference. It is not merged.

PR #1216, opened 2026-08-29
Hydrate durable route evidence in Agent run cards

### LINKED ISSUE #1214: Show durable explicit routes in Workshop run activity
## Problem

The explicit task route introduced by #1213 executes correctly and stores an immutable `workshop_run_routing_decisions` row, but the Workshop activity card does not reliably appear with the route decision for fast runs.

The authenticated run-state endpoint serializes `routing_decision`, while `run.lifecycle.changed` SSE events serialize only the run lifecycle projection. The client normally updates the activity card from SSE, so the durable route evidence is absent from the path that drives the card.

## Required outcome

- Include the immutable routing decision, when present, in lifecycle stream events.
- Parse and retain that decision on the client activity state.
- Preserve compatibility for accepted events emitted before a decision exists.
- Prove a fast accepted/started/completed route displays its final routed backend without polling or changing principal defaults.
- Keep ordinary run activity behavior unchanged.

## Qualification

An explicit Coding route to `opencode:deepseek` must leave a visible completed activity card reporting the routed disposition and backend option after the response arrives.

Parent: #1212

dcellison:
## Summary
- attach immutable routing decisions to command, run-state, and cancellation responses
- preserve routing evidence when the Workshop client parses or restores a run
- prove an explicit Coding response renders its route without relying on SSE timing

## Verification
- 6,286 Python tests passed
- 121 Workshop client tests passed
- make check passed
- Workshop production build passed

Closes #1214


dcellison:
Closing unmerged after installed qualification clarified that per-message backend routing is not a desired Workshop workflow. The correction will remove the premature composer and personal-policy surfaces instead of extending them.

